PR #83, opened 2021-06-29
Replaced declarative recipe with imperative recipe. fixes #78.

### LINKED ISSUE #78: NoAutowiredAnnotion should preserve `@Autowired` constructor annotations when multiple constructors exist.
[spring-beans-and-dependency-injection](https://docs.spring.io/spring-boot/docs/current-SNAPSHOT/reference/htmlsingle/#using.spring-beans-and-dependency-injection)
> If a bean has more than one constructor, the one intended for dependency injection should be `@Autowired` annotated.

```java
@Service
public class MyAccountService implements AccountService {

    private final RiskAssessor riskAssessor;

    private final PrintStream out;

    @Autowired
    public MyAccountService(RiskAssessor riskAssessor) {
        this.riskAssessor = riskAssessor;
        this.out = System.out;
    }

    public MyAccountService(RiskAssessor riskAssessor, PrintStream out) {
        this.riskAssessor = riskAssessor;
        this.out = out;
    }

    // ...

}
```

traceyyoshima:
The constructor will be inferred if there is only one exists.

This will leave other auto-wires on fields and setters in place, which might be redundant.
We may check the variables in the constructor for auto-wired fields/setters to remove redundant annotations in the future.


src/main/java/org/openrewrite/java/spring/NoAutowiredOnConstructor.java
```diff
@@ -0,0 +1,75 @@
...
+            for (Statement s : classDecl.getBody().getStatements()) {
+                if (s instanceof J.MethodDeclaration) {
+                    if (((J.MethodDeclaration) s).isConstructor()) {
+                        constructorCount++;
+                    }
+                }
+            }
+
+            if (constructorCount == 1) {
+                doAfterVisit(new RemoveAnnotation("@org.springframework.beans.factory.annotation.Autowired(true)"));
```
tkvangorder, line 69: This will remove ALL auto-wired annotations from the file if one constructor exists. We ONLY want to remove the auto-wired from the Constructor and leave all other annotations alone.

src/test/kotlin/org/openrewrite/java/spring/NoAutowiredTest.kt
```diff
@@ -61,7 +56,6 @@ class NoAutowiredTest : JavaRecipeTest {
         """
     )
 
-    @Disabled
     @Issue("https://github.com/openrewrite/rewrite-spring/issues/78")
```
tkvangorder, line 59: These tests must be more exhaustive. 

have two constructors, one constructor plus auto-wired field and/or setter, etc.

tkvangorder (changes requested):
This is still not right, we need to ONLY remove the annotation from a constructor, when :

- only one constructor exists
- and that constructor has an `@Autowired` annotation.

src/test/kotlin/org/openrewrite/java/spring/NoAutowiredTest.kt
```diff
@@ -30,25 +29,335 @@ class NoAutowiredTest : JavaRecipeTest {
...
             public class DatabaseConfiguration {
                 private final DataSource dataSource;
                 
-                @Autowired
-                public DatabaseConfiguration(DataSource dataSource) {
+                public @Autowired DatabaseConfiguration(DataSource dataSource) {
                 }
             }
-        """,
+        """.trimIndent(),
```
jkschneider, line 360: `trimIndent()` is never needed on these tests.

tkvangorder (approved):
LGTM

Issue #140, opened 2022-01-12
2.4-2.5: Messages in the Default Error View


fabapp2:
[Release Notes](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.5-Release-Notes#messages-in-the-default-error-view)
The `message` attribute in the default error view is now removed rather than blanked when it is not shown. If you parse the error response JSON, you may need to deal with the missing item.

You can still use the `server.error.include-message` property if you want messages to be included.

### Condition
- Application is a web-application, has dependency `org.springframework:spring-webmvc` or  `org.springframework:spring-webflux`

#### Description
> The recipe found a dependency on `org.springframework:spring-webmvc` or `org.springframework:spring-webflux`
> The `message` attribute in the default error view is now removed rather than blanked when it is not shown. If you parse the error response JSON, you may need to deal with the missing item.
> You can still use the `server.error.include-message` property if you want messages to be included.

PR #154, opened 2022-01-21
Draft implementation for #140

fabapp2:
The current implementation does not look at the version when interpreting dependencies. 
Resulting in the recipe being shown as applicable for all dependency versions independent of the version. 
My understanding was that the version number cannot be reliably retrieved during a recipe?


pway99:
@fabapp2  this looks like it could be defined as a simple declarative recipe, something like the following

```yaml
name: org.openrewrite.java.spring.boot2.MessagesInTheDefaultErrorView
displayName: Find spring-web projects ...
description: |
  The message attribute in the default error view is now removed in Spring Boot 2.5.
  A dependency on `org.springframework:spring-webmvc` or `org.springframework:spring-webflux`.
  The `message` attribute in the default error view was removed in Spring Boot 2.5 rather than blanked when it is not shown.
  If you parse the error response JSON, you may need to deal with the missing item.
  You can still use the `server.error.include-message` property if you want messages to be included."
recipeList:
  - org.openrewrite.maven.search.FindDependency:
      groupId: org.springframework
      artifactId: spring-webmvc
  - org.openrewrite.maven.search.FindDependency:
      groupId: org.springframework
      artifactId: spring-webmvc
```

jkschneider:
Let's add a marker to the POM indicating that either of these dependencies are present.

pway99:
Hi @fabapp2 I have converted this to a declarative recipe per our discussion this morning

PR #157, opened 2022-01-24
Logging Shutdown Hook recipe initial impl

### LINKED ISSUE #141: 2.4-2.5: Logging Shutdown Hooks
[Release Notes](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.5-Release-Notes#logging-shutdown-hooks)
We now register a logging shutdown hook by default for jar based applications to ensure that logging resources are released when the JVM exits. If your application is deployed as a war then the shutdown hook is not registered since the servlet container usually handles logging concerns.

Most applications will want the shutdown hook. However, if your application has complex context hierarchies, then you may need to disable it. You can use the `logging.register-shutdown-hook` property to do that.

### Condition
- packaging type is `jar`

#### Description

> We now register a logging shutdown hook by default for jar based applications to ensure that logging resources are released when the JVM exits
> if your application has complex context hierarchies, then you may need to disable it. You can use the `logging.register-shutdown-hook` property to do that.

BoykoAlex:
@fabapp2 initial implementation marking the maven project
Fixes #141

### DIFF
--- a/src/main/java/org/openrewrite/java/spring/boot2/search/LoggingShutdownHooksRecipe.java
+++ b/src/main/java/org/openrewrite/java/spring/boot2/search/LoggingShutdownHooksRecipe.java
@@ -1,0 +1,105 @@
+/*
+ * Copyright 2022 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.java.spring.boot2.search;
+
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.Recipe;
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.internal.lang.Nullable;
+import org.openrewrite.maven.MavenVisitor;
+import org.openrewrite.maven.tree.Maven;
+import org.openrewrite.maven.tree.Pom;
+import org.openrewrite.maven.tree.Scope;
+import org.openrewrite.xml.XPathMatcher;
+import org.openrewrite.xml.tree.Xml;
+
+import java.util.Collection;
+
+/**
+ * Mark Pom's of projects where logging shutdown hook may need to be disabled
+ *
+ * @author Alex Boyko
+ */
+public class LoggingShutdownHooksRecipe extends Recipe {
+
+    private static final String NEEDS_SEARCH_MARKER = "NEEDS_SEARCH_MARKER";
+
+    private static final XPathMatcher PROJECT_MATCHER = new XPathMatcher("/project");
+
+    private static final String SEARCH_MARKER_DESCRIPTION = "LoggingShutDownHook";
+
+    @Override
+    public String getDisplayName() {
+        return "Logging Shutdown Hooks";
+    }
+
+    @Override
+    public String getDescription() {
+        return "We now register a logging shutdown hook by default for jar based applications to ensure that " +
+                "logging resources are released when the JVM exits. If your application is deployed as a war then " +
+                "the shutdown hook is not registered since the servlet container usually handles logging concerns" +
+                "\n\n" +
+                "Most applications will want the shutdown hook. However, if your application has complex context" +
+                " hierarchies, then you may need to disable it. You can use the logging.register-shutdown-hook" +
+                " property to do that.";
+    }
+
+    @Override
+    protected @Nullable TreeVisitor<?, ExecutionContext> getSingleSourceApplicableTest() {
+        return new MavenVisitor() {
+
+            @Override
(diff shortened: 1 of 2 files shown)


jkschneider:
According to #141 let's add the additional condition of "on Spring Boot 2.5"

BoykoAlex:
Update the PR with condition for boot 2.5.

src/main/java/org/openrewrite/java/spring/boot2/LoggingShutdownHooksRecipe.java
```diff
@@ -0,0 +1,76 @@
...
+                "logging resources are released when the JVM exits. If your application is deployed as a war then " +
+                "the shutdown hook is not registered since the servlet container usually handles logging concerns" +
+                "\n\n" +
+                "Most applications will want the shutdown hook. However, if your application has complex context" +
+                " hierarchies, then you may need to disable it. You can use the logging.register-shutdown-hook" +
+                " property to do that.";
+    }
+
+    @Override
+    protected @Nullable TreeVisitor<?, ExecutionContext> getApplicableTest() {
```
pway99, line 41: a suggestion for the applicable test would be to do something similar to the DependencyInsight recipe which will find both direct and transitive dependencies. Also moving the packaging check to the recipe visitor will improve recipe performace by not having to inspect every tag of every pom 
```java
@Override
protected MavenVisitor getApplicableTest() {
    return new MavenVisitor() {
        @Override
        public Xml visitTag(Xml.Tag tag, ExecutionContext context) {
            Xml.Tag t = (Xml.Tag) super.visitTag(tag, context);
            if (isDependencyTag()) {
                Pom.Dependency dependency = findDependency(t);
                if (dependency != null) {
                    Collection<Pom.Dependency> match = dependency.findDependencies("org.springframework.boot", "spring-boot");
                    if (match.stream().anyMatch(dep -> dep.getVersion().startsWith("2.5"))) {
                        t = t.withMarkers(t.getMarkers().searchResult());
                    }
                }
            }
            return t;
        }
    };
}
```

src/main/java/org/openrewrite/java/spring/boot2/LoggingShutdownHooksRecipe.java
```diff
@@ -0,0 +1,76 @@
...
+                        return tag.withMarkers(tag.getMarkers().searchResult());
+                    }
+                }
+                return tag;
+            }
+        };
+    }
+
+    @Override
+    protected TreeVisitor<?, ExecutionContext> getVisitor() {
```
pway99, line 68: Consider using a cursor message provide as an alternative to visitor state variables.  In the following example a cursor message is added to the first Maven parent wich is then removed and inspected when when visiting the Maven tree.
```java
@Override
protected MavenVisitor getVisitor() {
    return new MavenVisitor() {
        @Override
        public Maven visitMaven(Maven maven, ExecutionContext ctx) {
            Maven mvn = super.visitMaven(maven, ctx);
            if (getCursor().pollMessage("NEEDS_SEARCH_MARKER") != null) {
                return mvn.withMarkers(maven.getMarkers().searchResult(SEARCH_MARKER_DESCRIPTION));
            }
            return mvn;
        }
        @Override
        public Xml visitTag(Xml.Tag tag, ExecutionContext context) {
            if(PROJECT_MATCHER.matches(getCursor())) {
                if (tag.getChildValue("packaging").orElse("jar").equalsIgnoreCase("jar")) {
                    getCursor().dropParentUntil(Maven.class::isInstance).putMessage("NEEDS_SEARCH_MARKER", true);
                }
            }
            return tag;
        }
    };
}
```

BoykoAlex:
@pway99 thanks for the review :-) I have updated the PR. Lets try these changes.

src/main/java/org/openrewrite/java/spring/boot2/LoggingShutdownHooksRecipe.java
```diff
@@ -0,0 +1,74 @@
...
+            @Override
+            public Maven visitMaven(Maven maven, ExecutionContext ctx) {
+                Collection<Pom.Dependency> deps = maven.getModel().getDependencies(Scope.Compile);
+                if (deps.stream()
+                        .filter(d -> d.getArtifactId().equals("spring-boot")
+                                && d.getVersion().startsWith("2.5.")
+                                && d.getGroupId().equals("org.springframework.boot"))
+                        .findFirst()
+                        .isPresent()
+                ) {
```
pway99, line 54: for this recipe the maven dependency check can be the [applicable test](https://github.com/openrewrite/rewrite/blob/6809be2616697debf5360337dc656f29efa57503/rewrite-core/src/main/java/org/openrewrite/Recipe.java#L169-L180)

> A recipe can optionally include an applicability test that can be used to determine whether it should run on a
   set of source files (or even be listed in a suggested list of recipes for a particular codebase).

pway99: thanks for catching my error.  the `singleSourceApplicability` test is where it belongs, then tags are only visited when the maven pom has a boot 2.5 dependency

src/main/java/org/openrewrite/java/spring/boot2/LoggingShutdownHooksRecipe.java
```diff
@@ -0,0 +1,74 @@
+package org.openrewrite.java.spring.boot2;
```
pway99, line 1: Since this is just a search and mark recipe, can you move this to a new search package, `org.openrewrite.java.spring.boot2.search`.

BoykoAlex:
Updated the PR

pway99:
Looks good thanks @BoykoAlex !

PR #160, opened 2022-01-26
Spring Integration Scheduler Pool Size. Boot -> 2.5

### LINKED ISSUE #138: 2.4-2.5: Task Scheduling Harmonization with Spring Integration
[Release Notres](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.5-Release-Notes#task-scheduling-harmonization-with-spring-integration)
Spring Integration now reuses an available `TaskScheduler` rather than configuring its own. In a typical application setup relying on the auto-configuration, this means that Spring Integration uses the auto-configured task scheduler that has a pool size of 1. To restore Spring Integration’s default of 10 threads, use the `spring.task.scheduling.pool.size` property.

### Condition
- spring integration is on the classpath `org.springframework.integration:spring-integration-core`

#### Description

> The scan found `org.springframework.integration:spring-integration-core` on the classpath.
> Spring Integration now reuses an available `TaskScheduler` rather than configuring its own. In a typical application setup relying on the auto-configuration, this means that Spring Integration uses the auto-configured task scheduler that has a pool size of 1. To restore Spring Integration’s default of 10 threads, use the `spring.task.scheduling.pool.size` property.

BoykoAlex:
Fixes #138 

PR for initial feedback because there is usually plenty of feedback ;-)

(cc @fabapp2 )

If conditions are met comment made in `application.properties` file either over the `spring.task.scheduling.pool.size` property or just at the end if property not found. YAML properties still left to add.

### DIFF
--- a/src/main/java/org/openrewrite/java/spring/boot2/search/IntegrationSchedulerPoolRecipe.java
+++ b/src/main/java/org/openrewrite/java/spring/boot2/search/IntegrationSchedulerPoolRecipe.java
@@ -1,0 +1,184 @@
+/*
+ * Copyright 2022 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.java.spring.boot2.search;
+
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.Recipe;
+import org.openrewrite.SourceFile;
+import org.openrewrite.Tree;
+import org.openrewrite.internal.ListUtils;
+import org.openrewrite.java.AnnotationMatcher;
+import org.openrewrite.java.JavaIsoVisitor;
+import org.openrewrite.java.marker.JavaProject;
+import org.openrewrite.java.tree.Comment;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.java.tree.TextComment;
+import org.openrewrite.marker.Markers;
+import org.openrewrite.maven.MavenVisitor;
+import org.openrewrite.maven.tree.Maven;
+import org.openrewrite.maven.tree.Pom;
+import org.openrewrite.maven.tree.Scope;
+import org.openrewrite.properties.PropertiesVisitor;
+import org.openrewrite.properties.search.FindProperties;
+import org.openrewrite.properties.tree.Properties;
+import org.openrewrite.semver.DependencyMatcher;
+import org.openrewrite.yaml.YamlIsoVisitor;
+import org.openrewrite.yaml.search.FindProperty;
+import org.openrewrite.yaml.tree.Yaml;
+
+import java.util.*;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+/**
+ * @author Alex Boyko
+ */
+public class IntegrationSchedulerPoolRecipe extends Recipe {
+
+    private static Pattern APP_PROPS_FILE_REGEX = Pattern.compile("^application.*\\.properties$");
+    private static Pattern APP_YAML_FILE_REGEX = Pattern.compile("^application.*\\.ya?ml$");
+
+    private static final String PROPERTY_KEY = "spring.task.scheduling.pool.size";
+
+    private static final String PROPS_MIGRATION_MESSAGE = " TODO: Consider Scheduler thread pool size for Spring Integration";
+    private static final String GENERAL_MIGRATION_MESSAGE = " TODO: Scheduler thread pool size for Spring Integration either in properties or config server\n";
+
+    @Override
+    public String getDisplayName() {
+        return "Integration Sceduler Pool Size";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Spring Integration now reuses an available TaskScheduler rather than configuring its own. In a" +
+                " typical application setup relying on the auto-configuration, this means that Spring Integration" +
+                " uses the auto-configured task scheduler that has a pool size of 1. To restore Spring Integration’s" +
+                " default of 10 threads, use the spring.task.scheduling.pool.size property.";
+    }
+
+    private boolean isApplicableMavenProject(Maven maven) {
+        DependencyMatcher boot25Matcher = DependencyMatcher.build("org.springframework.boot:spring-boot:2.5.X").getValue();
+        DependencyMatcher integrationMatcher = DependencyMatcher.build("org.springframework.integration:spring-integration-core").getValue();
+        Collection<Pom.Dependency> deps = maven.getModel().getDependencies(Scope.Compile);
+        boolean boot25 = false;
+        boolean si = false;
+        for (Pom.Dependency d : deps) {
+            if (!boot25) {
+                boot25 = boot25Matcher.matches(d.getGroupId(), d.getArtifactId(), d.getVersion());
+            }
+            if (!si) {
+                si = integrationMatcher.matches(d.getGroupId(), d.getArtifactId());
+            }
+            if (boot25 && si) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    @Override
+    protected MavenVisitor getApplicableTest() {
+        return new MavenVisitor() {
+            @Override
+            public Maven visitMaven(Maven maven, ExecutionContext ctx) {
+                if (isApplicableMavenProject(maven)) {
+                  return maven.withMarkers(maven.getMarkers().searchResult());
+                }
+                return maven;
+            }
+        };
+    }
+
+    @Override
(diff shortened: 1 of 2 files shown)


src/main/java/org/openrewrite/java/spring/boot2/IntegrationSchedulerPoolRecipe.java
```diff
@@ -0,0 +1,135 @@
...
+                    Properties.Entry schedulerPoolEntry = getCursor().pollMessage(SCHEDULER_PROPERTY_ENTRY);
+                    int idx = schedulerPoolEntry != null ? p.getContent().indexOf(schedulerPoolEntry) : -1;
+                    List<Properties.Content> contents = new ArrayList<>(p.getContent());
+                    Properties.Comment comment = new Properties.Comment(Tree.randomId(), "\n", Markers.EMPTY, MIGRATION_MESSAGE);
+                    if (idx >= 0) {
+                        contents.add(idx, comment);
+                    } else {
+                        contents.add(comment);
+                    }
+                    p = p.withContent(contents);
```
pway99, line 117: consider using `ListUtils#insertAll` and `ListUtils#concat` for these list modifications

src/main/java/org/openrewrite/java/spring/boot2/IntegrationSchedulerPoolRecipe.java
```diff
@@ -0,0 +1,135 @@
...
+//                if (boot25 && si) {
+//                    return maven.withMarkers(maven.getMarkers().searchResult());
+//                }
+//                return maven;
+//            }
+//        };
+//    }
+
+    @Override
+    protected List<SourceFile> visit(List<SourceFile> before, ExecutionContext ctx) {
```
pway99, line 70: Hi @BoykoAlex, just curious if you have tried leveraging your `applicableTest`, returning a `PropertiesVisitor` from `Recipe.getVisitor` override, and adding a singleSourceApplicable test to check for the `application.properties' file?

pway99:
Hi @BoykoAlex,  added a bit of feedback ;-)

src/main/java/org/openrewrite/java/spring/boot2/IntegrationSchedulerPoolRecipe.java
BoykoAlex: I was also thinking of running a YamlVisitor to modify application yaml file. I could specify both yaml and properties file filters in singleSourceApplicable test but how shall i proceed with `getVisitor()` then?

src/main/java/org/openrewrite/java/spring/boot2/IntegrationSchedulerPoolRecipe.java
BoykoAlex: Aha, thanks, will consider this of course!

src/main/java/org/openrewrite/java/spring/boot2/IntegrationSchedulerPoolRecipe.java
pway99: one idea is to convert the yaml and properties visitors to recipes which can scheduled to run next from the applicableTest
```java
if (boot25 && si) {
     doAfterVisit(new IntegrationSchedulerPoolYamlRecipe());
     doAfterVisit(new IntegrationSchedulerPoolPropertiesRecipe());
```

pway99: although, maybe it's just better to use the before method instead

BoykoAlex:
@pway99 Seems like I need help organizing this recipe's implementation a bit.
How it works:
1. Look at maven files and create a set of all classpath resources (not sources, i.e. `maven.getResources(...)`. The collected paths can be further filtered to have only application properties or yaml files.
2. Visit all application properties files that are in the set created in step 1. to create a comment
3. Visit all application yaml files that are in the set created in step 1.

The set of classpath resources (or classpath application properties/yaml files) needs to be assembled at some point and then passed on to visitors or recipes processing boot properties/yaml files. I've been advised previously to use `visit(List<SourceFile> before, ExecutionContext ctx)` if i'm to pass state between visitors rather than use execution context messages.
The order of various recipe calls seems to be:
`getApplicableTest()`
  <start loop over recipes list>
    <start loop over source files>
      `getSingleSourceApplicableTest()`
       main Visitor (`getVisitor()`)
   <end loop over source files>
  <end loop over recipes list>
`visit(List<SourceFile> before, ExecutionContext ctx)`

At the moment I packed everything in the last method... What would be the good alternatives?
I was thinking 3 recipes chined togther. First produces the set of classpath resources boot props.yaml paths. The other two are properties and yaml recipes to create a comment. The issue is how to collects and pass the set of paths from the first recipe to the other two?

pway99:
@BoykoAlex exellent pairing with you on this one, below is an approach very similar to your original effort 
```java
protected MavenVisitor getApplicableTest() {
    return new MavenVisitor() {
        @Override
        public Maven visitMaven(Maven maven, ExecutionContext ctx) {
            // if maven has both spring-boot 2.5 and spring-integration dependency then add search marker
        }
    };
}

@Override
protected List<SourceFile> visit(List<SourceFile> before, ExecutionContext ctx) {

    Set<Path> resources;
    for (SourceFile sourceFile : before) {
        // if Maven then set resources to mavens resources
        // break;
    }

    for (SourceFile sourceFile : before) {
        // if resources contains path then
        // visit with either Properties or Yaml visitor to add comment
    }

    return before;
}
```

BoykoAlex:
Pushed what i think should be more or less inline with what we've decided.
A few notes:
1. Collect applicable maven java project using markers created by `MavenProjectParser` Sources corresponding to a specific maven project are marked with the same marker.
2. Leave comment over `spring.task.scheduling.pool.size` property in properties/yaml files if property is there
3. If property is not externalized to a file leave a comment over boot app class declaration.

pway99:
thanks @BoykoAlex I'll have a look

src/main/java/org/openrewrite/java/spring/boot2/search/IntegrationSchedulerPoolRecipe.java
```diff
@@ -0,0 +1,185 @@
...
+                    }
+                } else if (APP_YAML_FILE_REGEX.matcher(fileName).matches() && source instanceof Yaml) {
+                    Set<Yaml.Block> foundEntriesValues = FindProperty.find((Yaml) source, PROPERTY_KEY, false);
+                    if (!foundEntriesValues.isEmpty()) {
+                        projectsWithCommentOverProperty.add(javaProject);
+                        return (SourceFile) new YamlIsoVisitor<ExecutionContext>(){
+                            @Override
+                            public Yaml.Mapping.Entry visitMappingEntry(Yaml.Mapping.Entry entry, ExecutionContext context) {
+                                if (foundEntriesValues.contains(entry.getValue())) {
+//                                    return maybeAutoFormat(entry, entry.withPrefix("\n#" + MIGRATION_MESSAGE + entry.getPrefix()), ctx);
```
pway99, line 152: can you double-check this, the autoformat will ensure the comment has the correct indent

BoykoAlex: If I try the `autoFormat` like below:
```
return maybeAutoFormat(entry, entry.withPrefix("\n#" + PROPS_MIGRATION_MESSAGE + entry.getPrefix()), ctx);
```
I get one of my tests failing which shows that the comment is not indented and the next line indentation is broken as well. Seemed like a bug to me.

pway99: In this case autoFormat needs to be aware of the parent tree.  Try invoking autoFormat with the parent-cursor
```java
return maybeAutoFormat(entry, entry.withPrefix("\n#" + PROPS_MIGRATION_MESSAGE + entry.getPrefix()),
  ctx, getCursor().getParent());
```

BoykoAlex: Doesn't help, still the comment not indented. Event if drop parent until Yaml (the root). I've added code to get the indent without auto format. Seems like auto format does not work with comments. I quickly looked at format tests in yaml rewrite and did not notice any tests with comments...
I don't think it is a big deal if this comment is not indented :-)

BoykoAlex: Returning as agreed: 
```
return maybeAutoFormat(entry, entry.withPrefix("\n#" + PROPS_MIGRATION_MESSAGE + entry.getPrefix()),
  ctx, getCursor().getParent());
```

pway99: Hi @BoykoAlex, after giving this a bit more thought, I agree with your original approach of not using autoFormat. Given that the Yaml model only supports comments within the prefix of an element, it's most likely not possible to auto-format comments.

pway99:
@BoykoAlex 
here is an example of how to add source files with a given path to your test source set that should help with the two remaining test failures.

https://github.com/openrewrite/rewrite-spring/blob/68141096cfbf585928102cbfc9066b677967bfc4/src/testWithSpringBoot_2_4/kotlin/org/openrewrite/java/spring/boot2/MergeBootstrapYamlWithApplicationYamlTest.kt#L31-L42

BoykoAlex:
Adjusted the PR

pway99:
Looks good thanks @BoykoAlex

PR #165, opened 2022-02-07
Boot version check fix. Annotation matching fix

BoykoAlex:
- Fix boot version check. Should be 2.4.x rather thaan 2.5.x
- `@SpringBootApplication` annotation matcher created not used by accident

### DIFF
--- a/src/main/java/org/openrewrite/java/spring/boot2/search/IntegrationSchedulerPoolRecipe.java
+++ b/src/main/java/org/openrewrite/java/spring/boot2/search/IntegrationSchedulerPoolRecipe.java
@@ -70,7 +70,7 @@ public class IntegrationSchedulerPoolRecipe extends Recipe {
     }
 
     private boolean isApplicableMavenProject(Maven maven) {
-        DependencyMatcher boot25Matcher = DependencyMatcher.build("org.springframework.boot:spring-boot:2.5.X").getValue();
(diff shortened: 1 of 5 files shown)


src/main/java/org/openrewrite/java/spring/boot2/search/IntegrationSchedulerPoolRecipe.java
```diff
@@ -169,9 +169,13 @@ public Properties visitFile(Properties.File file, ExecutionContext context) {
                 return (SourceFile) new JavaIsoVisitor<ExecutionContext>() {
                     @Override
                     public J.Annotation visitAnnotation(J.Annotation annotation, ExecutionContext context) {
-                        List<Comment> comments = new ArrayList<>(annotation.getComments());
-                        comments.add(new TextComment(false, GENERAL_MIGRATION_MESSAGE, "", Markers.EMPTY));
-                        return annotation.withComments(comments);
+                        J.Annotation a = super.visitAnnotation(annotation, context);
+                        if (annotationMatcher.matches(a)) {
+                            List<Comment> comments = new ArrayList<>(a.getComments());
```
jkschneider, line 174: @BoykoAlex I'd prefer we use `ListUtils.concat` or `ListUtils.concatAll` for these kinds of scenarios, which will allow us to swap out `ArrayList` later if we find a better performing implementation.

jkschneider:
Merged with https://github.com/openrewrite/rewrite-spring/commit/3dbdea859cce283ae6aae4fa976f106aac23bf5a.

PR #242, opened 2022-11-28
Adjust Boot 3 upgrade to use Boot 3 release version

BoykoAlex:


src/main/resources/META-INF/rewrite/spring-boot-30.yml
```diff
@@ -113,8 +113,8 @@ recipeList:
 ########################################################################################################################
 # SpringBoot 3.0 Configuration Changes
 type: specs.openrewrite.org/v1beta/recipe
-name: org.openrewrite.java.spring.boot3.SpringBootProperties_3_0_0
-displayName: Migrate Spring Boot properties to 3.0.0
+name: org.openrewrite.java.spring.boot3.SpringBootProperties_3_0
```
tkvangorder, line 116: Hi, @BoykoAlex Thanks for the PR. 

We recently started [automating our change log (of recipes](https://docs.openrewrite.org/changelog/7-33-0-release)) and it made us painfully aware that we are too often making changes to the actual recipe names which will break downstream projects that rely on those names.

We can change the displayName, as this won't impact our users, but can you please revert the actual name of the recipe back to `org.openrewrite.java.spring.boot3.SpringBootProperties_3_0_0` ?

We have had some internal discussions about adding recipe aliases to make this type of change safely, but we have not settled on an implementation.

src/main/resources/META-INF/rewrite/spring-boot-30.yml
```diff
@@ -33,7 +33,7 @@ recipeList:
   - org.openrewrite.java.spring.boot3.LegacyJmxExposure
   - org.openrewrite.java.spring.boot3.ActuatorEndpointSanitization
   - org.openrewrite.java.spring.boot3.Saml
-  - org.openrewrite.java.spring.boot3.SpringBootProperties_3_0_0
+  - org.openrewrite.java.spring.boot3.SpringBootProperties_3_0
```
tkvangorder, line 36: See comment below, we cannot make this change to the recipe name itself (without possibly breaking downstream users that are directly calling this recipe)

tkvangorder (changes requested):

src/main/resources/META-INF/rewrite/spring-boot-30.yml
BoykoAlex: Yes, makes sense. I reverted recipe id change :-)

tkvangorder (approved):
LGTM. Thank you!

src/main/resources/META-INF/rewrite/spring-boot-30.yml
BoykoAlex: Sounds good - revert this change

PR #243, opened 2022-11-28
Make field final when folding autowired field into constructor

BoykoAlex:
Feedback from Spring Team was make the field final


jkschneider:
Merged with 3c967b713d2a23ffd60412ca5f76b1c9947368e9. Thanks!

PR #245, opened 2022-11-29
No Repository annotation over Repository subclasses

BoykoAlex:


src/testWithSpringBoot_2_1/java/org/openrewrite/java/spring/NoRepoAnnotationOnRepoSubclassTest.java
```diff
@@ -0,0 +1,353 @@
...
+            """
+
+              public class MyRepo implements org.springframework.data.repository.Repository {
+              }
+              """
+          )
+        );
+    }
+
+    @Test
```
odrotbohm, line 54: This is not a valid test, actually. The annotation must only be removed from repository *interfaces* extending a Spring Data repository, not manual implementations of such. Otherwise, the implementation types are not registered as Spring beans anymore.

BoykoAlex: Aha, got it

BoykoAlex:
@odrotbohm it is great that you're watching over this work! I think I have corrected the implementation only to deal with interfaces instead. Please have a look at it when you get a few free minutes :-)

BoykoAlex:
Need to make another adjustment to this Recipe. Only `@Repository` without parameters can be removed.

BoykoAlex:
Should be good to go now

odrotbohm:
Looks good to me!

PR #249, opened 2022-12-05
Add recipe to move auto-configuration from spring.factories. Fixes #197

### LINKED ISSUE #197: 2.7 Changes to Auto-Configuration
https://docs.spring.io/spring-boot/docs/2.7.x/reference/html/howto.html#howto.testing.slice-tests

> **Auto-configuration Registration**
If you have created your own auto-configurations, you should move the registration from spring.factories to a new file named META-INF/spring/org.springframework.boot.autoconfigure.AutoConfiguration.imports. Each line contains the fully qualified name of the auto-configuration. See [the included auto-configurations](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-autoconfigure/src/main/resources/META-INF/spring/org.springframework.boot.autoconfigure.AutoConfiguration.imports) for an example.
> 
> For backwards compatibility, entries in spring.factories will still be honored.

So:
1. Any `src/main/resources/META-INF/spring.factories` file should move to `META-INF/spring/org.springframework.boot.autoconfigure.AutoConfiguration.imports`
2. the comma separate value for key `org.springframework.boot.autoconfigure.EnableAutoConfiguration` should be split across separate lines.

> **New @AutoConfiguration Annotation**
A new @AutoConfiguration annotation has been introduced. It should be used to annotate top-level auto-configuration classes that are listed in the new META-INF/spring/org.springframework.boot.autoconfigure.AutoConfiguration.imports file, replacing @Configuration. Configuration classes that are nested within or imported by an @AutoConfiguration class should continue to use @Configuration as before.
> 
> For your convenience, @AutoConfiguration also supports auto-configuration ordering via the after, afterNames, before and beforeNames attributes. This can be used as a replacement for @AutoConfigureAfter and @AutoConfigureBefore.

So:
1. Replace usage of `@Configuration` with `@AutoConfiguration` where classes are mentioned in the above replaced `src/main/resources/META-INF/spring.factories`

> **Test Slice Configuration**
> If you have created your own test-slices, you should move the registration from spring.factories to the new place under META-INF/spring/<name of your test slice annotation>.imports. The format is the same as the new file described in the "Auto-configuration Registration" section, see above.
> 
> See [the included test slices for an example](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-test-autoconfigure/src/main/resources/META-INF/spring/).

Above steps are complicated even more for users defining their own test slices, requiring a different filename with the same one class per line structure as above.

tkvangorder:


tkvangorder:
This recipe will move any auto-configuration classes from `spring.factories` (Under the key `org.springframework.boot.autoconfigure.EnableAutoConfiguration` to the corresponding `org.springframework.boot.autoconfigure.AutoConfiguration.imports` file.

The property entry will be removed from the spring.factories and the path of the corresponding imports file is computed relative to the `spring.factories`. If the import file already exists, this recipe will merge any non-duplicates into the import file. The fully-qualified names will be sorted in alpha-numeric order.

Additionally, this recipe will change the annotations on any of the fully-qualified configurations. It will replace `@Configuration` with `@AutoConfiguration`

PR #262, opened 2023-01-06
Add support for properties files

shanman190:


src/main/java/org/openrewrite/java/spring/data/UseTlsJdbcConnectionString.java
```diff
@@ -79,42 +84,165 @@ protected TreeVisitor<?, ExecutionContext> getVisitor() {
...
+
+            @Override
+            public Tree preVisit(Tree tree, ExecutionContext ctx) {
+                if (tree instanceof Yaml.Documents) {
+                    doAfterVisit(new UseTlsJdbcConnectionStringYaml(oldPort, port, validatedAttribute));
+                } else if (tree instanceof Properties.File) {
+                    doAfterVisit(new UseTlsJdbcConnectionStringProperties(oldPort, port, validatedAttribute));
+                }
+                return tree;
+            }
```
sambsnyd, line 109: Could simplify this a bit
```suggestion
            public @Nullable Tree visit(@Nullable Tree tree, ExecutionContext ctx) {
                if (tree instanceof Yaml.Documents) {
                    return new UseTlsJdbcConnectionStringYaml(oldPort, port, validatedAttribute).visitNonNull(tree, ctx);
                } else if (tree instanceof Properties.File) {
                    return new UseTlsJdbcConnectionStringProperties(oldPort, port, validatedAttribute).visitNonNull(tree, ctx);
                }
                return tree;
            }
```

sambsnyd (approved):

PR #272, opened 2023-01-19
ChangeSpringPropertyKey no longer expands entire affected yaml

nmck257:
Current state: if this recipe touches a yaml, then the entire file gets expanded to the indented syntax (no dot separation)

With this PR: yaml props modified by this recipe will always be dot-separated, but the rest of the file remains as it was

Ideally, we'd have some Style logic to attempt to maintain the source file's preferences, but in the interim, this PR should make the recipe more minimally-invasive


sambsnyd:
Thank you @nmck257 ! Less invasive changes is always better

PR #277, opened 2023-01-23
Update MySQL Driver artifactId & version number

### LINKED ISSUE #274: update mysql-driver coordinates for Spring Boot 2.7.8
Upgrading projects to the latest version of Spring Boot 2.7 (`2.7.8`) is broken at the moment due to a missing artifact adjustment of the mysql driver:

https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.7-Release-Notes#mysql-jdbc-driver

timtebeek:
Fixes #274 ; @martinlippert


timtebeek:
The `doNotPinWhenNotVersioned` test seems to have uncovered an imperfection when switching dependency groupId/artifactId: When you switch the version from the parent is not taken into account; that part is in https://github.com/openrewrite/rewrite, but could trip up users nonetheless. I'll log an issue there.

timtebeek:
@martinlippert any other feedback welcome; hope this solves the issue on your end.

martinlippert:
> @martinlippert any other feedback welcome; hope this solves the issue on your end.

Thanks @timtebeek for looking into this so quickly, very much appreciated. Will try out the latest snapshot once it is around and integrated on our side so that I can easily try this.

PR #279, opened 2023-01-31
Periodic regression test of Spring Petclinic migration

timtebeek:
Whenever I run a demo, I want to show a quick migration from Spring Petclinic branch 1.5.x to 2.x (maybe 3.x soon).
This time around I discovered a few [small](https://github.com/openrewrite/rewrite-testing-frameworks/issues/306) [imperfections](https://github.com/openrewrite/rewrite-spring/pull/277); with this periodic check I hope to catch those sooner.

What the regression test should ideally do:

1. checkout the [1.5.x branch](https://github.com/spring-projects/spring-petclinic/tree/1.5.x) of pet clinic
2. Upgrade that to Spring Boot 2.x
3. Update then to Java 17
4. Run Maven verify
5. Assert there's two test failures (or patch those just before)
6. Notify if there's any unexpected results

By using `-U org.openrewrite.maven:rewrite-maven-plugin:run` and `rewrite-spring:LATEST` this pipeline is not tied to any particular version of rewrite or rewrite-spring. That way it'll only signal issues after a new release of either rewrite or spring.

### DIFF
--- a/.github/workflows/regression-test.yml
+++ b/.github/workflows/regression-test.yml
@@ -16,7 +16,6 @@ jobs:
     steps:
       - uses: actions/checkout@v3
         with:
-          fetch-depth: 0
           name: spring-projects/spring-petclinic
           ref: 1.5.x
       - uses: actions/setup-java@v3
--- a/.github/workflows/regression-test.yml
+++ b/.github/workflows/regression-test.yml
@@ -16,6 +16,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
         with:
+          fetch-depth: 0
           name: spring-projects/spring-petclinic
           ref: refs/tags/1.5.x
       - uses: actions/setup-java@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/regression-test.yml
```diff
@@ -0,0 +1,42 @@
+---
+name: regression-test
+
+on: [push]
+#  workflow_dispatch: {}
+#  schedule:
+#    - cron: 0 18 * * *
```
timtebeek, line 7:
```suggestion
on:
  workflow_dispatch: {}
  schedule:
    - cron: 0 18 * * *
```

timtebeek:
Might need to switch this over to snapshots if we don't want to wait for new releases, and pick up issues sooner.

timtebeek:
Largely replaced by internal monitoring.

murdos:
@timtebeek : then https://github.com/openrewrite/rewrite/issues/1715 should probably be closed too, since it was similar, but with a larger scope to what you were proposing here.

timtebeek:
Ah perfect, thanks for pointing that one out!

PR #286, opened 2023-02-14
Use the new requestMatchers with Spring Security 5.8 - partially resolves #255

### LINKED ISSUE #255: Migration Recipe for Spring Security 5.8
The Spring Security team has produced a [migration guide for Spring Security 5.8](https://docs.spring.io/spring-security/reference/5.8/migration/index.html) that includes the following: 

- [ ] [Update Pbkdf2PasswordEncoder](https://docs.spring.io/spring-security/reference/5.8/migration/index.html#_update_pbkdf2passwordencoder)
- [ ] [Update SCryptPasswordEncoder](https://docs.spring.io/spring-security/reference/5.8/migration/index.html#_update_scryptpasswordencoder)
- [ ] [Update Argon2PasswordEncoder](https://docs.spring.io/spring-security/reference/5.8/migration/index.html#_update_argon2passwordencoder)
- [ ] Use delegating PasswordEncoder for Pbkdf2PasswordEncoder
- [ ] Use delegating PasswordEncoder for SCryptPasswordEncoder
- [ ] Use delegating PasswordEncoder for Argon2PasswordEncoder
- [ ] [Search Recipe for Encryptors.queryableText(CharSequence,CharSequence)](https://docs.spring.io/spring-security/reference/5.8/migration/index.html#_stop_using_encryptors_queryabletext)
- [ ] [Require Explicit Saving of SecurityContextRepository](https://docs.spring.io/spring-security/reference/5.8/migration/servlet/session-management.html#_require_explicit_saving_of_securitycontextrepository)
- [ ] [Change HttpSessionSecurityContextRepository to DelegatingSecurityContextRepository](https://docs.spring.io/spring-security/reference/5.8/migration/servlet/session-management.html#_change_httpsessionsecuritycontextrepository_to_delegatingsecuritycontextrepository)
- [ ] [Use `DeferredSecurityContext loadDeferredContext(HttpServletRequest request)`](https://docs.spring.io/spring-security/reference/5.8/migration/servlet/session-management.html#_address_securitycontextrepository_deprecations)
- [ ] [Optimize Querying of RequestCache](https://docs.spring.io/spring-security/reference/5.8/migration/servlet/session-management.html#requestcache-query-optimization)
- [ ] [Require Explicit Invocation of SessionAuthenticationStrategy](https://docs.spring.io/spring-security/reference/5.8/migration/servlet/session-management.html#_require_explicit_invocation_of_sessionauthenticationstrategy)
- [ ] [Defer Loading CsrfToken](https://docs.spring.io/spring-security/reference/5.8/migration/servlet/exploits.html#_defer_loading_csrftoken)
- [ ] [Protect against CSRF Breach](https://docs.spring.io/spring-security/reference/5.8/migration/servlet/exploits.html#_protect_against_csrf_breach)
- [ ] [Use the new requestMatchers methods](https://docs.spring.io/spring-security/reference/5.8/migration/servlet/config.html#use-new-requestmatchers) 
- [ ] [Use the new securityMatchers methods](https://docs.spring.io/spring-security/reference/5.8/migration/servlet/config.html#use-new-security-matchers)
- [x] [Publish a SecurityFilterChain Bean](https://docs.spring.io/spring-security/reference/5.8/migration/servlet/config.html#_publish_a_securityfilterchain_bean)
- [x] [Publish a WebSecurityCustomizer Bean](https://docs.spring.io/spring-security/reference/5.8/migration/servlet/config.html#_publish_a_websecuritycustomizer_bean)
- [ ] [Publish an AuthenticationManager Bean](https://docs.spring.io/spring-security/reference/5.8/migration/servlet/config.html#_publish_an_authenticationmanager_bean)
- [ ] [https://docs.spring.io/spring-security/reference/5.8/migration/servlet/config.html#_add_configuration_to_enable_annotations](https://docs.spring.io/spring-security/reference/5.8/migration/servlet/config.html#_add_configuration_to_enable_annotations)
- [ ] [Use SHA-256 in Remember Me](https://docs.spring.io/spring-security/reference/5.8/migration/servlet/authentication.html#servlet-opt-in-sha256-rememberme)
- [ ] [Propagate AuthenticationServiceExceptions](https://docs.spring.io/spring-security/reference/5.8/migration/servlet/authentication.html#_propagate_authenticationserviceexceptions)
- [ ] [Replace global method security with method security](https://docs.spring.io/spring-security/reference/5.8/migration/servlet/authorization.html#servlet-replace-globalmethodsecurity-with-methodsecurity)
- [ ] [Publish a MethodSecurityExpressionHandler instead of a PermissionEvaluator](https://docs.spring.io/spring-security/reference/5.8/migration/servlet/authorization.html#servlet-replace-permissionevaluator-bean-with-methodsecurityexpression-handler)
- [ ] [Replace any custom method-security AccessDecisionManagers](https://docs.spring.io/spring-security/reference/5.8/migration/servlet/authorization.html#_replace_any_custom_method_security_accessdecisionmanagers)
- [ ] [Ensure all messages have defined authorization rules](https://docs.spring.io/spring-security/reference/5.8/migration/servlet/authorization.html#_ensure_all_messages_have_defined_authorization_rules)
- [ ] [Add @EnableWebSocketSecurity](https://docs.spring.io/spring-security/reference/5.8/migration/servlet/authorization.html#_add_enablewebsocketsecurity)
- [ ] [Use an AuthorizationManager<Message<?>> instance](https://docs.spring.io/spring-security/reference/5.8/migration/servlet/authorization.html#_use_an_authorizationmanagermessage_instance)
- [ ] [Stop Implementing AbstractSecurityWebSocketMessageBrokerConfigurer](https://docs.spring.io/spring-security/reference/5.8/migration/servlet/authorization.html#_stop_implementing_abstractsecuritywebsocketmessagebrokerconfigurer)
- [ ] [Use AuthorizationManager for Request Security](https://docs.spring.io/spring-security/reference/5.8/migration/servlet/authorization.html#_use_authorizationmanager_for_request_security)
- [ ] [Switch to AuthorizationManager](https://docs.spring.io/spring-security/reference/5.8/migration/servlet/authorization.html#_switch_to_authorizationmanager)
- [ ] [Switch to filter all dispatcher types](https://docs.spring.io/spring-security/reference/5.8/migration/servlet/authorization.html#switch-filter-all-dispatcher-types)
- [ ] [Permit FORWARD when using Spring MVC](https://docs.spring.io/spring-security/reference/5.8/migration/servlet/authorization.html#_permit_forward_when_using_spring_mvc)
- [ ] [Replace any custom filter-security AccessDecisionManagers](https://docs.spring.io/spring-security/reference/5.8/migration/servlet/authorization.html#_replace_any_custom_filter_security_accessdecisionmanagers)
- [ ] [Change Default oauth2Login() Authorities](https://docs.spring.io/spring-security/reference/5.8/migration/servlet/oauth2.html#_change_default_oauth2login_authorities)
- [ ] [Address OAuth2 Client Deprecations](https://docs.spring.io/spring-security/reference/5.8/migration/servlet/oauth2.html#_address_oauth2_client_deprecations)
- [ ] [Use OpenSAML 4](https://docs.spring.io/spring-security/reference/5.8/migration/servlet/saml2.html#_use_opensaml_4)
- [ ] [Use OpenSaml4AuthenticationProvider](https://docs.spring.io/spring-security/reference/5.8/migration/servlet/saml2.html#_use_opensaml4authenticationprovider)
- [ ] [Stop Using SAML 2.0 Converter constructors](https://docs.spring.io/spring-security/reference/5.8/migration/servlet/saml2.html#_stop_using_saml_2_0_converter_constructors)
- [ ] [Change to Using Saml2AuthenticationRequestResolver](https://docs.spring.io/spring-security/reference/5.8/migration/servlet/saml2.html#_change_to_using_saml2authenticationrequestresolver)
- [ ] [Use the latest Saml2AuthenticationToken constructor](https://docs.spring.io/spring-security/reference/5.8/migration/servlet/saml2.html#_use_the_latest_saml2authenticationtoken_constructor)
- [ ] [Use RelyingPartyRegistration updated methods](https://docs.spring.io/spring-security/reference/5.8/migration/servlet/saml2.html#_use_relyingpartyregistration_updated_methods)
- [ ] [Exploit Protection Migrations](https://docs.spring.io/spring-security/reference/5.8/migration/reactive.html#_exploit_protection_migrations)
- [ ] [Use AuthorizationManager for Method Security](https://docs.spring.io/spring-security/reference/5.8/migration/reactive.html#_use_authorizationmanager_for_method_security)
- [ ] [Propagate AuthenticationServiceExceptions](https://docs.spring.io/spring-security/reference/5.8/migration/reactive.html#_propagate_authenticationserviceexceptions)
- [ ] [Address OAuth2 Client Deprecations](https://docs.spring.io/spring-security/reference/5.8/migration/reactive.html#_address_oauth2_client_deprecations)

jjank:
(Related discussion [in Slack](https://rewriteoss.slack.com/archives/C01A843MWG5/p1674752673977169) )

Hi there 😃 

I finally had time to work on my first recipe! See the related [Spring Security Docs](https://docs.spring.io/spring-security/reference/5.8/migration/servlet/config.html#use-new-requestmatchers) for background

Before I continue I'd like to get some quick feedback if this generally goes into the right direction. A couple of remarks:

* The gradle build obviously is just copy & paste atm but I wanted to get it working for now. This needs some love before merging
* Generally: Should I put everything in one recipe i.e. one visitor? Specifically `mvcMatchers` and `regexMatchers` need to be replaced in the same fashion. `if else`-ing seems a bit repetitive but it certainly will work. If you have any ideas or preferences please let me know
* The test "only fails due to a single line difference:
Screenshot 2023-02-14 at 23 37 28

But strangely when I try to add or remove a line to match actual and expected I get a type failure in the AST:

```
java.lang.IllegalStateException: AST contains missing or invalid type information
MethodInvocation->MethodInvocation->MethodInvocation->MethodInvocation->MethodInvocation->MethodInvocation->Lambda->MethodInvocation->Block->MethodDeclaration->Block->ClassDeclaration->CompilationUnit
/*~~(type information has a different method name 'antMatchers')~~>*/authz
                        .requestMatchers("/api/admin/**")
```


jjank:
I've updated the test so that it fails because of the invalid type - need to figure that one out though....

timtebeek:
Hi @jjank ; thanks for your work so far! I've had a look at it locally, but was not able to quickly fix the recipe yet. What you're seeing is that your recipe makes the changes needed to have the output match the expected output in text, but not yet at the type level.

One of the things I'm wondering about is if you actually need the JavaTemplate in this case; maybe you do and I'm just not doing enough recipe development lately, but had you looked into using `org.openrewrite.java.ChangeMethodName` for instance? That could potentially help, although it seems the response types of the method will also change, so maybe not.

Right now you have one test that handles a few different replacements; perhaps it's easier if you add a specific test for each use case, such that when you fix one case, you'll see the corresponding test pass. Also your test seems to have a lot of wrapping setup that does show how it would be used in practice, but is maybe not needed to test the relevant replacements. Reducing that test setup to just enough for a call on a `AuthorizationManagerRequestMatcherRegistry` might make sense, as it will also allow you to more easily debug and step through your visitor. Optionally you can leave one integration test like this to show the bigger context and ensure we remain compatible there as well.

BoykoAlex:
@jjank I wouldn't use java template for this kind of code modification. Java template is meant for something more complex.
I'd match the method invocation and then once matched:
1. Get the MethodType from the invocation.
2. Look at the declaration type of the MethodType from 1 and try to find in the list of methods method `requestMatchers` with the same parameter types in the MethodType from 1.
3. If such method type is found then `return methodInvocation.withMethodType(foundMethodTypeFrom2).withName(methodInvocation.getName().withSimpleName("requestMatchers");

jjank:
@BoykoAlex and @timtebeek thanks for the hints and suggestions (for some reason I did not get any GH notfications 🙈 ) 

I'll try that as soon as I can 👍

jjank:
@timtebeek  and @BoykoAlex 

So I go all the test cases to pass - thank you very much for your help 😄 

Before I lose myself in refactoring I have a couple of questions:
* I don't know if it is worth testing if somebody _may_ have overridden `antMatchers` or `regexMatchers` I haven't seen that but I hope you can say if that is a common thing.
* I'm generally unsure how to handle nulls and/or validation. I've coded in some `requireNonNull` but I don't know if I need to be more defensive here.
* To the same point: `.orElse(null)` is not exactly pretty - I know. I haven't seen a consistent pattern across rewrite how you guys want to handle this

Any feedback and/or suggestions are highly appreciated 👍

jjank:
Another question:

How should this recipe be integrated? I was looking into `spring-boot-30.yml` but then I wasn't sure: 
* Can a recipe successfully run if the respective methods (`mvcMatchers`& friends) already have been removed?
* Conversely: Can a recipe successfully run if the new methods (`requestMatchers`)  do not exist yet? 

I _guess_ the answer to both is no? 

That would be unfortunate for end users. Because just checking start.spring.io real quick: Boot 2.7.9 "only" ships with spring-security 5.7.7. and [requestMatchers(..) was only added in 5.8](https://docs.spring.io/spring-security/site/docs/5.8.x/api/org/springframework/security/config/annotation/web/AbstractRequestMatcherRegistry.html#requestMatchers(org.springframework.http.HttpMethod)).

timtebeek:
Hi @jjank Good to see you got the tests working already with the above feedback! Thanks for that. :)

I've pushed a few minor commits just now, before signing off for today. Most importantly I've added an override of `getSingleSourceApplicableTest`, which will limit the recipe visitor calls to classes that actually use those methods. That helps both performance and the what preconditions are met when executing the visitor. You can similarly add more conditionals to for instance limit on the availability of Spring Security 5.8+ (I think).

Before we hook this into `spring-boot-30.yml` it might be good to test this on a few projects in the wild, and only when confirmed working on of those do we hook it into the chained upgrade.

Hope to come back to this review later. @BoykoAlex welcome to chime in before then!

src/main/java/org/openrewrite/java/spring/security5/UseNewRequestMatchers.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    @Nullable
+    private JavaType.Method findRequestMatchersMethodWithMatchingParameterTypes(J.MethodInvocation mi) {
+        JavaType.Method methodType = requireNonNull(mi.getMethodType(), "methodType");
+        List<JavaType> parameterTypes = methodType.getParameterTypes();
+        List<JavaType.Method> methods;
+        boolean isOverride = TypeUtils.isOverride(mi.getMethodType());
+        if (isOverride) {
+            methods = requireNonNull(methodType.getDeclaringType().getSupertype(), "superType").getMethods();
```
BoykoAlex, line 100: Not sure you'd need to look for overrides... The methods you're matching and the replacement method all declared in the same type... Therefore, line 102 alone should handle everything correctly... Don't think you'd need to go up to the super type from the declaring type

jjank: I thought so too but without this I could not find a `requestMatchers` method in the `mvcMatchers` cases.

In my head it makes sense because: `mvcMatchers` [is abstract](https://docs.spring.io/spring-security/site/docs/5.8.x/api/org/springframework/security/config/annotation/web/AbstractRequestMatcherRegistry.html#mvcMatchers(java.lang.String...)) so any methodInvocation will always be on a subclass of `AbstractRequestMatcherRegistry`.

Without looking at the supertype I could not find the `requestMatchers` method.

src/main/java/org/openrewrite/java/spring/security5/UseNewRequestMatchers.java
```diff
@@ -0,0 +1,110 @@
...
+                    .withMethodType(requestMatchersMethod)
+                    .withName(mi.getName().withSimpleName("requestMatchers"));
+        } else {
+            return mi;
+        }
+    }
+
+    @Nullable
+    private JavaType.Method findRequestMatchersMethodWithMatchingParameterTypes(J.MethodInvocation mi) {
+        JavaType.Method methodType = requireNonNull(mi.getMethodType(), "methodType");
```
timtebeek, line 95: I think the method matchers already guard against a null `methodType` here; and even if we do get a null methodType, this recipe specifically is for Spring Boot 3, which requires Java 17, which will give us helpful null pointer exceptions further down. I may be wrong of course, but given the above I don't think there's much value in the explicit validation.
```suggestion
        JavaType.Method methodType = mi.getMethodType();
```

src/main/java/org/openrewrite/java/spring/security5/UseNewRequestMatchers.java
```diff
@@ -0,0 +1,110 @@
...
+        if (isOverride) {
+            methods = requireNonNull(methodType.getDeclaringType().getSupertype(), "superType").getMethods();
+        } else {
+            methods = methodType.getDeclaringType().getMethods();
+        }
+        return methods.stream()
+                .filter(m -> m.getName().equals("requestMatchers"))
+                .filter(m -> m.getParameterTypes().equals(parameterTypes))
+                .findFirst()
+                .orElse(null);
```
timtebeek, line 108: In terms of method signature it might be preferable to return an Optional indeed; slightly more expressive.

timtebeek:
> Another question:
> 
> How should this recipe be integrated? I was looking into `spring-boot-30.yml` but then I wasn't sure:
> 
> * Can a recipe successfully run if the respective methods (`mvcMatchers`& friends) already have been removed?

Ideally any recipe can repeatedly run without issues, typically by checking the preconditions through `getSingleSourceApplicableTest`. Do you see any other such checks we might need to add?

> * Conversely: Can a recipe successfully run if the new methods (`requestMatchers`)  do not exist yet?

If we hook this into the Spring Boot 3 migration, then the required methods should be available right? Then only if people explicitly run this recipe without that dependency present might they run into issues.

Good that you're thinking of these concerns yourself as well! Helps ensure users have good experience regardless of where they are in their migration.

src/main/java/org/openrewrite/java/spring/security5/UseNewRequestMatchers.java
jjank: It really is a matter of preference I guess. If you were to be picky the recipe is not specifically for Boot 3, so JDK17 is not a given.

But realistically: In 98% it _will_ be a Boot 3 migration

jjank:
> Ideally any recipe can repeatedly run without issues, typically by checking the preconditions through getSingleSourceApplicableTest. Do you see any other such checks we might need to add?

Is it possible to check the availability of a dependency `spring-security-config:5.8.+` in `getSingleSourceApplicableTest()`? I couldn't find any examples 🙈 

> If we hook this into the Spring Boot 3 migration, then the required methods should be available right?

That seems to be the trouble:
* `requestMatchers` was [first added in `spring-security:5:8` ](https://docs.spring.io/spring-security/site/docs/5.8.x/api/org/springframework/security/config/annotation/web/AbstractRequestMatcherRegistry.html#requestMatchers(java.lang.String...))
* At the same time: `mvcMatchers`, `antMatchers` and `requestMatchers` were [removed in 6.0](https://docs.spring.io/spring-security/site/docs/6.0.x/api/org/springframework/security/config/annotation/web/AbstractRequestMatcherRegistry.html#requestMatchers(java.lang.String...))

So unless I missed something, there's only a _tiny_ window when this recipe acutally will work: 
`5.8 <=  spring-security < 6.0`

src/main/java/org/openrewrite/java/spring/security5/UseNewRequestMatchers.java
jjank: Resolved in https://github.com/openrewrite/rewrite-spring/pull/286/commits/28748e9c764026f47ebb3e4a9f8b4494c52dd30c

timtebeek:
> > Ideally any recipe can repeatedly run without issues, typically by checking the preconditions through getSingleSourceApplicableTest. Do you see any other such checks we might need to add?
> 
> Is it possible to check the availability of a dependency `spring-security-config:5.8.+` in `getSingleSourceApplicableTest()`? I couldn't find any examples see_no_evil

Not that I know of either, but I don't think it's a problem here. See below.

> 
> > If we hook this into the Spring Boot 3 migration, then the required methods should be available right?
> 
> That seems to be the trouble:
> 
> * `requestMatchers` was [first added in `spring-security:5:8` ](https://docs.spring.io/spring-security/site/docs/5.8.x/api/org/springframework/security/config/annotation/web/AbstractRequestMatcherRegistry.html#requestMatchers(java.lang.String...))
> * At the same time: `mvcMatchers`, `antMatchers` and `requestMatchers` were [removed in 6.0](https://docs.spring.io/spring-security/site/docs/6.0.x/api/org/springframework/security/config/annotation/web/AbstractRequestMatcherRegistry.html#requestMatchers(java.lang.String...))
> 
> So unless I missed something, there's only a _tiny_ window when this recipe acutally will work: `5.8 <= spring-security < 6.0`

What we could look into is [having the recipe use a different parserClasspath](https://docs.openrewrite.org/authoring-recipes/multiple-versions). The documentation highlights that for tests specifically, but we have some [examples in rewrite-testing-frameworks](https://github.com/openrewrite/rewrite-testing-frameworks/blob/ad45570928ea62bb79301196602791234ea14109/src/main/java/org/openrewrite/java/testing/junit5/UseTestMethodOrder.java#L68) where we use that in src/main visitors as well. If you set that to use 5.8+ then the recipe should be able to find both the old and the new methods, and migrate between the two.

Unless anyone has any other ideas I think this would be the way to go. Then we should be able to hook in the migration anywhere on the upgrade chain; either when you want to adopt Spring Security 5.8 explicitly, or when moving from 5.7 to 6.0.

jjank:
Hi Tim,

thanks for the pointers. I'll have a look at that. Since I don't use `JavaTemplate` to change the method invocation I'm not sure if your suggestion can be applied, but I'll dig into it a little bit 👍

jjank:
@timtebeek I've refactored the recipe to use the [recipe dependencies](https://docs.openrewrite.org/authoring-recipes/multiple-versions) (at least for the tests).

But I couldn't figure out how to incorporate a `javaParser` without using a `JavaTemplate` (which would make the recipe much more complex).

Ideally this should be done in `getSingleSourceApplicableTest()` correct - I'll ask in Slack

sambsnyd:
Thanks @jjank !

PR #287, opened 2023-02-15
Flatten configuration class if possible

### LINKED ISSUE #278: Receipe for WebSecurityConfigurerAdapter deprecation not working if multiple such classes are defined
I used the `org.openrewrite.java.spring.boot2.WebSecurityConfigurerAdapter` receipe for an application that uses multiple http security. You can view an example of such a setup in [the docs of an older version of Spring Security](https://docs.spring.io/spring-security/site/docs/5.0.0.RC1/reference/htmlsingle/#multiple-httpsecurity):

```
@EnableWebSecurity
public class MultiHttpSecurityConfig {
	@Bean                                                             
	public UserDetailsService userDetailsService() throws Exception {
		// ensure the passwords are encoded properly
		UserBuilder users = User.withDefaultPasswordEncoder();
		InMemoryUserDetailsManager manager = new InMemoryUserDetailsManager();
		manager.createUser(users.username("user").password("password").roles("USER").build());
		manager.createUser(users.username("admin").password("password").roles("USER","ADMIN").build());
		return manager;
	}

	@Configuration
	@Order(1)                                                        
	public static class ApiWebSecurityConfigurationAdapter extends WebSecurityConfigurerAdapter {
		protected void configure(HttpSecurity http) throws Exception {
			http
				.antMatcher("/api/**")                               
				.authorizeRequests()
					.anyRequest().hasRole("ADMIN")
					.and()
				.httpBasic();
		}
	}

	@Configuration                                                   
	public static class FormLoginWebSecurityConfigurerAdapter extends WebSecurityConfigurerAdapter {

		@Override
		protected void configure(HttpSecurity http) throws Exception {
			http
				.authorizeRequests()
					.anyRequest().authenticated()
					.and()
				.formLogin();
		}
	}
}
```

The [newest version of the docs](https://docs.spring.io/spring-security/reference/5.7.0/servlet/configuration/java.html#_multiple_httpsecurity) shows how this should be done now:

```
@EnableWebSecurity
public class MultiHttpSecurityConfig {
	@Bean                                                             
	public UserDetailsService userDetailsService() throws Exception {
		// ensure the passwords are encoded properly
		UserBuilder users = User.withDefaultPasswordEncoder();
		InMemoryUserDetailsManager manager = new InMemoryUserDetailsManager();
		manager.createUser(users.username("user").password("password").roles("USER").build());
		manager.createUser(users.username("admin").password("password").roles("USER","ADMIN").build());
		return manager;
	}

	@Bean
	@Order(1)                                                        
	public SecurityFilterChain apiFilterChain(HttpSecurity http) throws Exception {
		http
			.antMatcher("/api/**")                                   
			.authorizeHttpRequests(authorize -> authorize
				.anyRequest().hasRole("ADMIN")
			)
			.httpBasic(withDefaults());
		return http.build();
	}

	@Bean                                                            
	public SecurityFilterChain formLoginFilterChain(HttpSecurity http) throws Exception {
		http
			.authorizeHttpRequests(authorize -> authorize
				.anyRequest().authenticated()
			)
			.formLogin(withDefaults());
		return http.build();
	}
}
```

The receipe currently keeps the inner classes. The following logic should be implemented:

* The inner classes should be removed
* The `configure` methods that are now renamed to `filterChain` should be removed to something unique to avoid having multiple beans with the same name.
* Any `@Order` or `@Profile` annotations that are on the inner classes should be moved to the `filterChain` methods.
* In my cases, one of those inner classes also had a constructor that allowed to inject `Environment`. I had to change the `filterChain` method to also have this `Environment` as an extra parameter next to `HttpSecurity`. Not sure if that is feasible to automate, seems a bit complex.

BoykoAlex:
Fixes #278 

We cannot solve `@Order` annotation port on the bean methods. However, if it is important we can have a white list of annotations that if present over the configuration class can be moved over bean methods of that class.

### DIFF
--- a/src/testWithSpringBoot_2_4/java/org/openrewrite/java/spring/boot2/WebSecurityConfigurerAdapterTest.java
+++ b/src/testWithSpringBoot_2_4/java/org/openrewrite/java/spring/boot2/WebSecurityConfigurerAdapterTest.java
@@ -32,7 +32,7 @@ class WebSecurityConfigurerAdapterTest implements RewriteTest {
     public void defaults(RecipeSpec spec) {
         spec.recipe(new WebSecurityConfigurerAdapter())
           .parser(JavaParser.fromJavaVersion()
-            .classpath("spring-beans", "spring-context", "spring-boot", "spring-security", "spring-web", "tomcat-embed"));
+            .classpath("spring-beans", "spring-context", "spring-boot", "spring-security", "spring-web", "tomcat-embed", "spring-core"));
     }
 
     @Test
@@ -78,7 +78,7 @@ class WebSecurityConfigurerAdapterTest implements RewriteTest {
               public class SecurityConfiguration {
               
                   @Bean
-                  SecurityFilterChain filterChain(HttpSecurity http) throws Exception {
+                  SecurityFilterChain securityConfigurationSecurityFilterChain(HttpSecurity http) throws Exception {
                       http
                           .authorizeHttpRequests((authz) -> authz
                               .anyRequest().authenticated()
@@ -153,7 +153,7 @@ class WebSecurityConfigurerAdapterTest implements RewriteTest {
               public class SecurityConfiguration {
 
                   @Bean
-                  WebSecurityCustomizer webSecurityCustomizer() {
+                  WebSecurityCustomizer securityConfigurationWebSecurityCustomizer() {
                       return (web) -> {
                           web.ignoring().antMatchers("/ignore1", "/ignore2");
                       };
@@ -325,7 +325,7 @@ class WebSecurityConfigurerAdapterTest implements RewriteTest {
               public class SecurityConfiguration {
                             
                   @Bean
-                  SecurityFilterChain filterChain(HttpSecurity http) {
+                  SecurityFilterChain securityConfigurationSecurityFilterChain(HttpSecurity http) {
                       System.out.println(getApplicationContext());
                       http
                           .authorizeHttpRequests((authz) -> authz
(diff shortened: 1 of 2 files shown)


sambsnyd:
@BoykoAlex can you fix the indentation of the test case where the `after` text is misaligned?

src/testWithSpringBoot_2_4/java/org/openrewrite/java/spring/boot2/WebSecurityConfigurerAdapterTest.java
```diff
@@ -78,7 +78,7 @@ void someMethod() {}
               public class SecurityConfiguration {
               
                   @Bean
-                  SecurityFilterChain filterChain(HttpSecurity http) throws Exception {
+                  SecurityFilterChain securityConfigurationSecurityFilterChain(HttpSecurity http) throws Exception {
```
wimdeblauwe, line 81: Is it possible to keep the `filterChain` method name if there is only 1 configuration class? You need the different names if you have multiple, but if there is only one, then using `filterChain` would be nicer I think.

BoykoAlex:
@sambsnyd thanks for the feedback :-) I have pushed a commit to address your comments.

BoykoAlex:
(BTW I'm making more changes to address in-memory, ldap and jdbc authentication. Started with in-memory - this will include quite a bit of cleanup for this recipe)

sambsnyd:
Thanks @BoykoAlex ! Feel free to submit all of those improvements in a subsequent PR, this looked good enough to accept

PR #290, opened 2023-02-21
Security DSL

BoykoAlex:
Convert method invocation http on `HttpSecurity` and `ServerHttpSecurity` to the new Lambda DSL syntax. See [Spring Security - Lambda DSL](https://spring.io/blog/2019/11/21/spring-security-lambda-dsl)

### DIFF
--- a/src/main/java/org/openrewrite/java/spring/boot2/ConvertToSecurityDslVisitor.java
+++ b/src/main/java/org/openrewrite/java/spring/boot2/ConvertToSecurityDslVisitor.java
@@ -1,0 +1,178 @@
+/*
+ * Copyright 2023 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.java.spring.boot2;
+
+import org.openrewrite.Cursor;
+import org.openrewrite.Tree;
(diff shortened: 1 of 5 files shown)


BoykoAlex:
Gradle `build.kts` file changes need to be sorted out. The purpose of the changes bring source compatibility to Java 11. Had to put 17 to make the tests happy. I'm nowhere near a Gradle expert and the change was purely to make things compile.

src/main/java/org/openrewrite/java/spring/boot2/ConvertToSecurityDslVisitor.java
```diff
@@ -0,0 +1,188 @@
...
+
+    @Override
+    public Space visitSpace(Space space, Space.Location loc, P executionContext) {
+        return space;
+    }
+
+    @Override
+    public Markers visitMarkers(Markers markers, P executionContext) {
+        return markers;
+    }
```
sambsnyd, line 51:
```suggestion
```

Unless I'm missing something this looks unneeded. Perhaps code which was used as a convenient place to hang a breakpoint during debugging?

BoykoAlex: I have removed those 2 methods.

However, I should probably explain why I add these...
I add those two methods on the IDE side for visitors that mark CUs for various issues. Those visitrs meant to be very quick hence I override visit space and markers to prune these AST branches as what I'm looking for isn't there. It just seem to win a a few ms for the visitor... Looks like space can contain comments, some of these comments are javadoc comments... also provenance-kind markers seem to be attached to all AST nodes (or many if not all and pruning markers seem to to help

sambsnyd: It makes sense that skipping parsing javadocs would save time. I wonder if we should make visiting javadoc comments opt-in for visitors, as most aren't going to be interested. Not sure what you mean about the markers, though. Most tree elements have no markers whatsoever. It is only the root element which has provenance markers on it.

BoykoAlex: @sambsnyd it is the positions markers not provenance. Those markers are everywhere for me as I care about line/column positional data...

PR #293, opened 2023-02-23
Fix unit test

BoykoAlex:
New boot patch version became available: 2.7.9

### DIFF
--- a/src/testWithSpringBoot_2_7/java/org/openrewrite/java/spring/boot2/UpdateMysqlDriverArtifactIdTest.java
+++ b/src/testWithSpringBoot_2_7/java/org/openrewrite/java/spring/boot2/UpdateMysqlDriverArtifactIdTest.java
@@ -110,7 +110,7 @@ public class UpdateMysqlDriverArtifactIdTest implements RewriteTest {
                 <parent>
                   <groupId>org.springframework.boot</groupId>
                   <artifactId>spring-boot-starter-parent</artifactId>
-                  <version>2.7.8</version>
+                  <version>2.7.9</version>
                   <relativePath/> <!-- lookup parent from repository -->
                 </parent>
                 <dependencies>


src/testWithSpringBoot_2_7/java/org/openrewrite/java/spring/boot2/UpdateMysqlDriverArtifactIdTest.java
```diff
@@ -110,7 +110,7 @@ void doNotPinWhenNotVersioned() {
                 <parent>
                   <groupId>org.springframework.boot</groupId>
                   <artifactId>spring-boot-starter-parent</artifactId>
-                  <version>2.7.8</version>
+                  <version>2.7.9</version>
```
timtebeek, line 113: We might want to assert these version numbers differently, [as seen here for Axon](https://github.com/AxonFramework/AxonFramework/blob/axon-4.7.x/migration/src/test/java/org/axonframework/migration/AxonJakartaTest.java#L88).
That way we don't have to update this test anymore after new releases of Spring 2.7.x.
Not saying you should adapt the tests like that here; just sharing that knowledge and can potentially pick this up tomorrow.

BoykoAlex: My goal with this PR was to get the **rewrite-spring** snapshot build through with the latest changes such that I could push my STS changes.

timtebeek: Would not want to block you on that! I've merged this one as is then, sorry for the hold up; I'd imagined you'd be off the bed already as well (as I should be).

PR #294, opened 2023-02-23
Add @Configuration over a class having @Bean methods

BoykoAlex:
(Kind of a consequence from `@EnableXXXSEcurity` annotations)

If a class has methods annotated with `@Beans` or meta-subclasses of `@Bean` a declaring class likely needs to have `@Configuration` or some meta-subclass annotation. If it is missing the recipe would add `@Configuration`.
Useful if someone upgraded to Boot 3 and Security 6 manually without adding missing `@Configuration` next to `@EnableXXXSecurity` then this recipe would be useful.

I think it'll be useful in the IDE world mostly for those who forgot `@Configuration` in a class defining beans.

### DIFF
--- a/src/main/java/org/openrewrite/java/spring/boot2/AddConfigurationAnnotationIfBeansPresent.java
+++ b/src/main/java/org/openrewrite/java/spring/boot2/AddConfigurationAnnotationIfBeansPresent.java
@@ -34,8 +34,14 @@ import java.util.Comparator;
 public class AddConfigurationAnnotationIfBeansPresent extends Recipe {
 
     private static final String FQN_BEAN = "org.springframework.context.annotation.Bean";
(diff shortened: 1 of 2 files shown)


timtebeek:
@BoykoAlex Thanks again! Hope you agree with the minor touch ups in https://github.com/openrewrite/rewrite-spring/pull/294/commits/6dcdc4e690716d71d51ab008edd20ca542dd748f; if you do I'll go ahead and merge this.

timtebeek (approved):

BoykoAlex:
@timtebeek I'd like to have public access to `boolean isApplicableClass(J.ClassDeclaration classDecl, Cursor cursor)` to use it on the IDE side for marking problems. Are you okay with me moving that method to the the recipe class and making it public static?

timtebeek:
Ah sorry I'd missed that! That'll undo most of my changes I suppose. Go ahead!

PR #297, opened 2023-02-26
Add recipe to consolidate @ExtendWith and @ContextConfiguration

### LINKED ISSUE #296: @ExtendWith and @ContextConfguration can be simplified to @SpringJUnitConfig
Spring introduced the [`@SpringJUnitConfig`](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/test/context/junit/jupiter/SpringJUnitConfig.html) annotation as a composed annotation for `@ExtendWith(SpringExtension.class)` and `@ContextConfiguration`. A recipe should be created to perform this migration.

The recipe should transform the following code:

    @ExtendWith(SpringExtension.class)
    @ContextConfiguration(classes = ExampleConfiguration.class)

Into:

    @SpringJUnitConfig(classes = ExampleConfiguration.class)

Note that `@SpringJUnitConfig` contains aliases for the following `@ContextConfiguration` attributes which should be maintained from the original `@ContextConfiguration` annotation when creating the `@SpringJUnitConfig` annotation:

- value
- classes
- locations
- initializers
- inheritLocations
- inheritInitializers
- name

It would be beneficial to also include this in the [migration from JUnit4 to JUnit5 recipe](https://docs.openrewrite.org/reference/recipes/java/spring/boot2/springboot2junit4to5migration), possibly as another step of the recipe that [removes unnecessary `@RunWith` annotations](https://docs.openrewrite.org/reference/recipes/java/spring/boot2/unnecessaryspringrunwith) or as a top-level recipe addition to the migration recipe.

nbruno:
This recipe will convert JUnit 5 test classes that use both `@ExtendWith` and `@ContextConfiguration` into [`@SpringJUnitConfig`](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/test/context/junit/jupiter/SpringJUnitConfig.html).

It changes the `@ContextConfiguration` annotation into `@SpringJUnitConfig`, with a few modifications:

- The `loader()` attribute on `@ContextConfiguration` is not supported on `@SpringJUnitConfig`, so if this is encountered, the transformation is abandoned out of caution and to adhere to the "do no harm" principle.
- The `value()` attribute on `@ContextConfiguration` needs to be renamed to `locations()`. It's invalid for an existing `@ContextConfiguration` annotation to be using both `value()` and `locations()`, but if this were encountered, the recipe would overwrite any pre-existing value for `locations()`. 

The recipe reuses the existing recipe [`UnnecessarySpringExtension`](https://docs.openrewrite.org/reference/recipes/java/spring/boot2/unnecessaryspringextension) to remove the `@ExtendsWith(SpringExtension.class)` annotation. 

It will also apply to the [JUnit 4 to JUnit 5 migration recipe](https://docs.openrewrite.org/reference/recipes/java/spring/boot2/springboot2junit4to5migration) as it is added to the [`UnnecessarySpringRunWith`](https://docs.openrewrite.org/reference/recipes/java/spring/boot2/unnecessaryspringrunwith) recipe.

Closes #296

### DIFF
--- a/src/main/resources/META-INF/rewrite/spring-boot-24.yml
+++ b/src/main/resources/META-INF/rewrite/spring-boot-24.yml
@@ -141,6 +141,8 @@ recipeList:
   - org.openrewrite.java.spring.boot2.OutputCaptureExtension
   - org.openrewrite.java.spring.boot2.UnnecessarySpringRunWith
   - org.openrewrite.java.spring.boot2.UnnecessarySpringExtension
+# TODO Enable once tested on more projects through public.moderne.io
+#  - org.openrewrite.java.spring.boot2.ReplaceExtendWithAndContextConfiguration
   - org.openrewrite.java.spring.boot2.RemoveObsoleteSpringRunners
   - org.openrewrite.maven.AddDependency:
       groupId: org.springframework.boot
@@ -158,7 +160,6 @@ recipeList:
         - org.springframework.test.context.junit4.SpringRunner
         - org.springframework.test.context.junit4.SpringJUnit4ClassRunner
       extension: org.springframework.test.context.junit.jupiter.SpringExtension
-  - org.openrewrite.java.spring.boot2.ReplaceExtendWithAndContextConfiguration
 ---
 type: specs.openrewrite.org/v1beta/recipe
 name: org.openrewrite.java.spring.boot2.RemoveObsoleteSpringRunners
--- a/src/main/resources/META-INF/rewrite/spring-boot-24.yml
+++ b/src/main/resources/META-INF/rewrite/spring-boot-24.yml
@@ -158,6 +158,7 @@ recipeList:
         - org.springframework.test.context.junit4.SpringRunner
         - org.springframework.test.context.junit4.SpringJUnit4ClassRunner
       extension: org.springframework.test.context.junit.jupiter.SpringExtension
+  - org.openrewrite.java.spring.boot2.ReplaceExtendWithAndContextConfiguration
 ---
 type: specs.openrewrite.org/v1beta/recipe
 name: org.openrewrite.java.spring.boot2.RemoveObsoleteSpringRunners
--- a/src/main/java/org/openrewrite/java/spring/boot2/ReplaceExtendWithAndContextConfiguration.java
+++ b/src/main/java/org/openrewrite/java/spring/boot2/ReplaceExtendWithAndContextConfiguration.java
@@ -15,6 +15,7 @@
  */
 package org.openrewrite.java.spring.boot2;
 
+import org.openrewrite.Applicability;
 import org.openrewrite.ExecutionContext;
 import org.openrewrite.Recipe;
 import org.openrewrite.TreeVisitor;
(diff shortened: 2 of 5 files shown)


src/main/resources/META-INF/rewrite/spring-boot-24.yml
```diff
@@ -158,6 +158,7 @@ recipeList:
         - org.springframework.test.context.junit4.SpringRunner
         - org.springframework.test.context.junit4.SpringJUnit4ClassRunner
       extension: org.springframework.test.context.junit.jupiter.SpringExtension
+  - org.openrewrite.java.spring.boot2.ReplaceExtendWithAndContextConfiguration
```
timtebeek, line 161: In the past we've occasionally held off on making new recipes part of the default chain of upgrades until after they had been tried on a large-ish set of projects, for instance through public.moderne.io. Just because when applied at scale, there might be uncovered edge cases that could trip up other users.

I'm impressed by the attention to detail and test cases already covered here; have you also already tried this on a few projects there? That might give us the confidence we need to indeed make this part of the default chain of recipes immediately.

timtebeek (approved):
Very impressed by the level of detail in both your tests, and in covering non-obvious cases in the recipe itself. That makes it a joy to review; thank you very much!

I've merely added  a few touch up commits to make it fit in with other recipes; overriding `getSingleSourceApplicableTest` for instance, and applying the default formatter & import order.

Also posed one question on making this immediately part of the upgrade chain, before we're good to merge this one (in my opinion). We could also run this soon after a merge through public.moderne.io on the Spring organization; that could uncover some more cases, or lead to contributions there.

src/main/resources/META-INF/rewrite/spring-boot-24.yml
nbruno: I wasn't aware of public.moderne.io and didn't try it on any projects there. I did test it locally on one of our projects. I wouldn't mind holding off on including it in a chain of upgrades. I think the important piece is that it eventually becomes part of the JUnit 4 to JUnit 5 migration - However that happens is fine with me.

timtebeek:
1. Good that you thought of requiring both annotations to be present already. Specifically for those cases we have a construct to express that clearly, as applied in https://github.com/openrewrite/rewrite-spring/pull/297/commits/610614146bc7e9e4e11febf824f26bad5c539bc8. Do you agree with that change?
2. Thanks for the suggestion! I've [logged that on our documentation repository](https://github.com/openrewrite/rewrite-docs/issues/146).
3. Sounds good indeed to keep each individual recipe as simple as possible, that helps maintainability in the long run. Would you agree to remove it for now then, and add it to `SpringBoot2JUnit4to5Migration` after we've tried it out on a larger set of Spring repositories?

nbruno:
All of these changes look great to me. Let's remove it from the recipe lists for now and then whenever it's deemed mature enough to be included, I'd recommend adding it to `SpringBoot2JUnit4to5Migration`.

Thanks for your help with this.

PR #301, opened 2023-03-03
Add @ to the annotation pattern string

BoykoAlex:


timtebeek:
I'd been looking for which recipe was logging that error. Glad that you've found it!

timtebeek (approved):

PR #303, opened 2023-03-07
Saml2 property changes only for 2.7

BoykoAlex:
Remove Saml2 property migration from boot 3.0 migration as it is meant for Boot 2.7 migration only


sambsnyd:
Thanks @BoykoAlex ! Appreciate you contributing this sort of detailed domain knowledge

PR #307, opened 2023-03-15
Add recipe `MaintainTrailingSlashURLMappings`

kunli2:
Introduce recipe `MaintainTrailingSlashURLMappings` to support [Maintain trailing slash in request mappings](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-3.0-Migration-Guide#spring-mvc-and-webflux-url-matching-changes)

Logic: 
If `WebMvcConfigurer` or `WebFluxConfigurer` is not overridden, add new mapping paths for all `@<HTTP Verb>Mapping` annotations.


src/main/java/org/openrewrite/java/spring/boot3/AddMappingPathsWithTrailingSlash.java
```diff
@@ -0,0 +1,145 @@
...
+    @Override
+    public String getDisplayName() {
+        return "Declare the additional route explicitly on the controller handler";
+    }
+
+    @Override
+    public String getDescription() {
+        return "This is part of Spring MVC and WebFlux URL Matching Changes, as of Spring Framework 6.0, the trailing" +
+               " slash matching configuration option has been deprecated and its default value set to false. This " +
+               "means that previously, a controller \"@GetMapping(\"/some/greeting\")\" would match both \"GET " +
```
sambsnyd, line 37:
```suggestion
               "means that previously, a controller `@GetMapping(\"/some/greeting\")` would match both \"GET " +
```

src/main/java/org/openrewrite/java/spring/boot3/AddMappingPathsWithTrailingSlash.java
```diff
@@ -0,0 +1,145 @@
...
+    @Override
+    protected TreeVisitor<?, ExecutionContext> getVisitor() {
+        return new JavaIsoVisitor<ExecutionContext>(){
+
+            @Override
+            public J.Annotation visitAnnotation(J.Annotation annotation, ExecutionContext ctx) {
+                J.Annotation anno = super.visitAnnotation(annotation, ctx);
+
+                if (anno.getType() != null &&
+                    isHttpVerbMappingAnnotation(anno.getType().toString())) {
```
sambsnyd, line 52: Stylistically I tend to prefer early return so that code does not get deeply indented.
```java
if (anno.getType() == null) {
    return anno;
}
```

There is nothing wrong with writing it as you have here. I find this early return style to keep code more legible even when there are many conditions being guarded against

src/main/java/org/openrewrite/java/spring/boot3/AddMappingPathsWithTrailingSlash.java
```diff
@@ -0,0 +1,145 @@
...
+               PUT_ANNOTATION_TYPE.equals(fqn) ||
+               PATCH_ANNOTATION_TYPE.equals(fqn) ||
+               DELETE_ANNOTATION_TYPE.equals(fqn);
+    }
+
+    private static J.NewArray getTwoStringsArrayTemplate() {
+        if (twoStringsArrayTemplate == null) {
+            twoStringsArrayTemplate = PartProvider.buildPart("class Test {\n" +
+                                                                "    String[] value = { \"a\", \"b\"};\n" +
+                                                                "}",
```
sambsnyd, line 122: For legibility I would put the string literal on the next line so the code snippet isn't indented so far to the right

src/testWithSpringBoot_3_0/java/org/openrewrite/java/spring/boot3/AddMappingPathsWithTrailingSlashTest.java
```diff
@@ -0,0 +1,199 @@
...
+                  }
+              }
+              """,
+            """
+              import org.springframework.web.bind.annotation.*;
+
+              @RestController
+              public class ExampleController {
+
+                  @GetMapping(value = {"/get", "/get/"})
```
sambsnyd, line 40: When there is only one argument to an annotation _and_ the annotation type has a field named `value`, the argument is understood by the java compiler to be an argument to `value`. 

Whenever there is more than one argument, or there is no field on the annotation type named `value`, specifying the name of the annotation field becomes mandatory.

In this case since the "before" argument did not need `value =` I would not expect that `value =` would be introduced

src/main/java/org/openrewrite/java/spring/boot3/AddMappingPathsWithTrailingSlash.java
```diff
@@ -0,0 +1,145 @@
...
+import org.openrewrite.java.PartProvider;
+import org.openrewrite.java.tree.*;
+
+import java.util.Collections;
+import java.util.List;
+
+import static java.util.Collections.emptyList;
+import static org.openrewrite.java.tree.Space.EMPTY;
+
+public class AddMappingPathsWithTrailingSlash extends Recipe {
```
sambsnyd, line 18:
```suggestion
public class AddRouteTrailingSlash extends Recipe {
```

src/main/java/org/openrewrite/java/spring/boot3/AddMappingPathsWithTrailingSlash.java
```diff
@@ -0,0 +1,145 @@
...
+    private static final String POST_ANNOTATION_TYPE = "org.springframework.web.bind.annotation.PostMapping";
+    private static final String PUT_ANNOTATION_TYPE = "org.springframework.web.bind.annotation.PutMapping";
+    private static final String PATCH_ANNOTATION_TYPE = "org.springframework.web.bind.annotation.PatchMapping";
+    private static final String DELETE_ANNOTATION_TYPE = "org.springframework.web.bind.annotation.DeleteMapping";
+    @Nullable private static J.NewArray twoStringsArrayTemplate;
+    @Nullable private static J.Assignment valueAssignmentTemplate;
+
+    @Override
+    public String getDisplayName() {
+        return "Declare the additional route explicitly on the controller handler";
```
sambsnyd, line 30: This display name does not do a great job of describing what this recipe does. Maybe something like this?

```suggestion
        return "Add trailing slash to Spring routes";
```

src/main/java/org/openrewrite/java/spring/boot3/MaintainTrailingSlashURLMappings.java
```diff
@@ -0,0 +1,72 @@
...
+
+        doNext(new AddMappingPathsWithTrailingSlash());
+        return before;
+    }
+
+    private static class FindWebMvcConfigurer extends JavaIsoVisitor<AtomicBoolean> {
+        static AtomicBoolean find(J j) {
+            return new FindWebMvcConfigurer()
+                .reduce(j, new AtomicBoolean());
+        }
```
sambsnyd, line 55: I would probably return `boolean`. Feels like an implementation detail that this is implemented with `AtomicBoolean` that callers do not need to know about

sambsnyd (commented):
Good work

src/testWithSpringBoot_3_0/java/org/openrewrite/java/spring/boot3/AddMappingPathsWithTrailingSlashTest.java
kunli2: Good one, fixed.

knutwannheden:
@kunli2 The recipe was not yet added to any of the YAML files. Will you do that in a second step? I am asking because I have also started working on the Spring recipes now and am still a bit unsure about the development practices in this repo.

Issue #306, opened 2023-03-14
Migration to Spring Boot 2.3 should maybe add spring-boot-starter-validation


jjank:
I've tried to migrate Spring-Petclinic from before Boot 2.3 to the lastest Boot 2.7.x.

`ValidatorTests` fails because there is no Bean Validation provider on the classpath:
```
[INFO] Running org.springframework.samples.petclinic.model.ValidatorTests
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.079 s <<< FAILURE! - in org.springframework.samples.petclinic.model.ValidatorTests
[ERROR] shouldNotValidateWhenFirstNameEmpty  Time elapsed: 0.075 s  <<< ERROR!
javax.validation.NoProviderFoundException: Unable to create a Configuration, because no Bean Validation provider could be found. Add a provider like Hibernate Validator (RI) to your classpath.
	at org.springframework.samples.petclinic.model.ValidatorTests.createValidator(ValidatorTests.java:24)
	at org.springframework.samples.petclinic.model.ValidatorTests.shouldNotValidateWhenFirstNameEmpty(ValidatorTests.java:36)
```

The [release-notes of Boot 2.3](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.3-Release-Notes#validation-starter-no-longer-included-in-web-starters) explain this change:

> As of [#19550](https://github.com/spring-projects/spring-boot/issues/19550), Web and WebFlux starters do not depend on the validation starter by default anymore. If your application is using validation features, for example you find that javax.validation.* imports are not being resolved, you’ll need to add the starter yourself.

In the Spring Boot 2.0 migration rewrite [adds the javax-validation-api](https://github.com/openrewrite/rewrite-spring/blob/main/src/main/resources/META-INF/rewrite/spring-boot-20.yml#L130) as a maven dependency. But that does not solve the problem of a missing BV implementation starting from 2.3

What is probably needed is something in [/spring-boot-23.yml](https://github.com/openrewrite/rewrite-spring/blob/main/src/main/resources/META-INF/rewrite/spring-boot-23.yml) along the lines of:

```yaml
  - org.openrewrite.maven.AddDependency:
      groupId: org.springframework.boot
      artifactId: spring-boot-starter-validation
      version: 2.3.x
      onlyIfUsing: javax.validation.constraints.*
```

[...]

PR #308, opened 2023-03-16
#306 Add spring-boot-starter-validation in Boot 2.3

jjank:
Since Boot 2.3 `spring-boot-starter-validation` is no longer a transitive dependency of the web starters.

While the code will probably compile fine due to the presence of `validation-api` which is added as part of the Boot 2.0 migration runtime errors may occur because no Bean Validation implementation is present on the classpath anymore.

See https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.3-Release-Notes#validation-starter-no-longer-included-in-web-starters


timtebeek (approved):
Thanks again @jjank !

jjank:
@timtebeek any particular reason why this is not merged yet? 🙈

src/main/resources/META-INF/rewrite/spring-boot-23.yml
```diff
@@ -39,6 +39,12 @@ recipeList:
       groupId: org.springframework.boot
       artifactId: spring-boot-starter-parent
       newVersion: 2.3.x
+  # https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.3-Release-Notes#validation-starter-no-longer-included-in-web-starters
+  - org.openrewrite.maven.AddDependency:
+      groupId: org.springframework.boot
+      artifactId: spring-boot-starter-validation
+      version: 2.3.x
+      onlyIfUsing: javax.validation.constraints.*
```
timtebeek, line 48: Might be best to [add this new flag](https://github.com/openrewrite/rewrite/blob/9ecb9a1f75b9f4a8289e97f9fd6ed5e2a930f008/rewrite-maven/src/main/java/org/openrewrite/maven/AddDependency.java#L128-L133) to prevent this from being added if it's also available transitively.
```suggestion
      acceptTransitive: true

```

timtebeek:
> @timtebeek any particular reason why this is not merged yet? see_no_evil

Sorry about that! Went to check something, and asked colleagues to have a look in case I missed something. I've added a new suggestion just to prevent from any side effects where we would not want this dependency added.

The thing I'm still wondering about is whether we've covered enough cases to "[do no harm](https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices#do-no-harm)". My thinking is we should _really_ only add this dependency in a migration coming from anything below 2.3, not when users are already on 2.3+ and applying this recipe as part of a chain to for instance go to 2.7 or 3.0.

My reasoning here is that althought ideally it might be better to also add this dependency in case it's missing, also on projects already on 2.3+, such as when users manually migrated before and forgot to add the dependency.
In practice adding the dependency might end up doing harm: imagine a company that migrates 100+ services without manual inspection: it's likely a case snuck in where now adding the dependency might break callers at runtime.
I've seen that in practice, and it's not a fun to debug when validation was missing for some time. Perhaps we should add a marker for such cases? Thinking out loud here.

jjank:
>  I've added a new suggestion just to prevent from any side effects where we would not want this dependency added.

Good idea -> committed 👍 

I share your do-no-harm concerns. Probably we need to do more work here.

Another idea: We could check if a BV-implementation (I suppose it Hibernate-Validator is the only "serious" implementation we'd have to check...) is already transitively or directly on the classpath? That's the core issue I'm trying to address here actually. At least that is what caused pet-clinic to fail some tests.

knutwannheden:
I think it would be nice if we could check for specific dependency artifact versions on the classpath. But I don't think we already offer an applicability test for that. It would have to work with at least Maven and Gradle and then the question is also what it should do in a multi-module setup. But I guess this last question applies regardless of what applicability test we use.

knutwannheden (approved):

knutwannheden:
LGTM. Thanks for the contribution, @jjank !

Issue #256, opened 2022-12-26
Migration Recipe for Spring Security 6.0


tkvangorder:
The Spring Security team has produced a [migration guide for Spring Security 6.0](https://docs.spring.io/spring-security/reference/6.0.0/migration/index.html) that includes the following:

- [x] [remove `requireExplicitSave(true)`](https://docs.spring.io/spring-security/reference/6.0.0/migration/servlet/session-management.html#_require_explicit_saving_of_securitycontextrepository) 
- [x] [Optimize Querying of RequestCache](https://docs.spring.io/spring-security/reference/6.0.0/migration/servlet/session-management.html#requestcache-query-optimization) ([recipe](https://github.com/openrewrite/rewrite-spring/pull/327))
- [x] [Propagate AuthenticationServiceExceptions](https://docs.spring.io/spring-security/reference/6.0.0/migration/servlet/authentication.html#_propagate_authenticationserviceexceptions)
- [x] [Use SHA-256 in Remember Me](https://docs.spring.io/spring-security/reference/6.0.0/migration/servlet/authentication.html#servlet-opt-in-sha256-rememberme)
- [x] [Default authorities for oauth2Login()](https://docs.spring.io/spring-security/reference/6.0.0/migration/servlet/authentication.html#_default_authorities_for_oauth2login) ([recipe](https://github.com/openrewrite/rewrite-spring/pull/328))
- [x] [Use AuthorizationManager for Message Security](https://docs.spring.io/spring-security/reference/6.0.0/migration/servlet/authorization.html#_use_authorizationmanager_for_message_security) ([recipe](https://github.com/openrewrite/rewrite-spring/pull/330))
  - This is an XML only migration step (presumably for `beans.xml`)
- [x] [Use AuthorizationManager for Request Security](https://docs.spring.io/spring-security/reference/6.0.0/migration/servlet/authorization.html#_use_authorizationmanager_for_request_security) ([recipe](https://github.com/openrewrite/rewrite-spring/blob/main/src/main/java/org/openrewrite/java/spring/security6/RemoveFilterSecurityInterceptorOncePerRequest.java))
- [x] [Use the new requestMatchers methods](https://docs.spring.io/spring-security/reference/5.8/migration/servlet/config.html#use-new-requestmatchers)
- [x] [Add @Configuration to @Enable* annotations](https://docs.spring.io/spring-security/reference/5.8/migration/servlet/config.html#_add_configuration_to_enable_annotations)

[...]

PR #311, opened 2023-03-20
Remove calls matching `AuthenticationEntryPointFailureHandler.setRethrowAuthenticationServiceException(true)`

knutwannheden:
See https://docs.spring.io/spring-security/reference/6.0.0/migration/servlet/authentication.html#_propagate_authenticationserviceexceptions for details.

Related: #256


knutwannheden:
This recipe still needs to be registered in one of the YAML files. I think we should maybe create a new `spring-security-60.yml` file, as AFAICT this recipe should both be applicable for Spring Framework and Spring Boot applications. Then the recipe should be added to the corresponding `spring-boot-30.yml` and `spring-framework-60.yml` files. Does that make sense?

PR #313, opened 2023-03-21
Change `ConfigurationOverEnableSecurity` to modify only before Boot3

kunli2:
Change `ConfigurationOverEnableSecurity` to modify only before Boot3. 
and no change if it's already Boot 3.

Slack discussion [here](https://moderneinc.slack.com/archives/C017J8U6ZGC/p1679331496438579).


sambsnyd (approved):
Let's add a "force" parameter which will add the `@Configuration` annotation even if the check for "is already spring boot 3" comes back true, just in case someone wants to use this recipe on code they already moved to spring boot 3

kunli2:
Sure, I will make a separate PR to add a "force" parameter

PR #314, opened 2023-03-21
New `RequireExplicitSavingOfSecurityContextRepository` recipe

knutwannheden:
Adds a new recipe for the https://docs.spring.io/spring-security/reference/6.0.0/migration/servlet/session-management.html#_require_explicit_saving_of_securitycontextrepository migration step.

Related: #256


src/test/java/org/openrewrite/java/spring/security6/RequireExplicitSavingOfSecurityContextRepositoryTest.java
```diff
@@ -0,0 +1,258 @@
...
+            """
+              import org.springframework.security.config.annotation.web.builders.HttpSecurity;
+              import org.springframework.security.web.SecurityFilterChain;
+
+              class T {
+                  public SecurityFilterChain chain(HttpSecurity http) {
+                      http
+                          .securityContext((securityContext) -> securityContext
+                              .requireExplicitSave(true)
+                              .requireExplicitSave(false)
```
knutwannheden, line 50: Please note that I don't check for conflicting calls to `requireExplicitSave()`. Rather I just want to make sure that the recipe works correctly regardless of where in the chain the invocation to be removed exists (first, middle, last).

PR #326, opened 2023-04-02
Replace http literals with constants

### LINKED ISSUE #325: Replace String literals with constants for HttpHeaders and MediaTypes
Spring defines various [constants for HttpHeaders](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/http/HttpHeaders.html) and [MediaTypes](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/http/MediaType.html).
These `_VALUE`  suffixed [constants can replace String literals](https://github.com/openrewrite/rewrite/blob/main/rewrite-java/src/main/java/org/openrewrite/java/ReplaceStringLiteralWithConstant.java#L40).

This recipe would likely consist of an extended version of this first start:
```yaml
name: ReplaceStringLiteralsWithMediaTypeConstants
displayName: Replace String literals with MediaType constants
description: Replace String literals anywhere in the code with fully qualified MediaType constants.
recipeList:
  - org.openrewrite.java.ReplaceStringLiteralWithConstant:
      fullyQualifiedConstantName: org.springframework.http.MediaType.APPLICATION_JSON_VALUE
  - org.openrewrite.java.ReplaceStringLiteralWithConstant:
      fullyQualifiedConstantName: org.springframework.http.MediaType.TEXT_PLAIN_VALUE
```

Optional to tag onto this recipe another recipe that [converts the FQN to static imports](https://docs.openrewrite.org/reference/recipes/java/usestaticimport) where possible, for more readable code.

As the [Spring RequestMapping annotations](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/bind/annotation/RequestMapping.html) take String literals as arguments, these recipes could be a nice way to potentially clean that up a bit to use framework provided constants.

timtebeek:
Fixes #325

Special thanks to @srmalkan and @AlexanderSkrock for the [ReplaceStringLiteralWithConstant recipe](https://github.com/openrewrite/rewrite/blob/main/rewrite-java/src/main/java/org/openrewrite/java/ReplaceStringLiteralWithConstant.java) and usability improvements.

Should be easy enough to add new recipes if we can think of more; these two in particular seemed interesting because of their use in request mapping annotations.


timtebeek:
Wasn't sure if we should hook this into best practices already; maybe try it out some more first.

src/main/resources/META-INF/rewrite/replace-MediaType-literals.yml
```diff
@@ -0,0 +1,86 @@
...
+---
+type: specs.openrewrite.org/v1beta/recipe
+name: org.openrewrite.java.spring.http.ReplaceStringLiteralsWithMediaTypeConstants
+displayName: Replace String literals with `MediaType` constants
+description: Replace String literals with `org.springframework.http.MediaType` constants.
+applicability:
+  anySource:
+    - org.openrewrite.maven.search.FindDependency:
+        groupId: org.springframework
+        artifactId: spring-web
```
timtebeek, line 25: This fails to pick up transitive dependencies; I see two quick options, but both require changes in rewrite:

1. Add an option to FindDependency to allow picking up transitive dependencies
2. Switch to DependencyInsight, and add an option there to generate markers

Any other ideas or preferences?

src/testWithSpringBoot_2_7/java/org/openrewrite/java/spring/http/GenerateReplaceStringLiteralWithConstantRecipeTest.java
```diff
@@ -0,0 +1,129 @@
...
+            .map(field -> """
+                - org.openrewrite.java.ReplaceStringLiteralWithConstant:
+                    fullyQualifiedConstantName: %s.%s
+              """.formatted(field.getDeclaringClass().getName(), field.getName()))
+            .collect(joining("")).getBytes(),
+          StandardOpenOption.APPEND);
+    }
+
+    @Test
+    @Disabled("This test is disabled because it is not a test. It is a utility for generating recipes.")
```
sambsnyd, line 97: Why add the `@Test` annotation if this is not a test?

sambsnyd:
I will be happy to merge this once the build passes, and ideally there aren't non-test utilities misleadingly annotated as `@Test`s when they are not tests.

timtebeek:
Build not fails with 
```
Caused by: org.gradle.internal.resolve.ModuleVersionNotFoundException: Could not find org.gradle:gradle-tooling-api:8.1-rc-3.
Searched in the following locations:
  - file:/home/runner/.m2/repository/org/gradle/gradle-tooling-api/8.1-rc-3/gradle-tooling-api-8.1-rc-3.pom
  - https://oss.sonatype.org/content/repositories/snapshots/org/gradle/gradle-tooling-api/8.1-rc-3/gradle-tooling-api-8.1-rc-3.pom
  - https://repo.maven.apache.org/maven2/org/gradle/gradle-tooling-api/8.1-rc-3/gradle-tooling-api-8.1-rc-3.pom
  - https://repo.spring.io/milestone/org/gradle/gradle-tooling-api/8.1-rc-3/gradle-tooling-api-8.1-rc-3.pom
Required by:
    project : > org.openrewrite:rewrite-gradle:7.40.0-SNAPSHOT:20230408.160754-29
```
I'll try again after the weekend.

timtebeek:
After https://github.com/openrewrite/rewrite-build-gradle-plugin/commit/fe833b06b4d2b9b7dd245937cb2a0767c7e505ba we're now getting
```
Caused by: org.gradle.internal.resolve.ModuleVersionNotFoundException: Could not find org.gradle:gradle-tooling-api:8.0.2.
Searched in the following locations:
  - file:/home/runner/.m2/repository/org/gradle/gradle-tooling-api/8.0.2/gradle-tooling-api-8.0.2.pom
  - https://oss.sonatype.org/content/repositories/snapshots/org/gradle/gradle-tooling-api/8.0.2/gradle-tooling-api-8.0.2.pom
  - https://repo.maven.apache.org/maven2/org/gradle/gradle-tooling-api/8.0.2/gradle-tooling-api-8.0.2.pom
  - https://repo.spring.io/milestone/org/gradle/gradle-tooling-api/8.0.2/gradle-tooling-api-8.0.2.pom
```
Curious 🤔

timtebeek:
Build issues have been resolved, and non test utilities cleaned up.

Thinking about this one again I wonder if we should instead create a higher-level recipe that merely takes a FQN option, and for that calls `Class.forName(fullyQualifiedClassName).getFields()`, and then for each just calls `doNext(new ReplaceStringLiteralWithConstant(fullyQualifiedClassName + '.' + field.getName()))`. That way we do not have to generate & maintain the yaml recipe list for each field, but rather only pass the FQN once. Or would that be too much magic?

Otherwise fine with merging this as is; the [GenerateReplaceStringLiteralWithConstantRecipeTest](https://github.com/openrewrite/rewrite-spring/pull/326/files#diff-77ca12c1619bd4eba923319446d2b45e16aee17159492b3fa3477d8fe640389aR45-R94) utility already takes much of the maintenance pain out, although the recipe would not be updated automatically yes if new fields are added.

sambsnyd:
> Thinking about this one again I wonder if we should instead create a higher-level recipe that merely takes a FQN option, and for that calls `Class.forName(fullyQualifiedClassName).getFields()`, and then for each just calls `doNext(new ReplaceStringLiteralWithConstant(fullyQualifiedClassName + '.' + field.getName()))`. That way we do not have to generate & maintain the yaml recipe list for each field, but rather only pass the FQN once. Or would that be too much magic?
I think that sounds like a useful recipe. Could call it "UseConstants"

PR #327, opened 2023-04-05
[Spring Security6 migration] Add recipe UpdateRequestCacheTest

kunli2:
This recipe is part of  [Spring Security 6.0 migration ](https://github.com/openrewrite/rewrite-spring/issues/256) to [Optimize Querying of RequestCache](https://docs.spring.io/spring-security/reference/6.0.0/migration/servlet/session-management.html#requestcache-query-optimization)

By default, Spring Security 5 queries the saved request on every request, which means that in a typical setup, the HttpSession is queried on every request to use the RequestCache. In Spring Security 6, the default behavior has changed, and RequestCache will only be queried for a cached request if the HTTP parameter `continue` is defined. This helps Spring Security avoid unnecessary HttpSession queries with the RequestCache. To maintain the same default behavior as Spring Security 5, either explicitly add the HTTP parameter `continue` to every request or use NullRequestCache to override the default behavior.

### DIFF
--- a/src/main/java/org/openrewrite/java/spring/security6/UpdateRequestCache.java
+++ b/src/main/java/org/openrewrite/java/spring/security6/UpdateRequestCache.java
@@ -16,14 +16,18 @@
 package org.openrewrite.java.spring.security6;
 
 import org.openrewrite.*;
+import org.openrewrite.internal.lang.Nullable;
 import org.openrewrite.java.JavaIsoVisitor;
 import org.openrewrite.java.JavaParser;
 import org.openrewrite.java.JavaTemplate;
 import org.openrewrite.java.MethodMatcher;
+import org.openrewrite.java.search.UsesMethod;
 import org.openrewrite.java.tree.Expression;
 import org.openrewrite.java.tree.J;
 import org.openrewrite.java.tree.Statement;
 import org.openrewrite.java.tree.TypeUtils;
+
+import java.time.Duration;
 import java.util.Collections;
 import java.util.List;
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/openrewrite/java/spring/security6/UpdateRequestCache.java
```diff
@@ -0,0 +1,177 @@
...
+                        JavaTemplate template = JavaTemplate.builder(this::getCursor,
+                                "#{any()}.setMatchingRequestParameterName(\"continue\");")
+                            .javaParser(JavaParser.fromJavaVersion()
+                                .classpath("spring-security-web"))
+                            .imports(
+                                "org.springframework.security.web.savedrequest.HttpSessionRequestCache")
+                            .build();
+
+                        statement = statement.withTemplate(template, statement.getCoordinates().replace(), getSelect(statement));
+                        statements.add(i + 1, statement);
```
sambsnyd, line 95: Never directly mutate a collection of AST elements. Ideally the collections themselves would be immutable. 
Our change detection and diff calculation can be messed up by this sort of mutation. To add an element use `ListUtils.concat()`. 

Even if it does not cause a problem in this specific case, we don't want examples of this mistake floating around for people to copy from

kunli2: Yeah, However, I see there is currently no method in `ListUtils` to support inserting an element into a specific index. I think we need to add two methods to `ListUtils`
1. add an element to a specific index
2. remove an element from a specific index

kunli2: I see an example that returning `null` for a `visitMethodDeclaration` in the recipe `RemoveUnusedPrivateMethods` can remove that method, so for the statement removing, I think I can do the similar without a method from ListUtils

kunli2: Added a insert method to `ListUtils` and changed to call that method instead here

PR #328, opened 2023-04-06
[Spring Security6 migration]Add recipe `RemoveOauth2LoginConfig`

kunli2:
This recipe is part of the [Spring Security6 migration](https://github.com/openrewrite/rewrite-spring/issues/256) to implement to keep [Default authorities for oauth2Login()](https://docs.spring.io/spring-security/reference/6.0.0/migration/servlet/authentication.html#_default_authorities_for_oauth2login).

### DIFF
--- a/src/main/java/org/openrewrite/java/spring/security6/RemoveOauth2LoginConfig.java
+++ b/src/main/java/org/openrewrite/java/spring/security6/RemoveOauth2LoginConfig.java
@@ -110,7 +110,7 @@ public class RemoveOauth2LoginConfig extends Recipe {
                         return method.getSelect().withPrefix(method.getPrefix());
                     }
                 }
-                return super.visitMethodInvocation(method, executionContext);
+                return method;
             }
         };
     }
--- a/src/main/java/org/openrewrite/java/spring/security6/RemoveOauth2LoginConfig.java
+++ b/src/main/java/org/openrewrite/java/spring/security6/RemoveOauth2LoginConfig.java
@@ -1,0 +1,77 @@
+package org.openrewrite.java.spring.security6;
+
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.Recipe;
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.internal.lang.Nullable;
+import org.openrewrite.java.JavaIsoVisitor;
+import org.openrewrite.java.MethodMatcher;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.java.tree.Statement;
+
+import java.time.Duration;
+import java.util.List;
+
+public class RemoveOauth2LoginConfig extends Recipe {
+    private static final MethodMatcher USER_AUTHORITIES_MAPPER_MATCHER = new MethodMatcher(
+        "org.springframework.security.config.annotation.web.configurers.oauth2.client.OAuth2LoginConfigurer userAuthoritiesMapper(org.springframework.security.core.authority.mapping.GrantedAuthoritiesMapper)"
+    );
+
+
+    @Override
+    public String getDisplayName() {
+        return "Remove unneeded `oauth2Login` config when upgrading to Spring Security 6";
+    }
+
+    @Override
+    public String getDescription() {
+        return "oauth2Login() is a Spring Security feature that allows users to authenticate with an OAuth2 or OpenID" +
+               " Connect 1.0 provider. When a user is authenticated using this feature, they are granted a set of " +
+               "authorities that determines what actions they are allowed to perform within the application.\n" +
+               "\n" +
+               "In Spring Security 5, the default authority given to a user authenticated with an OAuth2 or OpenID " +
+               "Connect 1.0 provider via oauth2Login() is ROLE_USER. This means that the user is allowed to access " +
+               "the application's resources as a regular user.\n" +
+               "\n" +
+               "However, in Spring Security 6, the default authority given to a user authenticated with an OAuth2 " +
+               "provider is OAUTH2_USER, and the default authority given to a user authenticated with an OpenID " +
+               "Connect 1.0 provider is OIDC_USER. These authorities are more specific and allow for better " +
+               "customization of the user's permissions within the application.\n" +
+               "\n" +
+               "If you are upgrading to Spring Security 6 and you have previously configured a " +
+               "GrantedAuthoritiesMapper to handle the authorities of users authenticated via oauth2Login(), you can " +
+               "remove it completely as the new default authorities should be sufficient.";
+    }
+
+    @Override
+    public @Nullable Duration getEstimatedEffortPerOccurrence() {
+        return Duration.ofMinutes(8);
+    }
+
+    @Override
+    protected TreeVisitor<?, ExecutionContext> getVisitor() {
+        return new JavaIsoVisitor<ExecutionContext>() {
+
+            @Override
+            public J.Block visitBlock(J.Block block, ExecutionContext executionContext) {
+                block = super.visitBlock(block, executionContext);
+
+
+                List<Statement> statements = block.getStatements();
+
+
+                return block;
+            }
+
+            @Override
+            public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method,
+                                                            ExecutionContext executionContext) {
+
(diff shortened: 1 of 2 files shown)


src/main/java/org/openrewrite/java/spring/security6/RemoveOauth2LoginConfig.java
```diff
@@ -0,0 +1,116 @@
...
+               "\n" +
+               "However, in Spring Security 6, the default authority given to a user authenticated with an OAuth2 " +
+               "provider is OAUTH2_USER, and the default authority given to a user authenticated with an OpenID " +
+               "Connect 1.0 provider is OIDC_USER. These authorities are more specific and allow for better " +
+               "customization of the user's permissions within the application.\n" +
+               "\n" +
+               "If you are upgrading to Spring Security 6 and you have previously configured a " +
+               "GrantedAuthoritiesMapper to handle the authorities of users authenticated via oauth2Login(), you can " +
+               "remove it completely as the new default authorities should be sufficient.";
+    }
```
sambsnyd, line 69:
```suggestion
    public String getDescription() {
        //language=markdown
        return "oauth2Login() is a Spring Security feature that allows users to authenticate with an OAuth2 or OpenID" +
               " Connect 1.0 provider. When a user is authenticated using this feature, they are granted a set of " +
               "authorities that determines what actions they are allowed to perform within the application.\n" +
               "\n" +
               "In Spring Security 5, the default authority given to a user authenticated with an OAuth2 or OpenID " +
               "Connect 1.0 provider via `oauth2Login()` is `ROLE_USER`. This means that the user is allowed to access " +
               "the application's resources as a regular user.\n" +
               "\n" +
               "However, in Spring Security 6, the default authority given to a user authenticated with an OAuth2 " +
               "provider is `OAUTH2_USER`, and the default authority given to a user authenticated with an OpenID " +
               "Connect 1.0 provider is `OIDC_USER`. These authorities are more specific and allow for better " +
               "customization of the user's permissions within the application.\n" +
               "\n" +
               "If you are upgrading to Spring Security 6 and you have previously configured a " +
               "`GrantedAuthoritiesMapper` to handle the authorities of users authenticated via `oauth2Login()`, you can " +
               "remove it completely as the new default authorities should be sufficient.";
    }
```

sambsnyd: We can use markdown inside these descriptions, particularly backtics "`" to indicate that a particular bit of text should be displayed like code

src/main/java/org/openrewrite/java/spring/security6/RemoveOauth2LoginConfig.java
```diff
@@ -0,0 +1,116 @@
...
+                        J.MethodInvocation m = (J.MethodInvocation) statement;
+                        if (USER_AUTHORITIES_MAPPER_MATCHER.matches(m) ||
+                            O_AUTH_2_LOGIN_MATCHER.matches(m) ||
+                            USER_INFO_ENDPOINT_MATCHER.matches(m)) {
+
+                            if (m.getSelect() instanceof J.Identifier) {
+                                // remove this statement
+                                // todo, maybe add a method `listUtils.remove(int index)`
+                                statements.remove(i);
+                                return block.withStatements(statements);
```
sambsnyd, line 96: Never directly mutate a collection of AST elements.
`ListUtils.map()` can be used to remove elements, just return `null` when you want to remove something

src/main/java/org/openrewrite/java/spring/security6/RemoveOauth2LoginConfig.java
```diff
@@ -0,0 +1,116 @@
...
+    protected TreeVisitor<?, ExecutionContext> getVisitor() {
+        return new JavaIsoVisitor<ExecutionContext>() {
+
+            @Override
+            public J.Block visitBlock(J.Block block, ExecutionContext executionContext) {
+                List<Statement> statements = block.getStatements();
+
+                for (int i = 0 ; i < statements.size(); i++) {
+                    Statement statement = statements.get(i);
+                    if (statement instanceof J.MethodInvocation) {
```
sambsnyd, line 86: As written this will miss method invocations that aren't directly within a block, such as:
```
if(true) methodYouWillNotVisit();
while(methodYouWillNotVisit()) { ... }
```
It's almost always more correct to use a visitor instead of iterating over a list of statements.

src/main/java/org/openrewrite/java/spring/security6/RemoveOauth2LoginConfig.java
kunli2: I see, good to know this! thanks

knutwannheden:
@kunli2 I haven't looked at your change in detail, nor at the requirements. But I can see that it is also about Spring Security and that you also have to remove some method invocations in their builder-style internal DSL API. I did something very similar already for `RequireExplicitSavingOfSecurityContextRepository`. The code may look a bit funky, but it is also surprisingly hard to do this correctly, given that this method call can appear at different places (beginning, middle, end of chain). Maybe you can take a look at what I did and if you find an easier / better solution, please don't hesitate to fix my recipe as well 😁

src/main/java/org/openrewrite/java/spring/security6/RemoveOauth2LoginConfig.java
kunli2: my original thought of writing this way to check statements in a block is concerning to avoid assignments to a variable.
but you are right, It's almost always more correct to use a visitor instead of iterating over a list of statements, so I changed to use `visitMethodInvocation` with adding assignments check.

kunli2:
@knutwannheden , Thanks for the reference of `RequireExplicitSavingOfSecurityContextRepository`, it looks great to me, I like the idea of using markers to delete items!
however, I think this algorithm assumes that all the chained calls return the same type, but this is not true.
so this recipe doesn't work correctly in the cases where chained calls return different types. if we remove one of the method invocations, it doesn't compile.

here is an example case:
```java

    @Test
    void callChainWithDifferentTypes() {
        rewriteRun(
          java(
            """
              import org.springframework.security.config.Customizer;
              import org.springframework.security.config.annotation.web.builders.HttpSecurity;
              import org.springframework.security.config.annotation.web.configurers.SecurityContextConfigurer;
              import org.springframework.security.config.annotation.web.configurers.oauth2.client.OAuth2LoginConfigurer;


              public class T {
                  SecurityContextConfigurer<HttpSecurity> mySecurityContext;
                  OAuth2LoginConfigurer<HttpSecurity> myOAuth2;

                  public OAuth2LoginConfigurer<HttpSecurity> customize(
                      Customizer<SecurityContextConfigurer<HttpSecurity>> securityContextCustomizer) {
                      securityContextCustomizer.customize(mySecurityContext);
                      return myOAuth2;
                  }

                  public void doSomething(SecurityContextConfigurer<HttpSecurity> myConfigurer) {
                      OAuth2LoginConfigurer<HttpSecurity> auth = this.customize(securityContext -> {
                          securityContext.requireExplicitSave(true);
                      }).permitAll();
                  }
              }
              """,
            """
              import org.springframework.security.config.Customizer;
              import org.springframework.security.config.annotation.web.builders.HttpSecurity;
              import org.springframework.security.config.annotation.web.configurers.SecurityContextConfigurer;
              import org.springframework.security.config.annotation.web.configurers.oauth2.client.OAuth2LoginConfigurer;


              public class T {
                  SecurityContextConfigurer<HttpSecurity> mySecurityContext;
                  OAuth2LoginConfigurer<HttpSecurity> myOAuth2;

                  public OAuth2LoginConfigurer<HttpSecurity> customize(
                      Customizer<SecurityContextConfigurer<HttpSecurity>> securityContextCustomizer) {
                      securityContextCustomizer.customize(mySecurityContext);
                      return myOAuth2;
                  }

                  public void doSomething(SecurityContextConfigurer<HttpSecurity> myConfigurer) {
                      OAuth2LoginConfigurer<HttpSecurity> auth = this.customize(securityContext -> {
                      }).permitAll();
                  }
              }
              """
          )
        );
    }
```

knutwannheden:
@kunli2 Thanks for the example. I will try to see what can be done about this. I hadn't noticed that these builder-like DSLs had different return types on some methods.

PR #332, opened 2023-04-10
Introduce generic visitor `RemoveMethodInvocationsVisitor` to be able to remove some specific method calls.

kunli2:
Introduce the visitor `RemoveMethodInvocationsVisitor` to be able to remove some specific method calls.
This visitor is a generic function to support at least 3 spring security migration recipes to delete different method calls which are unneeded anymore. 
The recipes are
1. `RemoveOauth2LoginConfig`, existing recipe, migrated to use new visitor.
2. `RemoveFilterSecurityInterceptorOncePerRequest` (New recipe introduced in this PR)
3. `RequireExplicitSavingOfSecurityContextRepository`. (just demonstrated in the test by replacing to use this visitor, not changed the recipe itself in this PR)


knutwannheden:
@kunli2 I think this looks useful. I was also contemplating implementing something like this for `RequireExplicitSavingOfSecurityContextRepository` but wanted to wait until I find some more cases, where this would actually be useful.

Looking at `RequireExplicitSavingOfSecurityContextRepositoryTest` it looks like you found yet one more case, where my recipe didn't do all it was supposed to do. Thanks for that. I have fixed that in my recipe for now. Comparing to your utility the only thing I am missing is the ability to also delete the calls with the lambda and the empty body in it. That should of course only be done, where this actually makes sense (a lambda with an empty body could theoretically be required somewhere). Possibly your visitor can be extended with this behavior or another independent visitor can be added which does this part. Since you modified the test cases to match the behavior of your utility, I wanted to check with you what your thoughts are on this.

knutwannheden:
@kunli2 It looks like you just made some changes in this area. I will try to check your changes again.

kunli2:
Hi @knutwannheden, yeah, I copied the same code from `RequireExplicitSavingOfSecurityContextRepository ` for the feature of empty body lambda cleaning up. but I am still concerned that cleaning up empty lambda would be risky for some cases we haven't noticed yet, so I feel it's optional, or maybe keeping the empty block is safer.

kunli2:
I reverted the change `RequireExplicitSavingOfSecurityContextRepositoryTest` and will keep it as is.

PR #336, opened 2023-04-13
Revert "Add Spring Security 5.8 migration to Spring Framework 5.3 and…

nmck257:
… Spring Boot 2.7"

This reverts commit 6f91bde39a3d48ee5cd8253468770935d47a6972.

Hey @knutwannheden -- let me know if you have some context behind this commit?
AFAIK, no release line of Spring Boot uses Spring Security 5.8; SB 2.7 is on SS 5.7, and SB 3.0 is on SS 6.0. So, I wouldn't expect the SS 5.8 migration to get called until SB 3.0 calls SS 6.0. SB 2.7 seems too early.

And, by having SS 5.8 in the SF 5.3 recipe, it actually gets called all the way back in the SB *2.4* recipe, which would normally be using SS 5.4. Way early.


knutwannheden (approved):
Makes sense. I am not sure where I got that from then. But since I have very limited Spring knowledge I am happy that you picked up on this. Thanks!

nmck257:
Cool - thanks for confirming!

Issue #275, opened 2023-01-23
superfluous version tag added to all dependencies


martinlippert:
When upgrading a project to Spring Boot 2.7 (`2.7.8` at the moment) results in a version tag inserted into every Spring Boot dependency in the pom.xml file, which should not be the case, since the parent defines that version already:

```
    <dependency>
      <groupId>org.springframework.boot</groupId>
      <artifactId>spring-boot-starter-actuator</artifactId>
    </dependency>
```

becomes

```
    <dependency>
      <groupId>org.springframework.boot</groupId>
      <artifactId>spring-boot-starter-actuator</artifactId>
      <version>2.7.8</version>
    </dependency>
```

timtebeek:
Hi @martinlippert ; Could you tell me a bit more about the version and recipe that you used? Because I'm having trouble replicating the issue here. I downloaded a Maven 2.7.x starter project; _manually downgraded that to 2.7.7_.
```
https://start.spring.io/#!type=maven-project&language=java&platformVersion=2.7.8&packaging=jar&jvmVersion=17&groupId=com.example&artifactId=demo&name=demo&description=Demo%20project%20for%20Spring%20Boot&packageName=com.example.demo&dependencies=actuator
```

Then I ran the following command to perform a Spring Boot upgrade within the 2.7.x range.
```
mvn org.openrewrite.maven:rewrite-maven-plugin:4.39.0:run   \
  -Drewrite.recipeArtifactCoordinates=org.openrewrite.recipe:rewrite-spring:4.32.0   \
  -DactiveRecipes=org.openrewrite.java.spring.boot2.UpgradeSpringBoot_2_7
```

Which only changed the parent version property for me to 2.7.8, as I would expect; no superfluous version

Notice that I'm using the plugin & rewrite-spring versions released a few days ago; we are still updating the documentation.

martinlippert:
Let me double check the latest version, I was probably using the previous one due to that being integrated into the tooling on our side.

martinlippert:
Double checked with the latest release, looks all good... :-)

timtebeek:
Glad to hear & thanks for confirming!

martinlippert:
Hey @timtebeek, I am re-opening this issue since I came across this again (using the latest `4.32.0` release of `rewrite-spring`) and I think I found the circumstances under which this problem appears.

[...]

PR #338, opened 2023-04-19
Do not override managed dependencies (#275)

thomaszub:


timtebeek:
Thanks a lot @thomaszub ; Intuitively I'd say this makes sense for most cases where people use a direct Spring parent POM. I'm thinking how this would affect those using an intermediate company parent POM that in turn depends on Spring parent POM. 🤔 
Don't have time to dive in fully today, but hoping @kunli2 can see the through and deploy the fix if it works.

knutwannheden (approved):
LGTM. Thanks!

PR #341, opened 2023-04-20
ImplicitWebAnnotationNames - auto-format variable declaration when an argument is removed.

### LINKED ISSUE #340: ImplicitWebAnnotationNames should auto-format the variable declaration if an annotation argument has been removed
Variable declarations should be auto-formatted after removing annotation arguments to prevent compile errors.

The following example demonstrates how removing the only argument from an annotation associated with a variable declaration having an empty prefix causes  a compile error. 

before:
```java
@GetMapping("/{id}")
public ResponseEntity<String> getUser(@PathVariable("id")Long id) {}
```
expected:
```java
@GetMapping("/{id}")
public ResponseEntity<String> getUser(@PathVariable Long id) {}
```

when the variable declaration is not formatted:
```java
@GetMapping("/{id}")
public ResponseEntity<String> getUser(@PathVariableLong id) {}
```

pway99:
Auto-format variable declaration when an argument is removed. Fixes #340

### DIFF
--- a/src/main/java/org/openrewrite/java/spring/ImplicitWebAnnotationNames.java
+++ b/src/main/java/org/openrewrite/java/spring/ImplicitWebAnnotationNames.java
@@ -77,6 +77,12 @@ public class ImplicitWebAnnotationNames extends Recipe {
         ).map(className -> "org.springframework.web.bind.annotation." + className).collect(toSet());
 
         @Override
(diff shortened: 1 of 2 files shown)


timtebeek:
Fails due to a new spring boot release.. :/ about to head out, but can fix that later if you want me to pick that up.

pway99:
> Fails due to a new spring boot release.. :/ about to head out, but can fix that later if you want me to pick that up.

Thanks for digging into that, I can fix it later this evening unless you get to it sooner

PR #23, opened 2021-02-12
7.x upgrade ConditionalOnBeanAnyNestedCondition Recipe

pway99:

PR #359, opened 2023-05-23
Removed ConditionalOnBeanAnyNestedCondition from Spring Boot Best Pra…

nmck257:
…ctices

As far as I can tell, replacing `ConditionalOnBean` with this `AnyNestedCondition` construct is actually a semantically meaningful change. The former is an `and`, and the latter is an `or`. So, this recipe should likely be used not as a best practice but instead as a targeted transformation.

https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/autoconfigure/condition/ConditionalOnBean.html
https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/autoconfigure/condition/AnyNestedCondition.html

### DIFF
--- a/src/main/resources/META-INF/rewrite/spring-boot-20.yml
+++ b/src/main/resources/META-INF/rewrite/spring-boot-20.yml
@@ -672,6 +672,5 @@ recipeList:
   - org.openrewrite.java.spring.ImplicitWebAnnotationNames
   - org.openrewrite.java.spring.boot2.UnnecessarySpringExtension
   - org.openrewrite.java.spring.NoAutowiredOnConstructor
-  - org.openrewrite.java.spring.boot2.ConditionalOnBeanAnyNestedCondition
   - org.openrewrite.java.spring.boot2.RestTemplateBuilderRequestFactory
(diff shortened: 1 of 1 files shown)


timtebeek:
Only seeing this one now; thanks for adding context here! Are you saying we might actually have to change the recipe to use `AllNestedConditions` instead? That might be the correct replacement from what I gather here.

nmck257:
I think `AllNestedConditions` would be an accurate replacement, but I'm not sure that it's necessarily an improvement: `ConditionalOnBean` isn't deprecated, and I'm not aware of any reason to disuse it

timtebeek:
I'll go ahead and merge this pull request as is then; best not to apply a recipe by default that we're not sure of.

Question remains open if `ConditionalOnBeanAnyNestedCondition` should actually be `ConditionalOnBeanAllNestedCondition`, or whether we even should have any such recipe. It was originally added in #23 ; with only @sambsnyd still on board of the developers involved at the time. No background as to what prompted the recipe at the time. Any insights there Sam?

PR #361, opened 2023-05-23
Feature/isolate upgrades

nmck257:
implementation of #351

on a private project (using recipes for a private spring-boot-wrapper framework), I was able to reduce recipe execution time from 6m52s to 3m06s using a local snapshot of rewrite-spring with this change. The (private) recipe chain had just 4 calls to `UpgradeSpringBoot_###` recipes.

these changes should even bring some small performance improvements to the `UpgradeSpringBoot` recipes directly, as I was able to untangle some redundant execution of spring-data, spring-framework, spring-security, etc recipes across the chain,

### DIFF
--- a/src/main/resources/META-INF/rewrite/spring-boot-21.yml
+++ b/src/main/resources/META-INF/rewrite/spring-boot-21.yml
@@ -41,6 +41,7 @@ description: >
 tags:
   - spring
   - boot
+  - hidden
 recipeList:
   - org.openrewrite.java.spring.framework.UpgradeSpringFramework_5_1_Only
   - org.openrewrite.maven.UpgradeDependencyVersion:
--- a/src/main/resources/META-INF/rewrite/spring-boot-21.yml
+++ b/src/main/resources/META-INF/rewrite/spring-boot-21.yml
@@ -30,6 +30,19 @@ tags:
   - boot
 recipeList:
   - org.openrewrite.java.spring.boot2.UpgradeSpringBoot_2_0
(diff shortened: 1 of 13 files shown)


jkschneider:
@nmck257 I'd ask that we instead add applicable tests to recipes like `Spring Boot 2.5` to check if the project is even on a version below 2.5 before executing.

nmck257:
@jkschneider
recapping some details from our discussion:
* I also would like to preserve the scenario where a project is already on (eg) 3.0, but we've improved the 2.7 recipe with some new features (eg handling some new obscure deprecation), and the project can pick up those changes on a future run of the 3.0 (or 3.1, etc) recipe
   * You're thinking we sacrifice that behavior and solve use cases like that with ad-hoc recipe campaigns, which I can see but am less excited about
* this pattern might be simplified by some abstract `FrameworkUpgradeRecipe` which can be configured with a guardrail condition visitor and a set of migrations to apply (ie a recipelist) -- ultimately, this would benefit from being able to configure fully-qualified recipes as options to another recipe in the yaml syntax (especially with the removal of declarative applicability tests in 8.0)

ultimately, I'd like to get to a solution where recipes like this:
* can use the declarative syntax for the list of (sub)recipes
* have some default guardrail to prevent redundant execution of older version recipes (eg the applicability test you proposed)
* have some capability to override that guardrail, and forcibly run/skip "the chain" of older version recipes

I'll keep thinking it over on my side to see if I can achieve all my goals there, and on your side, it sounds like we're mostly convinced about adding the support for recipes-as-options in yaml, which should help

nmck257:
Withdrawing this PR; it'll stay in the history if we ever need it again

PR #368, opened 2023-06-06
Basic Boot 3.1 Upgrade

BoykoAlex:
The base for Boot 3.1 upgrade.

Pom upgrade. Aggregate recipe with Pom and properties upgrade recipes.


timtebeek (approved):
Thanks a lot!

Issue #374, opened 2023-06-12
Plugin versions are not adapted for gradle UpgradeSpringBoot


Jenson3210:
## What version of OpenRewrite are you using?
I am using Gradle plugin 6.0.0

## How are you running OpenRewrite? 

- Gradle plugin
- [a7439c7 commit of spring-petclinic](https://github.com/spring-projects/spring-petclinic/tree/a7439c74ea718c4f59fe6c7c643c4afe59d7e718) as this is the first commit with gradle on petclinic repo

Can you share your configuration so that we can rule out any configuration issues? 

- gradle 7.4
- MAC OS

### Adapt build gradle to run rewrite: 
```groovy
plugins {
  id 'org.springframework.boot' version '2.6.1'
  id 'io.spring.dependency-management' version '1.0.11.RELEASE'
  id 'java'
  id 'org.openrewrite.rewrite' version '6.0.0'
}

rewrite {
    activeRecipe(
            'org.openrewrite.java.spring.boot3.UpgradeSpringBoot_3_0',
    )
}

//...
dependencies {
  rewrite platform('org.openrewrite.recipe:rewrite-recipe-bom:2.0.0')
  rewrite('org.openrewrite.recipe:rewrite-spring')
  ...
}
```

### Run openrewrite

Launch `gradle rewriteRun`. 
Wait for build succeeded message.

**Expected behaviour**

Plugin versions are adapted

**Actual behaviour**

Plugin versions are not adapted
```groovy
plugins {
  id 'org.springframework.boot' version '2.6.1' // <-- Version not adapted
  id 'io.spring.dependency-management' version '1.0.11.RELEASE' // <-- Version not adapted
  //...
}
```

## Are you interested in [contributing a fix to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#contributing-fixes)?
Sure, if guided properly I am always open to take "newbie" tickets!

knutwannheden:
@Jenson3210 Thanks for your report! The OpenRewrite Spring recipes only recently started supporting Gradle-based projects, so we are very glad to receive feedback in any form regarding this.

PR #382, opened 2023-06-27
Upgrade Gradle wrapper and Spring Gradle plugins

shanman190:


shanman190:
Fixes gh-374.

sambsnyd (approved):
Thanks! Looks like the build is complaining about missing license headers

shanman190:
So fixed those. The present failure appears to be an OOM with classgraph. Do you have any thoughts on how you'd like to proceed with that one?

timtebeek:
I think we'd need a release of openrewrite/rewrite to pick up https://github.com/openrewrite/rewrite/commit/65d4c7341810a6013f93a6efeb506f82521ddcbf , since we depend on a snapshot there, which the release then pins down the latest release.

shanman190:
Yeah, these changes will be dependent upon that commit. I don't see anything wrong with releasing rewrite proper though.

PR #389, opened 2023-07-17
Add a separate "Migrate to Spring Framework 6.0" recipe

timtebeek:
## What's changed?
Add a separate "Migrate to Spring Framework 6.0" recipe

## What's your motivation?
Such that we can support people not (yet) using Spring Boot.

- [ ] Are we OK with moving and slightly changing the order of  `UpgradeToJava17` and `JavaxMigrationToJakarta`?
- [ ] Any risk to `UpgradeSpringFramework_6_0` _also_ including `UpgradeSpringFramework_5_3`, which is [also in `spring-boot-24.yml`](https://github.com/openrewrite/rewrite-spring/blob/c40f54efaf0f4011011ec5e903cc65cb02559397/src/main/resources/META-INF/rewrite/spring-boot-24.yml#L33) ?

We could not support this explicitly, in which case we can likely remove the now empty `src/main/resources/META-INF/rewrite/spring-framework-60.yml`.

## Any additional context
This is in line with what we did for previous Spring Framework versions, but wasn't tackled separately for Spring Boot 3.


joanvr (approved):

src/main/resources/META-INF/rewrite/spring-boot-30.yml
```diff
@@ -53,7 +53,6 @@ recipeList:
   - org.openrewrite.gradle.UpdateGradleWrapper:
       version: ^7.4
       addIfMissing: false
-  - org.openrewrite.java.migrate.jakarta.JavaxMigrationToJakarta
```
timtebeek, line 56: Hmm; might this move now introduce dependencies that would otherwise be available transitively?

timtebeek:
Hmm; quite uncertain now about the order potentially negatively affecting checks on availability of transitive dependencies, so while this might have seemed a good idea, in reality I don't think we want to change this too much now. It should be rare that people run Spring Framework 6 without Spring Boot, and if they do they can always [compose together recipes themselves in a custom rewrite.yml](https://docs.openrewrite.org/reference/yaml-format-reference).

PR #394, opened 2023-07-23
fix: adapt J.Modifier and J.Identifier instantiation to use the new c…

SimonVerhoeven:
…onstructor

## What's changed?

Switched J.Modifier and J.Identifier instantiation to use the new constructors to resolve the build issues.

## What's your motivation?

To resolve the build issues

no

### Checklist
- [x] I've updated the documentation (if applicable)


timtebeek:
The failing test is unrelated to this change and something we will look into separately; thanks a lot for making these changes!

Issue #300, opened 2023-03-03
Upgrade to HttpClient5 for HttpComponentsClientHttpRequestFactory in Spring 6


nmck257:
> NOTE: Requires Apache HttpComponents 5.1 or higher, as of Spring 6.0.
https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/http/client/HttpComponentsClientHttpRequestFactory.html

HttpComponents did a GAV change with 5, so projects with explicit dependencies on this library may be stuck on 4.x and fall into the compatibility trap documented above.

This recipe could possibly have an applicability test for projects using HttpComponentsClientHttpRequestFactory, to decrease risk of performing this upgrade (and introducing some other breaking change) on projects which don't imminently need it.

MahatmaFatalError:
FWIW https://chat.openai.com/share/3b804a18-c795-41e6-8254-335e340e9347 I asked chatGPT to do the job ;)

aholland:
I'm assuming this is the same issue I have just seen, which is that in apply the recipe to upgrade to Spring Boot 3.2, the dependency 
`    implementation 'org.apache.httpcomponents:httpclient'` 
was not replaced with 
`    implementation 'org.apache.httpcomponents.client5:httpclient5'` 
and I had to do it manually.

If this is not the same issue (?) I can create a separate new issue.

timtebeek:
As an update here: we've spin off the Apache HttpClient recipes into rewrite-apache:
https://github.com/openrewrite/rewrite-apache/tree/main/src/main/java/org/openrewrite/apache

To resolve this issue we'd need to add a dependency on `rewrite-apache` here and link the migration into the Spring Boot 3 migration, which should be fairly straightforward. Anyone welcome to pitch in!

PR #397, opened 2023-07-27
first approach of HttpClient5 migration

joanvr:
## What's changed?
Added a first approach of Apache Http Client 5 migration. Right now it adds the new dependency coordinates and performs all the obvious mappings between compatible classes changing the namespace.
Some attempts on some Deprecated classes have been done, but probably with some breaking changes right now. Will need more iteration.

## What's your motivation?
Spring boot 3 migration requires HttpClient upgrade.
Issue: #300 

@timtebeek

### DIFF
--- a/src/main/resources/META-INF/rewrite/apache-httpclient-5.yml
+++ b/src/main/resources/META-INF/rewrite/apache-httpclient-5.yml
@@ -123,10 +123,6 @@ recipeList:
   - org.openrewrite.java.ChangeType:
       oldFullyQualifiedTypeName: org.apache.hc.client5.http.impl.classic.BasicCredentialsProvider
       newFullyQualifiedTypeName: org.apache.hc.client5.http.impl.auth.BasicCredentialsProvider
-  # Fixing Depreacted
-  - org.openrewrite.java.ChangeType:
-      oldFullyQualifiedTypeName: org.apache.hc.client5.http.impl.classic.DefaultHttpClient
-      newFullyQualifiedTypeName: org.apache.hc.client5.http.impl.classic.CloseableHttpClient
 
   - org.openrewrite.java.ChangePackage:
       oldPackageName: org.apache.http.auth
@@ -377,10 +373,6 @@ recipeList:
   - org.openrewrite.java.ChangeType:
       oldFullyQualifiedTypeName: org.apache.hc.client5.http.HttpConnectionFactory
       newFullyQualifiedTypeName: org.apache.hc.core5.http.io.HttpConnectionFactory
-  # Fixing Deprecated
-  - org.openrewrite.java.ChangeType:
-      oldFullyQualifiedTypeName: org.apache.hc.client5.http.ClientConnectionManager
-      newFullyQualifiedTypeName: org.apache.hc.client5.http.io.HttpClientConnectionManager
 
   - org.openrewrite.java.ChangePackage:
       oldPackageName: org.apache.http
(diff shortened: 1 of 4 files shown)


joanvr:
I'm using this document as a support to map 4->5 classes and track deprecations and errors:
https://docs.google.com/spreadsheets/d/1jFKxcDL-kCab_6-bDBweoimLZhKbu755ugxg5A6SJ0E/edit?usp=sharing

src/main/resources/META-INF/rewrite/apache-httpclient-5.yml
```diff
@@ -0,0 +1,453 @@
...
+      oldArtifactId: httpclient
+      newGroupId: org.apache.httpcomponents.client5
+      newArtifactId: httpclient5
+      newVersion: 5.1.x
+      overrideManagedVersion: true
+
+  - org.openrewrite.java.ChangePackage:
+      oldPackageName: org.apache.http.client.methods
+      newPackageName: org.apache.hc.client5.http.classic.methods
+      recursive: "False"
```
timtebeek, line 56: [False seems to be the default](https://github.com/openrewrite/rewrite/blob/bb9f47a7947a6ad6ebbe9dcece329856f3d1963f/rewrite-java/src/main/java/org/openrewrite/java/ChangePackage.java#L56-L60C13), so I propose we remove this here and everywhere further down such that it stands out when actually true.
```suggestion
```

timtebeek:
I'd say a good start; maybe good to complement with just the basics of a unit test before we merge, but otherwise we're mostly testing existing recipes and how they work when combined.

src/main/resources/META-INF/rewrite/apache-httpclient-5.yml
joanvr: I originally omitted the recursive: False, expecting to be the default value, but when I got weird results (because I was having ChangeType before ChangePackage), I got crazy and add it everywhere just to check if it was the reason... And then forgot about it 😅 
Gonna remove it now!

timtebeek (approved):
Great start & thanks for adding the test!

src/main/resources/META-INF/rewrite/apache-httpclient-5.yml
```diff
@@ -0,0 +1,406 @@
...
+  - org.openrewrite.java.ChangeType:
+      oldFullyQualifiedTypeName: org.apache.hc.client5.http.impl.classic.DefaultUserTokenHandler
+      newFullyQualifiedTypeName: org.apache.hc.client5.http.impl.DefaultUserTokenHandler
+  - org.openrewrite.java.ChangeType:
+      oldFullyQualifiedTypeName: org.apache.hc.client5.http.impl.classic.BasicCredentialsProvider
+      newFullyQualifiedTypeName: org.apache.hc.client5.http.impl.auth.BasicCredentialsProvider
+  # Fixing Depreacted
+  - org.openrewrite.java.ChangeType:
+      oldFullyQualifiedTypeName: org.apache.hc.client5.http.impl.classic.DefaultHttpClient
+      newFullyQualifiedTypeName: org.apache.hc.client5.http.impl.classic.CloseableHttpClient
```
timtebeek, line 129: Is the top one still available? Maybe best not to replace just yet in that case.

timtebeek: Because if we replace this early, then we might not be able to better target a replacement that introduces try-with-resources later.

joanvr: nope, it's not available anymore... But I'm thinking on not just facing this "deprecated" issues yet, because in general is not as simple as just changing the type, since usually they don't have the same methods available...

PR #400, opened 2023-08-01
Replace getStatusLine on HttpResponse

joanvr:
## What's changed?
HttpResponse::getStatusLine has been removed in Apache HttpClient 5.x, you can call the getStatusCode, getProtocolVersion and getReasonPhrase directly on the HttpResponse object.
When applicable, we will apply a simplify method chain, but for scenarios where the StatusLine is stored or used directly, we are going to instantiate a StatusLine object (which has the same previous 4.x methods) with the response object.

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/apache/httpclient5/NewStatusLine.java
+++ b/src/main/java/org/openrewrite/java/apache/httpclient5/NewStatusLine.java
@@ -24,7 +24,7 @@ import org.openrewrite.java.tree.J;
 public class NewStatusLine extends Recipe {
     @Override
     public String getDisplayName() {
-        return "Replaces deprecated HttpResponse::getStatusLine()";
+        return "Replaces deprecated `HttpResponse::getStatusLine()`";
     }
 
     @Override
--- a/src/main/java/org/openrewrite/java/apache/httpclient5/NewStatusLine.java
+++ b/src/main/java/org/openrewrite/java/apache/httpclient5/NewStatusLine.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2023 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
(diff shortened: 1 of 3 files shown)


src/main/java/org/openrewrite/java/apache/httpclient5/NewStatusLine.java
```diff
@@ -0,0 +1,59 @@
...
+    public String getDisplayName() {
+        return "Replaces deprecated HttpResponse::getStatusLine()";
+    }
+
+    @Override
+    public String getDescription() {
+        return "HttpResponse::getStatusLine() was deprecated in 4.x, so we replace it for new StatusLine(HttpResponse). " +
+                "Ideally we will try to simplify method chains for getStatusCode, getProtocolVersion and getReasonPhrase, " +
+                "but there are some scenarios wher ethe StatusLine object is assigned or used directly, and we need to " +
+                "instantiate the object.";
```
timtebeek, line 35:
```suggestion
        return "`HttpResponse::getStatusLine()` was deprecated in 4.x, so we replace it for `new StatusLine(HttpResponse)`. " +
                "Ideally we will try to simplify method chains for `getStatusCode`, `getProtocolVersion` and `getReasonPhrase`, " +
                "but there are some scenarios where the `StatusLine` object is assigned or used directly, and we need to " +
                "instantiate the object.";
```

src/main/java/org/openrewrite/java/apache/httpclient5/NewStatusLine.java
```diff
@@ -0,0 +1,59 @@
...
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.Recipe;
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.java.*;
+import org.openrewrite.java.tree.J;
+
+public class NewStatusLine extends Recipe {
+    @Override
+    public String getDisplayName() {
+        return "Replaces deprecated HttpResponse::getStatusLine()";
```
timtebeek, line 27:
```suggestion
        return "Replaces deprecated `HttpResponse::getStatusLine()`";
```

timtebeek (approved):

PR #401, opened 2023-08-01
Added Migrate DefaultHttpClient & some cookie constants

joanvr:
## What's changed?
Migrate Deprecated DefaultHttpClient to CloseableHttpClient + builder for instantiation. Right now only covering the default scenario of no custom HttpParams or ConnectionManager, since it's non-deterministic how to solve the migration for all generic scenarios.

Also added some migration of Cookie constants

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/apache/httpclient4/MigrateDefaultHttpClient.java
+++ b/src/main/java/org/openrewrite/java/apache/httpclient4/MigrateDefaultHttpClient.java
@@ -32,7 +32,7 @@ public class MigrateDefaultHttpClient extends Recipe {
     @Override
     public String getDescription() {
         return "Since DefaultHttpClient is deprecated, we need to change it to the CloseableHttpClient. " +
-                "Only covering the default scenario with no custom HttpParams or ConnectionManager.";
+                "It only covers the default scenario with no custom HttpParams or ConnectionManager.";
     }
 
     @Override


src/main/java/org/openrewrite/java/apache/httpclient4/MigrateDefaultHttpClient.java
```diff
@@ -0,0 +1,62 @@
...
+import org.openrewrite.Recipe;
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.java.*;
+import org.openrewrite.java.search.UsesType;
+import org.openrewrite.java.tree.J;
+
+public class MigrateDefaultHttpClient extends Recipe {
+    @Override
+    public String getDisplayName() {
+        return "Migrates Deprecated DefaultHttpClient";
```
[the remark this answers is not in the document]
timtebeek: Same for the description below: mixed casing is frowned upon, but for class names for instance it's fine if those are in backticks.

src/main/java/org/openrewrite/java/apache/httpclient4/MigrateDefaultHttpClient.java
```diff
@@ -0,0 +1,62 @@
...
+    public String getDescription() {
+        return "Since DefaultHttpClient is deprecated, we need to change it to the CloseableHttpClient. " +
+                "Only covering the default scenario with no custom HttpParams or ConnectionManager.";
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return Preconditions.check(new UsesType<>("org.apache.http.impl.client.DefaultHttpClient", false), new JavaVisitor<ExecutionContext>() {
+            MethodMatcher noArgsMatcher = new MethodMatcher("org.apache.http.impl.client.DefaultHttpClient <constructor>()");
+            JavaTemplate noArgsTemplate = JavaTemplate.builder("HttpClients.createDefault()")
```
timtebeek, line 42:
```suggestion
            final MethodMatcher noArgsMatcher = new MethodMatcher("org.apache.http.impl.client.DefaultHttpClient <constructor>()");
            final JavaTemplate noArgsTemplate = JavaTemplate.builder("HttpClients.createDefault()")
```

src/main/resources/META-INF/rewrite/apache-httpclient-4-5.yml
```diff
@@ -0,0 +1,53 @@
...
+  changes between versions.
+tags:
+  - apache
+  - httpclient
+recipeList:
+  - org.openrewrite.java.dependencies.UpgradeDependencyVersion:
+      groupId: org.apache.httpcomponents
+      artifactId: httpclient
+      newVersion: 4.5.x
+      overrideManagedVersion: true
```
timtebeek, line 35: Why the override here? I imagine that might accidentally pin a managed version that would otherwise be fixed with a parent upgrade. Is there anything in the 4.5.x range that requires an upgrade to 4.5.x specifally, like replacement classes added in the last versions?

timtebeek: If override of managed versions is not strictly necessary I'd opt for the default here.
```suggestion
```

src/main/resources/META-INF/rewrite/apache-httpclient-4-5.yml
```diff
@@ -0,0 +1,53 @@
...
+  - org.openrewrite.java.apache.httpclient4.MappingDeprecatedClasses
+---
+type: specs.openrewrite.org/v1beta/recipe
+name: org.openrewrite.java.apache.httpclient4.MappingDeprecatedClasses
+displayName: Maps deprecated classes from ApacheHttpClient 4.5.x to Newer ones
```
timtebeek, line 40:
```suggestion
displayName: Maps deprecated classes from Apache HttpClient 4.5.x to suggested replacements
```

timtebeek (approved):

src/main/resources/META-INF/rewrite/apache-httpclient-4-5.yml
joanvr: The idea is to migrate to the latest version of 4.5.x before moving to 5.x, since there are some extra deprecated classes and their replacements in 4.5.x that disappear in 5.x

joanvr: We are also overriding managed versions when moving to 5.x. I guess it's either necessary on both places, or neither. WDYT?

timtebeek: I try not to pin versions when possible, as that might get projects in a state that's slightly more cumbersome to move away from, if they use a shared or Spring Boot parent for managed versions. And since we change the GAV in the upgrade from 4 to 5 I don't think we need to override managed versions there either.

src/main/java/org/openrewrite/java/apache/httpclient4/MigrateDefaultHttpClient.java
```diff
@@ -0,0 +1,62 @@
...
+public class MigrateDefaultHttpClient extends Recipe {
+    @Override
+    public String getDisplayName() {
+        return "Migrates deprecated `DefaultHttpClient`";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Since DefaultHttpClient is deprecated, we need to change it to the CloseableHttpClient. " +
+                "Only covering the default scenario with no custom HttpParams or ConnectionManager.";
```
rpau, line 35:
```suggestion
                "It only covers the default scenario with no custom HttpParams or ConnectionManager.";
```

src/main/resources/META-INF/rewrite/apache-httpclient-4-5.yml
```diff
@@ -0,0 +1,52 @@
...
+#
+
+########################################################################################################################
+# Apache HttpClient 4.5.x
+---
+type: specs.openrewrite.org/v1beta/recipe
+name: org.openrewrite.java.apache.httpclient4.UpgradeApacheHttpClient_4_5
+displayName: Migrates to ApacheHttpClient 4.5.x
+description: >
+  Migrate applications to the latest Apache HttpClient 4.5.x release. This recipe will modify an
```
rpau, line 24:
```suggestion
  Migrate applications to the latest Apache HttpClient 4.5.x release. This recipe modifies 
```

src/main/resources/META-INF/rewrite/apache-httpclient-4-5.yml
```diff
@@ -0,0 +1,52 @@
...
+  application's build files, make changes to deprecated/preferred APIs, and migrate configuration settings that have
```
rpau, line 25:
```suggestion
  application's build files, make changes to deprecated/preferred APIs, and migrates configuration settings that have
```

rpau (approved):
Minor documentation improvements. Otherwise 👍

PR #402, opened 2023-08-01
`HttpSecurityLambdaDsl` and `ServerHttpSecurityLambdaDsl` are overly eager when running against very old Spring Security versions

shanman190:
Also added/fixed the precondition setup such that the recipe now only runs over `JavaSourceFile` instances that use the specific `HttpSecurity` type.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -150,6 +150,9 @@ dependencies {
     "testWithSpringBoot_2_1RuntimeOnly"("org.springframework.boot:spring-boot-actuator:2.1.0.RELEASE")
     "testWithSpringBoot_2_1RuntimeOnly"("org.springframework.data:spring-data-jpa:2.1.0.RELEASE")
     "testWithSpringBoot_2_1RuntimeOnly"("javax.persistence:javax.persistence-api:2.2")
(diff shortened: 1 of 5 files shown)


shanman190:
Two aspects that I think we should fix in future PRs are:

* Reliance on `JavaType.FullyQualified#getMethods()` as this requires reparsing the source code and rerunning the recipe in order to make the desired changes
* Investigate when Spring Security changes the recommendation of using the standard methods to the `Customizer` enabled ones. In particular, these methods show up in Spring Security 5.2, but we don't enable them until Spring Security 5.7.

timtebeek:
Thanks a lot for these fixes! I'd briefly explored if the test you added fails to make changes for the right reasons, which it appears to do. The wrong reasons not to make changes could be a missing type leading to the recipe not executed; I've seen that in the past, and it's tricky to assert & prevent, but the debugger indicates the recipe is executed here.

It looks like [Spring Security 7+ will drop the Lambda DSL ](https://docs.spring.io/spring-security/reference/migration-7/configuration.html), so these recipes will surely need to be part of that migration, although they're already part of the chain.

PR #403, opened 2023-08-02
AddTimeUnitArgument recipe

joanvr:
## What's changed?
Added a new recipe AddTimeUnitArgument to solve the timeout and timevalue migrations. This way is more maintainable since it only requires changes in the declarative yml recipe instead of changing the recipe themselves.

### Checklist
- [x] I've updated the documentation (if applicable)


src/main/java/org/openrewrite/java/apache/httpclient5/AddTimeUnitArgument.java
```diff
@@ -0,0 +1,93 @@
...
+                "`TimeUnit.MILLISECONDS`, but you can specify this as a parameter too if needed. Since all affected methods of " +
+                "the Apache Http Client 5.x migration only have one integer/long argument, it will only apply with matched method " +
+                "invocations of exactly one parameter.";
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return new JavaIsoVisitor<ExecutionContext>() {
+            final MethodMatcher matcher = new MethodMatcher(methodPattern);
+            final JavaTemplate template = JavaTemplate.builder("#{any(long)}, TimeUnit.#{}")
```
timtebeek, line 71: I'm not so sure about assuming the target method only has one argument, and that target argument is a long. I don't think either is necessary to make as an assumption in the recipe implementation; we could just take what ever arguments the method had and add the timeunit at the end.

src/main/java/org/openrewrite/java/apache/httpclient5/AddTimeUnitArgument.java
```diff
@@ -0,0 +1,93 @@
...
+    TimeUnit timeUnit;
+
+    @Override
+    public String getDisplayName() {
+        return "Adds a TimeUnit argument to the matched method invocations";
+    }
+
+    @Override
+    public String getDescription() {
+        return "In Apache Http Client 5.x migration, we need to add a TimeUnit argument to the timeout and duration methods. " +
```
rpau, line 59:
```suggestion
        return "In Apache Http Client 5.x migration, an extra TimeUnit argument is required in the timeout and duration methods. " +
```

src/main/java/org/openrewrite/java/apache/httpclient5/AddTimeUnitArgument.java
```diff
@@ -0,0 +1,93 @@
...
+                "Previously in 4.x, all this methods were implicitly having the timeout or duration expressed in milliseconds, " +
+                "but in 5.x we explicitly need to indicate the unit of the timeout or duration. So, by default it will add " +
```
rpau, line 61:
```suggestion
                "but in 5.x the unit of the timeout or duration is required. So, by default this recipe adds " +
```

src/main/java/org/openrewrite/java/apache/httpclient5/AddTimeUnitArgument.java
```diff
@@ -0,0 +1,93 @@
...
+                "`TimeUnit.MILLISECONDS`, but you can specify this as a parameter too if needed. Since all affected methods of " +
+                "the Apache Http Client 5.x migration only have one integer/long argument, it will only apply with matched method " +
```
rpau, line 63:
```suggestion
                "the Apache Http Client 5.x migration only have one integer/long argument, the recipe applies with matched method " +
```

src/test/java/org/openrewrite/java/apache/httpclient5/AddTimeUnitArgumentTest.java
```diff
@@ -0,0 +1,155 @@
...
+
+    @Test
+    void addTimeUnitDefaultMilliseconds() {
+        rewriteRun(
+          spec -> spec.recipe(new AddTimeUnitArgument("A method(int)", null)),
+          //language=java
+          java("""
+            import java.util.concurrent.TimeUnit;
+            
+            class A {
```
timtebeek, line 35: It looks like we repeat this class three times; would it make sense to extract this to a constant? That helps focus the tests on the replacements and recipe in/output instead.

src/main/resources/META-INF/rewrite/apache-httpclient-5.yml
```diff
@@ -415,6 +413,22 @@ recipeList:
       newMethodName: setResponseTimeout
 ---
 type: specs.openrewrite.org/v1beta/recipe
+name: org.openrewrite.java.apache.httpclient5.UpgradeApacheHttpClient_5_TimeUnit
+displayName: Add `TimeUnit` to timeouts and duration methods
```
rpau, line 417:
```suggestion
displayName: Adds `TimeUnit` to timeouts and duration methods
```

src/main/java/org/openrewrite/java/apache/httpclient5/AddTimeUnitArgument.java
```diff
@@ -68,21 +69,25 @@ public String getDescription() {
...
 
             @Override
             public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, ExecutionContext executionContext) {
                 J.MethodInvocation m = super.visitMethodInvocation(method, executionContext);
-                if (matcher.matches(method) && method.getArguments().size() == 1) {
+                if (matcher.matches(m)) {
+                    JavaTemplate template = JavaTemplate
+                            .builder(StringUtils.repeat("#{any()}, ", m.getArguments().size()) + "TimeUnit.#{}")
+                            .contextSensitive()
+                            .javaParser(JavaParser.fromJavaVersion().classpath("httpclient5", "httpcore5"))
```
timtebeek, line 80: Curious to see if this would work without this import; we're already using `contextSensitive`.. Maybe a quick experiment quick drop this requirement and make the recipe more broadly applicable? Unless this results in missing method type information.
```suggestion
```

joanvr: actually, the unit tests of the individual class, in which the parser of the JavaTemplate does not have visibility to the original A class, and does not know their methods, it's missing some method type information...
Without contextSensitive and adding the classpath of httpclient and httpcore, it's also missing proper type attribution in the output of the template for the migration recipe:
 
Captura de pantalla 2023-08-02 a las 12 25 13

timtebeek: Thanks for checking! Wasn't sure how much that `contextSensitive` pulls in; good to keep this as is then!

src/main/java/org/openrewrite/java/apache/httpclient5/AddTimeUnitArgument.java
```diff
@@ -0,0 +1,98 @@
...
+                            .imports("java.util.concurrent.TimeUnit")
+                            .build();
+
+                    List<Object> arguments = new ArrayList<>(m.getArguments());
+                    arguments.add(timeUnit != null ? timeUnit : TimeUnit.MILLISECONDS);
+
+                    m = template.apply(
+                            updateCursor(m),
+                            m.getCoordinates().replaceArguments(),
```
timtebeek, line 89: I'm wondering what happens to the method type when we don't replace the method, but only add an argument; wouldn't the method type post-recipe still refer to the original (usually single argument) method instead of the new multiple argument method? I'd always thought that `replaceArguments` was only when you replace with the same number and type or arguments; not when you switch to a different method.

If the method type is updated correctly there's no need for a change, otherwise it might be best to replace the whole method rather than just the arguments.

timtebeek (approved):
Already approved, but do look at my last question regarding replace arguments versus replace method and the resulting method type.

Issue #337, opened 2023-04-14
Simplify unnecessary MediaType.parse calls


nmck257:
> ```java
> MediaType.parse("application/json")
> // Could instead be
> MediaType.APPLICATION_JSON
> ```

As discussed in: https://github.com/openrewrite/rewrite-spring/issues/325#issuecomment-1508284641

timtebeek:
Good to track this separately; we might also want to include the case where a `MediaType` constant is used as a value, since we're migrating users towards that as of #325

PR #404, opened 2023-08-03
Simplify `MediaType.parse` Calls

AlekSimpson:
## What's changed?
This PR introduces a recipe to replaces uses of `MediaType.parse("application/json")` with `MediaType.APPLICATION_JSON`. 

## What's your motivation?
Fixes https://github.com/openrewrite/rewrite-spring/issues/337

Currently when the recipe runs the `methodType` for `MediaType.parse` is null which is causing the recipe to stop running early. I think it is because I am not including the correct classpath. I have tried using `classpath("spring-web")` but it kept telling me that this was not a valid resource name. I'm confused why it would tell me that since I've seen `spring-web` used in other tests in rewrite-spring.

@timtebeek 

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/spring/http/SimplifyMediaTypeParseCalls.java
+++ b/src/main/java/org/openrewrite/java/spring/http/SimplifyMediaTypeParseCalls.java
@@ -1,0 +1,47 @@
+package org.openrewrite.java.spring.http;
+
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.Recipe;
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.java.JavaTemplate;
+import org.openrewrite.java.JavaVisitor;
+import org.openrewrite.java.MethodMatcher;
+import org.openrewrite.java.tree.J;
+
+public class SimplifyMediaTypeParseCalls extends Recipe {
+    @Override
+    public String getDisplayName() {
+        return "Simplify Unnecessary `MediaType.parse` calls";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Replaces `MediaType.parse('application/json')` with `MediaType.APPLICATION_JSON`.";
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return new SimplifyParseCallsVisitor();
+    }
+
+    private static final class SimplifyParseCallsVisitor extends JavaVisitor<ExecutionContext> {
+        private final MethodMatcher MEDIATYPE_PARSE_MATCHER = new MethodMatcher("org.springframework.http.MediaType parse(..)");
+
+        @Override
+        public J visitMethodInvocation(J.MethodInvocation mi, ExecutionContext ctx) {
+            System.out.println(mi.getSimpleName());
+            if (!MEDIATYPE_PARSE_MATCHER.matches(mi) && !mi.getArguments().isEmpty()) {
+                return mi;
+            }
+
+            J.Literal test = (J.Literal) mi.getArguments().get(0);
+            if (!test.getValue().equals("application/json")) {
+                return mi;
+            }
+
+            return JavaTemplate.builder("MediaType.APPLICATION_JSON")
+                    .build()
+                    .apply(getCursor(), mi.getCoordinates().replace());
+        }
+    }
+}
--- a/src/main/java/org/openrewrite/java/spring/http/SimplifyMediaTypeParseCalls.java
+++ b/src/main/java/org/openrewrite/java/spring/http/SimplifyMediaTypeParseCalls.java
@@ -16,22 +16,19 @@
 package org.openrewrite.java.spring.http;
 
 import org.openrewrite.ExecutionContext;
+import org.openrewrite.Preconditions;
 import org.openrewrite.Recipe;
 import org.openrewrite.TreeVisitor;
-import org.openrewrite.java.JavaTemplate;
 import org.openrewrite.java.JavaVisitor;
 import org.openrewrite.java.MethodMatcher;
-import org.openrewrite.java.TypeMatcher;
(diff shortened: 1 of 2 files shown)


src/main/java/org/openrewrite/java/spring/http/SimplifyMediaTypeParseCalls.java
```diff
@@ -0,0 +1,61 @@
...
+            }
+
+            J.Literal test = (J.Literal) mi.getArguments().get(0);
+            if (!test.getValue().equals("application/json")) {
+                return mi;
+            }
+
+            return JavaTemplate.builder("MediaType.APPLICATION_JSON")
+                    .build()
+                    .apply(getCursor(), mi.getCoordinates().replace());
```
timtebeek, line 58: Probably better to instead call out to https://docs.openrewrite.org/recipes/java/reordermethodarguments in a doAfterVisit; that way any (unlikely) comments and whitespace are preserved.

AlekSimpson: I'm not sure I understand, why would I do a call out to `ReorderMethodArguments` here? From what I understand of the recipe its for rearranging the arguments in a method call. However, in this case the method call is just being replaced with a the call to `APPLICATION_JSON` which doesn't have any arguments? Maybe I am just not familiar enough with the `ReorderMethodArguments` recipe.

timtebeek: Ah sorry; had this one mixed up with the other problematic classpath PR!

joanvr:
The reason why it's returning `null` is because the method `parse(String)` does not actually exist in `MediaType`. I understand that you are using this method name because of the original issue, but if you check at the [docs](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/http/MediaType.html) it seems that the method you are looking for is called [`parseMediaType(String)`](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/http/MediaType.html#parseMediaType(java.lang.String)) or maybe [`valueOf(String)`](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/http/MediaType.html#valueOf(java.lang.String)).
Maybe it's a typo? Or the API just changed at some point? I could not find any reference to `MediaType.parse(String)` from Spring in a fast google search.

src/main/java/org/openrewrite/java/spring/http/SimplifyMediaTypeParseCalls.java
```diff
@@ -40,22 +40,21 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
...
+        private final MethodMatcher MEDIATYPE_VALUE_OF_MATCHER = new MethodMatcher("org.springframework.http.MediaType valueOf(String)");
 
         @Override
         public J visitMethodInvocation(J.MethodInvocation mi, ExecutionContext ctx) {
-            if (!MEDIATYPE_PARSE_MATCHER.matches(mi) && !mi.getArguments().isEmpty()) {
-                return mi;
+            if (MEDIATYPE_PARSE_MEDIA_TYPE_MATCHER.matches(mi) || MEDIATYPE_VALUE_OF_MATCHER.matches(mi)) {
+                J.Literal test = (J.Literal) mi.getArguments().get(0);
+                if ("application/json".equals(test.getValue())) {
+                    return JavaTemplate.builder("MediaType.APPLICATION_JSON")
```
timtebeek, line 51: This is perhaps too limiting; we already have a recipe to replace literals with constants:
- https://github.com/openrewrite/rewrite-spring/issues/325

A comment there was logged as an improvement to pick up cases where we (after the above recipe) pass a constant into a parse method:
- https://github.com/openrewrite/rewrite-spring/issues/337

If you ask me this implementation should detect String constants passed into parse, and replace those with the pre existing MediaType constants. Open to discuss.

AlekSimpson: Ok so just to make sure I understand this correctly instead of detecting something like `"application/json"` like we are doing now; We should instead detect constants and make the replacement then?

Are there are other `MediaType` constants this recipe should try covering also? Instead of just looking for `APPLICATION_JSON`?

timtebeek: yes when the value feeding into the parse call is a constant defined on `MediaType` then you want to replace the whole method invocation with the sister constant.

timtebeek: so
```diff
-MediaType.parseMediaType(MediaType.APPLICATION_JSON_VALUE)
+MediaType.APPLICATION_JSON
```
notice how we drop the method call, and replace it with a "sister constant" that lacks the `_VALUE` suffix.

timtebeek: the above is from memory, so be sure to check what's available on the MediaType class in your IDE, but that's the rough outline of what I would expect. And we should not limit ourselves to `APPLICATION_JSON`, instead just any constants that follow that pattern (without naming each explicitly).

AlekSimpson:
Ok I updated this recipe to work not just with `APPLICATION_JSON_VALUE` but also with the other constants that `MediaType` has.

timtebeek:
Make a small change not to use a hardcoded list of variables, instead using the convention that if there's a String constant defined on MediaType that there's a matching constant that lacks the `_VALUE` suffix. We also don't need to use JavaTemplate for such a replacement.

timtebeek (approved):

timtebeek (approved):

PR #408, opened 2023-08-08
removed autoformat when no necessary

joanvr:
## What's changed?
Remove the call to maybeAutoFormat on some places

## What's your motivation?
Autoformat in some places it's not strictly needed, due to the produced changes, and because the limited scope of the autoFormat it can lead to unwanted results.

I've added couple of tests to check for extra scenarios, but maybe I missed some corner case where the autoFormat was needed

@timtebeek @rpau 

Maybe trying to fix the AutoFormat (more specifically the TabsAndIndentsVisitor) could be a better fix, but it seems kinda tricky and complex thing to do, with much bigger (and hard to test) impact. 

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/spring/NoRequestMappingAnnotation.java
+++ b/src/main/java/org/openrewrite/java/spring/NoRequestMappingAnnotation.java
@@ -26,6 +26,7 @@ import org.openrewrite.java.ChangeType;
 import org.openrewrite.java.JavaIsoVisitor;
 import org.openrewrite.java.search.UsesType;
 import org.openrewrite.java.tree.J;
+import org.openrewrite.java.tree.Space;
 
 import java.time.Duration;
 import java.util.Collections;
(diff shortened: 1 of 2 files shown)


src/testWithSpringBoot_2_1/java/org/openrewrite/java/spring/NoRequestMappingAnnotationTest.java
```diff
@@ -281,7 +281,7 @@ public ResponseEntity<List<String>> getUsersPost(String userId) {
               
               @RestController
               public class UsersController {
-                  @PostMapping(value = "/user/{userId}/edit", produces = {MediaType.APPLICATION_JSON_VALUE})
+                  @PostMapping(value = "/user/{userId}/edit", produces = { MediaType.APPLICATION_JSON_VALUE })
```
joanvr, line 284: This one in particular, for me, was a "side-effect" of applying always auto-format and doing more changed than needed. The original parameter was in this format, but in the result, we are removing the spaces without reason. I thing it's better to preserve original format in this recipe.

src/main/java/org/openrewrite/java/spring/NoRequestMappingAnnotation.java
```diff
@@ -93,7 +94,8 @@ public J.Annotation visitAnnotation(J.Annotation annotation, ExecutionContext ct
                 // Remove the argument
                 if (requestMethodArg.isPresent() && methodArgumentHasSingleType(requestMethodArg.get()) && resolvedRequestMappingAnnotationClassName != null) {
                     if (a.getArguments() != null) {
-                        a = maybeAutoFormat(a, a.withArguments(ListUtils.map(a.getArguments(), arg -> requestMethodArg.get().equals(arg) ? null : arg)), ctx);
+                        //a = maybeAutoFormat(a, a.withArguments(ListUtils.map(a.getArguments(), arg -> requestMethodArg.get().equals(arg) ? null : arg)), ctx);
```
timtebeek, line 97:
```suggestion
```

src/testWithSpringBoot_2_1/java/org/openrewrite/java/spring/NoRequestMappingAnnotationTest.java
timtebeek: Agreed

timtebeek (approved):

PR #410, opened 2023-08-09
removed autoFormat and replaced with manual format

joanvr:
## What's changed?
Manually formatting potential introduced errors due to removal of annotation arguments

## What's your motivation?
Calling autoFormat has some unwanted side effects, probably due to some bugs in autoFormat. However, we do not actually need to call autoFormat here, since the span of potential format changes needed is small and we can take care of it in the recipe itself.

@timtebeek

### DIFF
--- a/src/main/java/org/openrewrite/java/spring/ImplicitWebAnnotationNames.java
+++ b/src/main/java/org/openrewrite/java/spring/ImplicitWebAnnotationNames.java
@@ -21,6 +21,7 @@ import org.openrewrite.java.JavaIsoVisitor;
 import org.openrewrite.java.search.UsesType;
 import org.openrewrite.java.tree.J;
 
+import java.util.List;
 import java.util.Set;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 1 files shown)


timtebeek:
I'm missing a bit of context here; some questions that came up on my end:

- how prevalent is this issue? We don't add any tests, so I'm not familiar with when this occurs; are there any reports I'm not aware of?
- why are we only adding prefix.whitespace and not the full prefix? That could be logical depending on when this problem occurs, but I don't have that context.

We seem to be sacrificing readability a bit here; starting to wonder if we should instead hold off on a patch here and prioritize a fix in auto format itself. What are your thoughts there?

joanvr:
> I'm missing a bit of context here; some questions that came up on my end:
> 
> * how prevalent is this issue? We don't add any tests, so I'm not familiar with when this occurs; are there any reports I'm not aware of?
> * why are we only adding prefix.whitespace and not the full prefix? That could be logical depending on when this problem occurs, but I don't have that context.
> 
> We seem to be sacrificing readability a bit here; starting to wonder if we should instead hold off on a patch here and prioritize a fix in auto format itself. What are your thoughts there?

I could not reproduce in a test the original reason why I'm doing this changes in the first place, because I think they happen when another recipe sets some messages on a close cursor, that later on are picked by this auto format, producing unwanted results.
However, the idea was to remove the need of autoFormat by applying manual format because the actual needed changes are not that complex.

I agree with you that a better fix would be to fix autoFormat, but I find it kinda complex right now, and a small change there can have a greater impact that is hard to predict... So... given the urgency of this fix because of a client request, WDYT? Shall we try to fix it it here for now and try to fix it it for good in auto format later on?

timtebeek (approved):

PR #413, opened 2023-08-10
Upgrade `httpcore` -> `httpcore5` in the `UpgradeApacheHttpClient_5` recipe

rickie:
## What's changed?

Let the `UpgradeApacheHttpClient_5` recipe upgrade the `org.apache.httpcomponents:httpcore` dependency to `org.apache.httpcomponents.core5:httpcore5`. 

## What's your motivation?

When running the `UpgradeApacheHttpClient_5` we had to perform this part of the upgrade manually. With this addition, not anymore 😉. 

@timtebeek

### DIFF
--- a/src/main/resources/META-INF/rewrite/apache-httpclient-5.yml
+++ b/src/main/resources/META-INF/rewrite/apache-httpclient-5.yml
@@ -35,6 +35,12 @@ recipeList:
       newGroupId: org.apache.httpcomponents.client5
       newArtifactId: httpclient5
       newVersion: 5.1.x
+  - org.openrewrite.java.dependencies.ChangeDependency:
+      oldGroupId: org.apache.httpcomponents
+      oldArtifactId: httpcore
+      newGroupId: org.apache.httpcomponents.core5
+      newArtifactId: httpcore5
+      newVersion: 5.1.x
   - org.openrewrite.java.apache.httpclient5.UpgradeApacheHttpClient_5_ClassMapping
   - org.openrewrite.java.apache.httpclient5.UpgradeApacheHttpClient_5_DeprecatedMethods
   - org.openrewrite.java.apache.httpclient5.UpgradeApacheHttpClient_5_TimeUnit


src/main/resources/META-INF/rewrite/apache-httpclient-5.yml
```diff
@@ -35,6 +35,12 @@ recipeList:
       newGroupId: org.apache.httpcomponents.client5
       newArtifactId: httpclient5
       newVersion: 5.1.x
+  - org.openrewrite.java.dependencies.ChangeDependency:
+      oldGroupId: org.apache.httpcomponents
+      oldArtifactId: httpcore
+      newGroupId: org.apache.httpcomponents.core5
+      newArtifactId: httpcore5
+      newVersion: 5.1.x
```
rickie, line 43: Added the `5.1.x` to be in line with the `httpclient5` dependency. 

However, for both dependencies there is already a 5.2.x version out ([httpcore](https://mvnrepository.com/artifact/org.apache.httpcomponents.core5/httpcore5), [httpclient](https://mvnrepository.com/artifact/org.apache.httpcomponents.client5/httpclient5)). So not sure what to pick here.

timtebeek: Not sure if it was intentional to stay on 5.1.x for now; perhaps @joanvr knows the backstory there.

[In the 5.2+ release notes](https://downloads.apache.org/httpcomponents/httpcore/RELEASE_NOTES-5.2.x.txt) do read these notable changes

> Release 5.2
> ------------------
> This is the first GA release in the 5.2 release series. This release finalizes the 5.2 APIs
> and also corrects a number of defects discovered since the previous release.
> 
> Please note that 5.2 upgrades the minimal JRE level to version 8 (8u251 is required).
> 
> Notable changes and features included in the 5.2 series:
> 
> * Upgrade to Java 8.
> * Improved support for TLS upgrade and HTTP protocol upgrade (async).
> * Improved HTTP protocol negotiation.
> * Improved customization of connection listeners (async).
> * Improved parsing and formatting of URI components.
> * Use of Java 8 date / time APIs

   
Perhaps the Java 8 + date / time APIs makes 5.2 a bigger leap than 5.1, especially since we're still [waiting for Joda time recipes](https://github.com/openrewrite/rewrite-migrate-java/issues/234) to become available.

I'll merge this one as-is, and log an improvement issue with the above to go to 5.2 down the line.

timtebeek: Logged as https://github.com/openrewrite/rewrite-spring/issues/414

timtebeek:
Thanks again!

src/main/resources/META-INF/rewrite/apache-httpclient-5.yml
joanvr: The reason was that Spring only needed 5.1.x, to work. 
I haven't added this one originally, since it was a dependency from the new httpclient5, and since it wasn't colliding with the previous GAV coordinates I prefered to not add an extra dependency if not needed, but looking at it right now, probably it's better to just do the change to remove the old one if anybody has add it it.
Thanks for the contribution! :)

PR #418, opened 2023-08-15
Chain in AuthorizeHttpRequests with spring-security-58.yml

timtebeek:
## What's changed?
- Moved into `security5` package, as both methods are available in 5.8
- Wrap `doAfterVisit` calls inside a method, instead of always executing those
- Linked the recipe into the 5.8 spring security upgrade
- Formatting changes

## What's your motivation?
Got reports of these changes not being applied, despite a recipe being available.

- Was there any particular reason this recipe was not hooked into the yaml chain of recipes?
- Was there a reason to keep this under `boot2` instead of `security5`?

@BoykoAlex


timtebeek:
Failures unrelated: there's an OutOfMemoryError that occurs ever more frequent here.

joanvr (approved):

Issue #415, opened 2023-08-10
HttpSecurity Lambda DSL - cover nested DSLs / contentSecurityPolicy


nmck257:
Here are two (failing) test cases for `HttpSecurityLambdaDslTest`.

I haven't pored through the Spring Security code/docs to see if there are other nested configurers which we should handle, but `contentSecurityPolicy` within `headers` is definitely one.

```java
    @Test
    void contentSecurityPolicy() {
        //language=java
        rewriteRun(
          java(
            """
import org.springframework.security.config.annotation.web.builders.HttpSecurity;
import org.springframework.security.config.annotation.web.configuration.EnableWebSecurity;
import org.springframework.security.config.annotation.web.configuration.WebSecurityConfigurerAdapter;

@EnableWebSecurity
public class ConventionalSecurityConfig extends WebSecurityConfigurerAdapter {

    @Override
    protected void configure(HttpSecurity http) throws Exception {
        http
            .headers().contentSecurityPolicy("foobar");
    }
}
              """,
            """
import org.springframework.security.config.annotation.web.builders.HttpSecurity;
import org.springframework.security.config.annotation.web.configuration.EnableWebSecurity;
import org.springframework.security.config.annotation.web.configuration.WebSecurityConfigurerAdapter;

@EnableWebSecurity
public class ConventionalSecurityConfig extends WebSecurityConfigurerAdapter {

    @Override
    protected void configure(HttpSecurity http) throws Exception {
        http
            .headers(headers -> headers
                    .contentSecurityPolicy(contentSecurityPolicy -> contentSecurityPolicy
                            .policyDirectives("foobar")));
    }
}
              """
          )
        );
    }

    @Test
    void contentSecurityPolicyPartialLambda() {
        //language=java
        rewriteRun(
          java(
            """
import org.springframework.security.config.annotation.web.builders.HttpSecurity;
import org.springframework.security.config.annotation.web.configuration.EnableWebSecurity;
import org.springframework.security.config.annotation.web.configuration.WebSecurityConfigurerAdapter;

@EnableWebSecurity
public class ConventionalSecurityConfig extends WebSecurityConfigurerAdapter {

    @Override
    protected void configure(HttpSecurity http) throws Exception {
        http
            .headers(headers -> headers
                    .contentSecurityPolicy("foobar"));
    }
}
              """,
            """
import org.springframework.security.config.annotation.web.builders.HttpSecurity;
[...]

PR #419, opened 2023-08-16
Security lambda DSL enhancements

nmck257:
## What's changed?
Two things:
- moved Spring Security Lambda DSL recipes from SB 2.4 to the point at which the non-lambda DSL was deprecated: Spring Security 6.1 (new file) used by SB 3.1
- added argReplacements option to ConvertToSecurityDslVisitor, enabling migration of non-lambda-DSL overloads which accept an argument. then used that new feature to implement HeadersConfigurerLambdaDsl.

## What's your motivation?
- avoiding making changes "too early"
- #415

The changes to `ConvertToSecurityDslVisitor` haven't broken any existing tests, but, still good to spot-check.
Also, if anyone is aware of other HttpSecurity-relevant methods which can be covered now that we have the `argReplacements` mechanism, then it would be good to (minimally) define that scope.

@shanman190 @BoykoAlex @timtebeek 

### Checklist
- [X] I've added the license header to any new files through `./gradlew licenseFormat`
- [X] I've used the IntelliJ auto-formatter on affected files
- ~~[ ] I've updated the documentation (if applicable)~~

### DIFF
--- a/src/main/java/org/openrewrite/java/spring/boot2/ConvertToSecurityDslVisitor.java
+++ b/src/main/java/org/openrewrite/java/spring/boot2/ConvertToSecurityDslVisitor.java
@@ -35,13 +35,23 @@ public class ConvertToSecurityDslVisitor<P> extends JavaIsoVisitor<P> {
 
     private static final String MSG_TOP_INVOCATION = "top-method-invocation";
 
+    public static final String FQN_CUSTOMIZER = "org.springframework.security.config.Customizer";
+
     private final String securityFqn;
 
     private final Collection<String> convertableMethods;
 
+    private final Map<String, String> argReplacements;
+
     public ConvertToSecurityDslVisitor(String securityFqn, Collection<String> convertableMethods) {
+        this(securityFqn, convertableMethods, new HashMap<>());
+    }
+
+    public ConvertToSecurityDslVisitor(String securityFqn, Collection<String> convertableMethods,
+            final Map<String, String> argReplacements) {
         this.securityFqn = securityFqn;
         this.convertableMethods = convertableMethods;
+        this.argReplacements = argReplacements;
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


src/main/java/org/openrewrite/java/spring/boot2/ConvertToSecurityDslVisitor.java
```diff
@@ -137,17 +155,20 @@ private Optional<JavaType.Method> findDesiredReplacement(J.MethodInvocation m) {
                 .filter(availableMethod -> availableMethod.getName().equals(m.getSimpleName()) &&
                         availableMethod.getParameterTypes().size() == 1 &&
                         availableMethod.getParameterTypes().get(0) instanceof JavaType.FullyQualified &&
-                        "org.springframework.security.config.Customizer".equals(((JavaType.FullyQualified) availableMethod.getParameterTypes().get(0)).getFullyQualifiedName()))
+                        FQN_CUSTOMIZER.equals(((JavaType.FullyQualified) availableMethod.getParameterTypes().get(0)).getFullyQualifiedName()))
                 .findFirst();
     }
 
-    public boolean isApplicableTopLevelMethodInvocation(J.MethodInvocation m) {
```
nmck257, line 144: this method was unused (and sneaking by IDEA's "unused" inspection by being recursive)

BoykoAlex: I made this method for myself to use in the IDE integration. Can you please keep it? Perhaps a comment that this method is used in STS might be helpful

nmck257: done

BoykoAlex (approved):

PR #422, opened 2023-08-21
`disable` is a terminal function

shanman190:
## What's changed?
`disable` is similar to `and` with the important difference that it should be _included_ rather than _excluded_ from the chain.

@BoykoAlex @nmck257 @timtebeek 

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/spring/boot2/ConvertToSecurityDslVisitor.java
+++ b/src/main/java/org/openrewrite/java/spring/boot2/ConvertToSecurityDslVisitor.java
@@ -18,6 +18,7 @@ package org.openrewrite.java.spring.boot2;
 import org.openrewrite.Cursor;
 import org.openrewrite.Tree;
 import org.openrewrite.internal.StringUtils;
+import org.openrewrite.internal.lang.Nullable;
 import org.openrewrite.java.JavaIsoVisitor;
 import org.openrewrite.java.MethodMatcher;
 import org.openrewrite.java.tree.*;
@@ -181,17 +182,17 @@ public class ConvertToSecurityDslVisitor<P> extends JavaIsoVisitor<P> {
         return false;
     }
 
-    private boolean isApplicableCallCursor(Cursor c) {
-        if (c != null && c.getValue() instanceof J.MethodInvocation) {
-            J.MethodInvocation inv = c.getValue();
-            if (!TypeUtils.isOfClassType(inv.getType(), securityFqn)) {
-                return true;
-            }
-            if (new MethodMatcher("org.springframework.security.config.annotation.web.configurers.AbstractHttpConfigurer disable()").matches(inv)) {
-                return true;
-            }
+    private boolean isApplicableCallCursor(@Nullable Cursor c) {
+        if (c == null) {
+            return false;
         }
-        return false;
+
+        if (!(c.getValue() instanceof J.MethodInvocation)) {
+            return false;
+        }
+
+        J.MethodInvocation inv = c.getValue();
+        return !(isAndMethod(inv) && TypeUtils.isOfClassType(inv.getType(), securityFqn)) && !isDisableMethod(inv);
     }
 
     private List<J.MethodInvocation> computeAndMarkChain() {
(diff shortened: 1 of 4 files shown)


timtebeek:
Thanks for taking this on @shanman190 ; The visitor is shared between three different implementations, and a few of those such as 
https://github.com/openrewrite/rewrite-spring/blob/b41366a0f5ed21da7740c186af08b2bbede16dd0/src/testWithSpringBoot_2_4/java/org/openrewrite/java/spring/boot2/HttpSecurityLambdaDslTest.java#L215
[now fail](https://ge.openrewrite.org/s/sk3qkjhkr32es/tests/task/:testWithSpringBoot_2_4/details/org.openrewrite.java.spring.boot2.HttpSecurityLambdaDslTest/retainFormatting()?top-execution=1) as the `http.csrf().disable()` from before takes on the prefix from the method select. 
```diff
org.opentest4j.AssertionFailedError: [Unexpected result in "ConventionalSecurityConfig.java":
diff --git a/ConventionalSecurityConfig.java b/ConventionalSecurityConfig.java
index 2b70c9d..cf157ea 100644
--- a/ConventionalSecurityConfig.java
+++ b/ConventionalSecurityConfig.java
@@ -11,6 +11,8 @@ 
                 .antMatchers("/blog/**").permitAll()
                 .anyRequest().authenticated());
 
-        http.csrf(csrf -> csrf.disable());
+        http.csrf(csrf ->
+
+                csrf.disable());
     }
 }
```
Didn't immediately spot an easy way to fix this as we do want to retain formatting and comments in other places. 🤔  Input welcome!

src/main/java/org/openrewrite/java/spring/boot2/ConvertToSecurityDslVisitor.java
```diff
@@ -203,9 +203,14 @@ private List<J.MethodInvocation> computeAndMarkChain() {
                         .withName(initialMethodInvocation.getName().withSimpleName(methodType.getName()))));
         cursor = cursor.getParent(2);
         if (isApplicableCallCursor(cursor)) {
-            for (; cursor != null && isApplicableCallCursor(cursor); cursor = cursor.getParent(2)) {
```
nmck257, line 206: can we add a test case like `.csrf().ignoringRequestMatchers("").disable()` ?
the config before disabling is ineffective, but not unrealistic for code in the wild, I think. and since this condition to catch `disabled` isn't interacting with the cursor-moving loop at all, now, I'm suspicious that this case might not get handled cleanly.

shanman190: Added via https://github.com/openrewrite/rewrite-spring/commit/7bbd005639796a227f4765b0c4e1cd755fefa44a.

shanman190:
@timtebeek, thanks! I think I've gotten everything taken care of now. The chain computation needed a bit more fine tuning, but all three test classes are passing for me now.

src/main/java/org/openrewrite/java/spring/boot2/ConvertToSecurityDslVisitor.java
```diff
@@ -181,17 +182,17 @@ public boolean isApplicableTopLevelMethodInvocation(J.MethodInvocation m) {
...
+            return false;
         }
-        return false;
+
+        if (!(c.getValue() instanceof J.MethodInvocation)) {
+            return false;
+        }
+
+        J.MethodInvocation inv = c.getValue();
+        return !(isAndMethod(inv) && TypeUtils.isOfClassType(inv.getType(), securityFqn)) && !isDisableMethod(inv);
```
timtebeek, line 195: We're also calling `isOfClassType` in `isAndMethod`, so I'll go ahead and remove that before a merge.
```suggestion
        return !isAndMethod(inv) && !isDisableMethod(inv);
```

timtebeek (approved):

src/main/java/org/openrewrite/java/spring/boot2/ConvertToSecurityDslVisitor.java
shanman190: Good catch! 

I had to move it back during the refactoring as initially I was attempting to move toward a `MethodMatcher` based solution which wasn't checking the return type. With Spring Security there are a few instances of `and` that I found that don't inherit that method from the common interface, so I had to double back on that and missed getting it here. Oops!

PR #423, opened 2023-08-22
Update usage of list and call getItems when needed

### LINKED ISSUE #411: MigrateItemWriterWrite should update invocations of list to `Chunk.getItems()`
https://github.com/openrewrite/rewrite-spring/blob/f9fd034692c5121ea36de006638274fad0a05a92/src/main/java/org/openrewrite/java/spring/batch/MigrateItemWriterWrite.java#L32

Will correctly change the method declaration which has a `List` parameter to instead be a `Chunk`. 

Throughout the method body invocations of `List` methods also need to be updated to call `Chunk.getItems()` first

joanvr:
## What's changed?
After changing the type of the method, we can have scenarios where the method was manipulating the original List, and the used methods does not exist in Chunk. So we need to call `Chunk::getItems` to retrieve the actual list and keep compilable code.
Also calling getItems when needed on assignments, variable initialization and method invocations (if the destination type is not `java.lang.Iterable` since it's the only common interface between List and Chunk).

## What's your motivation?
Fixes #411 

Any other corner case where we should call getItems?

@timtebeek @sambsnyd 

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/spring/batch/MigrateItemWriterWrite.java
+++ b/src/main/java/org/openrewrite/java/spring/batch/MigrateItemWriterWrite.java
@@ -182,7 +182,7 @@ public class MigrateItemWriterWrite extends Recipe {
 
         private boolean notAssignableFromChunk(@Nullable JavaType type) {
             // Iterable is the only common type between List and Chunk
-            return type != null && !type.toString().startsWith(ITERABLE_FQN);
+            return !TypeUtils.isOfClassType(type, ITERABLE_FQN);
         }
 
         private boolean isParameter(J.Identifier maybeParameter) {
--- a/src/main/java/org/openrewrite/java/spring/batch/MigrateItemWriterWrite.java
+++ b/src/main/java/org/openrewrite/java/spring/batch/MigrateItemWriterWrite.java
@@ -122,17 +122,17 @@ public class MigrateItemWriterWrite extends Recipe {
         @Override
         public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, ExecutionContext executionContext) {
             J.MethodInvocation mi = super.visitMethodInvocation(method, executionContext);
-            if (LIST_MATCHER.matches(mi) && mi.getSelect() instanceof J.Identifier) {
-                if (isParameter((J.Identifier) mi.getSelect())) {
-                    assert mi.getPadding().getSelect() != null;
-                    // No need to take care of typing here, since it's going to be printed and parsed on the JavaTemplate later on.
-                    mi = mi.withSelect(newGetItemsMethodInvocation(mi.getPadding().getSelect()));
-                }
+            if (LIST_MATCHER.matches(mi) && isParameter(mi.getSelect())) {
+                assert mi.getPadding().getSelect() != null;
+                // No need to take care of typing here, since it's going to be printed and parsed on the JavaTemplate later on.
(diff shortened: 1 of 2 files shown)


src/main/java/org/openrewrite/java/spring/batch/MigrateItemWriterWrite.java
```diff
@@ -102,4 +107,105 @@ public J.MethodDeclaration visitMethodDeclaration(J.MethodDeclaration method, Ex
...
+            return a;
+        }
+
+        private boolean notAssignableFromChunk(J.VariableDeclarations.NamedVariable var) {
+            return var.getVariableType() != null && notAssignableFromChunk(var.getVariableType().getType());
+        }
+
+        private boolean notAssignableFromChunk(@Nullable JavaType type) {
+            // Iterable is the only common type between List and Chunk
+            return type != null && !type.toString().startsWith(ITERABLE_FQN);
```
timtebeek, line 185: Shouldn't we prefer TypeUtils over String comparisons here?
```suggestion
            return type != null && !TypeUtils.isOfClassType(type, ITERABLE_FQN);
```

src/main/java/org/openrewrite/java/spring/batch/MigrateItemWriterWrite.java
```diff
@@ -102,4 +107,105 @@ public J.MethodDeclaration visitMethodDeclaration(J.MethodDeclaration method, Ex
...
+        }
+
+        private boolean isParameter(J.Identifier maybeParameter) {
```
timtebeek, line 188: Rather than doing four casts from each call above, wouldn't it be easier to pull the J.Identifier type check in here?

src/main/java/org/openrewrite/java/spring/batch/MigrateItemWriterWrite.java
```diff
@@ -102,4 +107,105 @@ public J.MethodDeclaration visitMethodDeclaration(J.MethodDeclaration method, Ex
...
+        public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, ExecutionContext executionContext) {
+            J.MethodInvocation mi = super.visitMethodInvocation(method, executionContext);
+            if (LIST_MATCHER.matches(mi) && mi.getSelect() instanceof J.Identifier) {
+                if (isParameter((J.Identifier) mi.getSelect())) {
+                    assert mi.getPadding().getSelect() != null;
+                    // No need to take care of typing here, since it's going to be printed and parsed on the JavaTemplate later on.
+                    mi = mi.withSelect(newGetItemsMethodInvocation(mi.getPadding().getSelect()));
+                }
+            }
+            if (mi.getMethodType() != null && !ITEM_WRITER_MATCHER.matches(mi)) {
```
timtebeek, line 132: Matcher already does that null check.
```suggestion
            if (!ITEM_WRITER_MATCHER.matches(mi)) {
```

timtebeek (approved):
A few small comments on my end; otherwise approved. Hard to say if there's any area's we haven't covered yet.

src/main/java/org/openrewrite/java/spring/batch/MigrateItemWriterWrite.java
timtebeek: Might need an import if applied through GitHub.

timtebeek (approved):

PR #424, opened 2023-08-23
Convert `oauth2*` Spring Security subcomponents to lambda dsl

shanman190:
## What's changed?
Add lambda dsl conversions for `oauth2Login`, `oauth2Client`, and `oauth2ResourceServer` subcomponents.

## What's your motivation?
Preparing for the Spring Security 7.x migration where the lambda DSL will be required.

@timtebeek @nmck257 @BoykoAlex 

### Checklist


timtebeek (approved):
Nothing much to add other than express my thanks for both improving the visitor earlier, and now putting it to even more use.

PR #425, opened 2023-08-24
fixed usage of PomDownloader without settings

joanvr:
## What's changed?
Moved some code to inner MavenVisitor and passing the settings and profiles to the MavenPomDownloader class

## What's your motivation?
Without proper usage of maven settings, some people might find out that their configured repositories are ignored. 

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/maven/spring/UpgradeExplicitSpringBootDependencies.java
+++ b/src/main/java/org/openrewrite/maven/spring/UpgradeExplicitSpringBootDependencies.java
@@ -17,7 +17,6 @@
 package org.openrewrite.maven.spring;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
-import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import lombok.EqualsAndHashCode;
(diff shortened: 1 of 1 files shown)


timtebeek:
I'm wondering if we should update `MavenPomDownloader` to deprecate the old constructors, and add a new one that expects (or calls) getResolutionResult(). What would you think about that?

joanvr:
> I'm wondering if we should update `MavenPomDownloader` to deprecate the old constructors, and add a new one that expects (or calls) getResolutionResult(). What would you think about that?

Maybe there are scenarios where the current constructor without taking into consideration the settings is needed? I might know better when I explore all the usages that we found...

timtebeek:
> > I'm wondering if we should update `MavenPomDownloader` to deprecate the old constructors, and add a new one that expects (or calls) getResolutionResult(). What would you think about that?
> 
> Maybe there are scenarios where the current constructor without taking into consideration the settings is needed? I might know better when I explore all the usages that we found...

Sure! Let me know what you find; the reason I'd change the API around PomDownloader is that I don't want to have to look out for that in reviews in the future. Not sure if there's ever a good case to skip it, but I wouldn't think there's a good reason to bypass the user configured preferences.

PR #448, opened 2023-11-21
Add recipe to change relocated mysql dep in 2.6 migration

fabapp2:
## What's changed?

ChangeDependencyGroupIdAndArtifactId recipe was added to the 2.4 to 2.5 Boot migration recipe to change the GAV of a relocated dependency found in the Sprig PetClinic example.

## What's your motivation?

Running the migration from a 2.5 PetClinic to 2.6 the parse fails due to a relocated mysql dependency.

This is more a "hotfix" than a solution but I am in conversation with @timtebeek to provide something more solid.

### Checklist


src/main/resources/META-INF/rewrite/spring-boot-25.yml
```diff
@@ -44,6 +44,15 @@ recipeList:
       addIfMissing: false
   - org.openrewrite.java.spring.boot2.MigrateDatabaseCredentials
 
+  # Replace relocated dependency
+  - org.openrewrite.maven.ChangeDependencyGroupIdAndArtifactId:
+      oldGroupId: mysql
+      newGroupId: com.mysql
+      oldArtifactId: mysql-connector-java
+      newArtifactId: mysql-connector-j
+      newVersion: : 8.0.33
```
timtebeek, line 53: Our approach in 2.7 right now uses a wildcard, and is a little more involved; not quite sure if all of that is necessary, but we can try that out with the tests we have.
```suggestion
      newVersion: : 8.0.x
```

timtebeek (approved):

PR #455, opened 2023-12-05
HttpSecurity - replace `apply` with `with`

### LINKED ISSUE #449: Spring Security 6.2 - HttpSecurity `apply` to `with`
As part of the conversion to Lambda DSL, Spring Security 6.2 has deprecated `apply` in favor of a new method `with`:
https://docs.spring.io/spring-security/site/docs/current/api/org/springframework/security/config/annotation/AbstractConfiguredSecurityBuilder.html#apply(C)

```java
import org.springframework.security.config.annotation.web.builders.HttpSecurity;
import org.springframework.security.config.annotation.web.configurers.SessionManagementConfigurer;
import org.springframework.security.web.SecurityFilterChain;

public class MySecurityConfig {
    public SecurityFilterChain configure(HttpSecurity http) throws Exception {
        return http
                .apply(new SessionManagementConfigurer<>())
                .invalidSessionUrl("junk").and()
                .build();
    }
}
```
Note: I'm using `SessionManagementConfigurer` just to demonstrate the usage of `apply`/`with` using an existing class in Spring Security config; "real" code would likely call `http.sessionManagement(...)` instead, and only use `apply`/`with` for custom configurers.

```java
import org.springframework.security.config.annotation.web.builders.HttpSecurity;
import org.springframework.security.config.annotation.web.configurers.SessionManagementConfigurer;
import org.springframework.security.web.SecurityFilterChain;

public class MySecurityConfig {
    public SecurityFilterChain configure(HttpSecurity http) throws Exception {
        return http
                .with(new SessionManagementConfigurer<>(), sessionManagementConfigurer -> sessionManagementConfigurer
                        .invalidSessionUrl("junk"))
                .build();
    }
}
```

## Any additional context

- Should also handle injecting a `Customizer.withDefaults()` if there was nothing between `apply` and `and` in the original
- Can likely reuse some part of `ConvertToSecurityLambdaDslVisitor`, but should be sure to keep this separate from the rest of the `HttpSecurityLambdaDsl` recipe (and the reactive version) since that applies to an earlier Spring Security version, and the `with` method is new to 6.2

Yeah, if nobody else grabs it, I'll probably work on this sometime in the next few weeks.

nmck257:
Fixes #449

- Added another SourceSet for Spring Security 6.2
- The overloaded constructor options on `ConvertToSecurityDsl` are starting to feel a little clunky, but I think that with the addition of this recipe we've pretty much covered the full scope of the Lambda DSL migration, so I don't anticipate we'd need to extend this visitor much further
- (not directly related to the PR) I wonder if we could create some generic / fluent API for slicing and dicing a method invocation chain... I feel like I've seen that problem solved several times in individual recipes/visitors, and it tends to be non-trivial, but with the right abstraction, it might be reusable?...

Others who've been in this file a bunch: @shanman190 @BoykoAlex 

### Checklist

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -6,7 +6,7 @@ group = "org.openrewrite.recipe"
 description = "Eliminate legacy Spring patterns and migrate between major Spring Boot versions. Automatically."
 
 val springBootVersions: List<String> = listOf("1_5", "2_1", "2_2", "2_3", "2_4", "2_5", "2_6", "2_7", "3_0")
-val springSecurityVersions: List<String> = listOf("5_7", "5_8")
+val springSecurityVersions: List<String> = listOf("5_7", "5_8", "6_2")
 
 val sourceSetNames: Map<String, List<String>> = mapOf(
         Pair("testWithSpringBoot_", springBootVersions),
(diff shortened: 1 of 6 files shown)


shanman190 (approved):
LGTM!

PR #465, opened 2023-12-15
Refactored ConvertToSecurityDslVisitor to assemble target method type…

nmck257:
…s "manually" instead of searching for them on some existing type's available methods, to fix a bug where ApplyToWithLambdaDsl did not work for projects parsed with an older version of Spring Security (aka the expected case for apps using the recipe to upgrade)

- The resultant type information is slightly less robust with this change, but I'm pretty sure it should still be sufficient for the vast majority of recipe usages which might follow this recipe in a chain?

### Checklist


timtebeek (approved):
Thanks a lot! Indeed a common source of issues when leaping from old versions to new. I've applied a very minor change, but otherwise looks good to me.

PR #485, opened 2024-01-29
Boot 3 upgrade doesn't add XML Bind dependency

BoykoAlex:
Unit test demonstrating an issue upgrading Petclinic to Boot 3.
`jakarta.xml.bind-api` dependency should be added explicitly but it's not because `rewrite-maven` dependency resolution thinks that this dependency is present on the compile scope already for some reason via transitive dependencies.


timtebeek:
Locally I've found the test to still fail with the following, but we'll get a second opinion from CI just to be sure
```diff
diff --git a/project/pom.xml b/project/pom.xml
index 4369491..f80ac0a 100644
--- a/project/pom.xml
+++ b/project/pom.xml
@@ -30,10 +30,6 @@ 
       <artifactId>jakarta.validation-api</artifactId>
     </dependency>
     <dependency>
-      <groupId>jakarta.xml.bind</groupId>
-      <artifactId>jakarta.xml.bind-api</artifactId>
-    </dependency>
-    <dependency>
       <groupId>org.springframework.boot</groupId>
       <artifactId>spring-boot-starter-data-jpa</artifactId>
     </dependency>
```

timtebeek:
Ok CI passes; that's an improvement! I'll see about polishing this up for addition to your tests then, and figure out what's troubling locally. :)

timtebeek (approved):
Great to see this resolved, thanks a lot! Must help you as well to see this work on a common example case, finally.

timtebeek:
If I understood this comment correctly we should remove the expected added dependency here again
- https://github.com/openrewrite/rewrite-migrate-java/pull/499#issuecomment-2181164965

I'll go ahead and try that here.

timtebeek:
Thanks again @BoykoAlex ! Really great not to have those unnecessary dependencies added, and good to have a comprehensive test to guard that going forward.

BoykoAlex:
@timtebeek Thanks a lot for merging all those PRs so quickly!!!

timtebeek:
Of course! Really appreciated how you dove deep to get very targeted fixes with great effects downstream.

PR #517, opened 2024-04-29
Add property change from management.metrics.export.atlas.aggrConfig to management.atlas.metrics.export.aggrConfig in UpgradeSpringBoot_3_0 recipe

jjacobs44:
## What's changed?
Added property change from management.metrics.export.atlas.aggrConfig to management.atlas.metrics.export.aggrConfig in UpgradeSpringBoot_3_0

## What's your motivation?
Just noticed from personal experience this property change was missing in the recipe 🙂

### DIFF
--- a/src/main/resources/META-INF/rewrite/spring-boot-30.yml
+++ b/src/main/resources/META-INF/rewrite/spring-boot-30.yml
@@ -62,6 +62,7 @@ recipeList:
   - org.openrewrite.java.spring.boot3.DowngradeServletApiWhenUsingJetty
   - org.openrewrite.java.spring.boot3.ConfigurationOverEnableSecurity
   - org.openrewrite.java.spring.boot3.SpringBootProperties_3_0
+  - org.openrewrite.java.spring.boot3.MigrateAtlasAggrConfig
   - org.openrewrite.java.spring.boot3.MigrateThymeleafDependencies
   - org.openrewrite.java.spring.boot3.UpgradeSpringDoc_2
   - org.openrewrite.java.spring.boot3.MigrateDropWizardDependencies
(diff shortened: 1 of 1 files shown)


timtebeek:
Thanks for reporting that here! There's an interesting bit to this in that we generate those properties from meta data that the folks from Spring themselves make available in their starter jars, meaning it's probably missing there as well.
https://github.com/openrewrite/rewrite-spring/blob/006b595420b62765595b7acca75f15bb05bd4c8f/.github/workflows/properties.yml#L22-L26

Ideally this is fixed upstream such that it's also shown in IDEs when old properties are still used; the place for that change would be here:
https://github.com/spring-projects/spring-boot/blob/f6d8111a1a6e2da9b7743547f4d2189443caa503/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json#L463-L468

Until then we can include this property with our recipes. I'd opt to move this addition to [src/main/resources/META-INF/rewrite/spring-boot-30.yml](https://github.com/openrewrite/rewrite-spring/blob/main/src/main/resources/META-INF/rewrite/spring-boot-30.yml) though, such that it's not lost next time we generate the migration properties.

jjacobs44:
I'll get a PR opened up in Spring boot itself, and will remove these changes when that's been released. Thank you for all the helpful context! Also, happy to move things around if anything about this setup is not what you were suggesting.

timtebeek:
Awesome, thanks a lot for picking this up and also updating Spring Boot!

timtebeek (approved):

philwebb:
It turns out there's no `aggrConfig` property to migrate (see https://github.com/spring-projects/spring-boot/pull/40968#issuecomment-2145739703) so you might want to revert this one.

PR #518, opened 2024-04-29
Create spring 2.7 data YML file with getReferenceById changes

lucashan:
## What's changed?
This pull request creates a new Spring 2.7 data YML file with the recipe information for `GetReferenceById` changes.

## What's your motivation?
As of Springboot v2.7, both `getOne()` and `getById()` methods have been deprecated for `JpaRepository`. The [documentation](https://docs.spring.io/spring-data/jpa/docs/current/api/org/springframework/data/jpa/repository/JpaRepository.html) states that these deprecated methods should be replaced with `getReferenceById(ID)`.

I'm unfamiliar with how to write the unit test and update the build.gradle.kts file.

@timtebeek 

### Checklist


lucashan:
let me know if there's anything else you need from me!

src/testWithSpringBoot_2_7/java/org/openrewrite/java/spring/data/UseJpaRepositoryGetReferenceByIdTest.java
```diff
@@ -0,0 +1,79 @@
...
+              package foo;
+              public class Book {}
+              """
+          ),
+          java(
+            """
+              package foo;
+              import org.springframework.data.jpa.repository.JpaRepository;
+              public interface BookRepository extends JpaRepository<Book, Long> {
+              }
```
timtebeek, line 52: There's something odd here when I inspect this interface, in that the JPA interface we extend is not reported in the type system, leading to mismatches when we want to replace the methods.
image

Any idea what might be going on here @knutwannheden ?

timtebeek (approved):
Thanks for the help here @lucashan ! Turns out the test failure got stuck on a silly transitive dependency; fixed now.

PR #520, opened 2024-05-06
Update spring-boot-30.yml

askoog:
Add `recursive: true` to MigrateThymeleafDependencies to allow migrating all child packages as well

## What's changed?

Added `recursive: true` to `ChangePackage` for thymeleaf5 -> 6 migration to allow updating sub-packages like `org.thymeleaf.spring5.messageresolver` as well

## What's your motivation?

The migration didn't work unless I added this.

### Checklist


timtebeek (approved):
Thanks a lot for the immediate fix!

Issue #521, opened 2024-05-07
org.springframework.util.Assert in Spring Framework 6 


pativa:
## What problem are you trying to solve?

Migrating many projects to Spring Boot 3 / Spring Framework 6.
There does not seem to be any recipes available for Spring Framework 6.

Specifically, we encountered that `org.springframework.util.Assert` have removed some functions that were previously deprecated. See the example below for an example change.

The easiest way to migrate this would be with `org.openrewrite.java.AddLiteralMethodArgument`.

## Describe the situation before applying the recipe

```java
import org.springframework.util.Assert;

class A {
    void foo(String bar) {
        Assert.notNull(bar);
    }
}
```

## Describe the situation after applying the recipe

```java
class A {
    void foo(String bar) {
        Assert.notNull(bar, "must not be null");
    }
}
```

Ideally I guess the message would be something like "bar must not be null", but that would be a harder change to apply automatically.

## Are you interested in [contributing this recipe to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#adding-new-recipes)?

I have the necessary yaml changes ready and could provide that, but as there is no Spring Framework 6 migration ready, I have nowhere to put them.

So this is more more wondering if there is any interest in that I contribute these changes to the project.

timtebeek:
Hi @pativa ! Good seeing you back here; would love to include your recipe as a first step for a `spring-framework-60.yml`; we can hook that into the rest of the migrations as usual; Feel free to throw up a rough draft PR and I'll help get it in shape and merged.

Up to now we hadn't been very specific in what changes are needed for Spring Framework 6 and Spring Boot 3, but I like your suggestion of being explicit in what folks can apply, even if the Spring Boot recipes also work for Spring Framework applications.

pativa:
Thank you for the quick response.

I did an attempt at this change in https://github.com/openrewrite/rewrite-spring/pull/522

timtebeek:
Indeed fixed in
- https://github.com/openrewrite/rewrite-spring/pull/522

Thanks a lot!

PR #522, opened 2024-05-07
Spring Framework 6.0 recipe with migration for Spring `Assert`

pativa:
## What's changed?
Added a base recipe for migrating to Spring Framework 6.
This is not a complete migration, but can serve as a place to put Spring-related migrations necessary (as opposed to changes related to only Spring Boot).

## What's your motivation?
https://github.com/openrewrite/rewrite-spring/issues/521

I copied the format from the earlier `UpgradeSpringFramework_5_3`.

## Any additional context
I have tested the changes under `MigrateSpringAssert` in our internal migration and it seems to work fine.
I have taken the literal text added from the Spring class and reformatted them a bit.

Ideally, the message contain variable name but that would be a harder change to write automatically.
Usage of this is probably not too common anyway as the functions have been deprecated for a long time.

### Checklist


timtebeek (approved):
Great to see; thanks a lot!

PR #528, opened 2024-05-11
Add recipe to simplify WebTestClient methods

FieteO:
## What's changed?

Introduce a `SimplifyWebTestClientCalls` recipe that simplifies some invocations of `WebTestClient`

## What's your motivation?

Better readability

### Checklist


timtebeek:
Great to see your progress here; be sure to reach out if you'd like some more help with the final details.

FieteO:
Sure, I'll continue tomorrow. I think I will come back to you to get some help regarding inspecting the value of the `HttpStatus` without including that as an import.

I will also need to reconsider the general structure once more status codes need to be rewritten

FieteO:
Okay, I decided to only deal with the `isEqualTo(int)` signature in this PR now. The other case of `isEqualTo(HttpStatus)` is much harder to implement for me and I lack the experience here.
This way things are not getting too big and we have something mergeable already.

FieteO:
@timtebeek I assume [the tests are failing](https://github.com/openrewrite/rewrite-spring/actions/runs/9101527488/job/25041185099?pr=528#step:6:6264) due to the added `parserClasspath("org.springframework:spring-test:6.+")` dependency, right?
Shall I push all those updated jars, or how does this normally work?

timtebeek:
Thanks for the offer, but I'll pick that up from here, as we're also working through this related PR
- https://github.com/openrewrite/rewrite-spring/pull/519

Just got back from travel so I hope to get this in as soon as I'm caught up. Thanks again for the contribution!

timtebeek (approved):
Thanks a lot for this addition @FieteO ! I've made it part of our best practice recipes for Spring Boot 3.3, which I expect to release this week.

PR #538, opened 2024-06-12
Relocate Launcher classes renamed in Spring Boot 3.2

timtebeek:
## What's your motivation?
- https://github.com/spring-projects/spring-boot/issues/37667
- https://github.com/spring-projects/spring-boot/commit/c22548982abf0b9c5c02406fe789b0423fcfa02d

References to these classes could occur anywhere, but in particular in Dockerfiles & manifest files.
References from Java files should be rare; as such I didn't feel a need to add a precondition to limit where we match and replace.
We could still add a precondition if this proves troublesome, for instance when it affects pom.xml files that might refer to these classes.


timtebeek:
As reported one on one via Twitter; figured quick enough to immediately add for others that might otherwise see a broken startup.

PR #543, opened 2024-06-20
Attempt not to run boot upgrade 1.x -> 2.x for projects on 2.x already

### LINKED ISSUE #542: `jakarta.validation-api` dependency added while upgrading Spring Petclinic example
The `jakarta.validation-api` dependency shouldn't be present while upgrading Spring Petclinic to Boot 3.x. This dependency is transitive coming in via `spring-boot-validation-starter`.

The problem is that first javax `validation-api` dependency added explicitly. Then the `validation-api` is transformed into `jakarta.validation-api`.

I'd try to work with the problem of adding javax `validation-api`. It is added via 
`org.openrewrite.java.spring.boot2.UpgradeSpringBoot_2_0` -> `org.openrewrite.java.spring.boot2.MigrateHibernateConstraintsToJavax`

Namely this bit of YAML:
```yaml
  - org.openrewrite.java.dependencies.AddDependency:
      groupId: javax.validation
      artifactId: validation-api
      version: 2.x
      onlyIfUsing: javax.validation.constraints.*
```

At this moment the petclinic is on Boot 2.7.x which is on `jakarta.validation-api` version 2.0.x already. Consequently, the old `validation-api` dependency has no issues to be added.

Wonder if there is a condition in the YAML recipe that could prevent execution of recipes under 2.7.x for the 2.7.x project that we could add in spring upgrade yaml recipes...

Anything you can recommend @timtebeek and @sambsnyd ?

BoykoAlex:
Fixes #542

Used `preconditions` if **spring-core** is 4.x then we can upgrade to boot 2.x

### DIFF
--- a/src/main/resources/META-INF/rewrite/spring-boot-20.yml
+++ b/src/main/resources/META-INF/rewrite/spring-boot-20.yml
@@ -28,6 +28,11 @@ description: >
 tags:
   - spring
   - boot
+preconditions:
+  - org.openrewrite.java.dependencies.DependencyInsight:
+      groupIdPattern: org.springframework
+      artifactIdPattern: spring-core
+      version: 4.x
 recipeList:
   # Upgrade 2.0.x from 1.x
   - org.openrewrite.java.dependencies.UpgradeDependencyVersion:
--- a/src/testWithSpringBoot_2_7/java/org/openrewrite/java/spring/boot2/Boot3UpgradeTest.java
+++ b/src/testWithSpringBoot_2_7/java/org/openrewrite/java/spring/boot2/Boot3UpgradeTest.java
@@ -95,21 +95,8 @@ class Boot3UpgradeTest implements RewriteTest {
                   <properties>
                     <java.version>17</java.version>
                   </properties>
-                  <dependencyManagement>
-                    <dependencies>
-                      <dependency>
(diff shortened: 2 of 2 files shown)


src/main/resources/META-INF/rewrite/spring-boot-20.yml
```diff
@@ -28,6 +28,11 @@ description: >
 tags:
   - spring
   - boot
+preconditions:
+  - org.openrewrite.java.dependencies.DependencyInsight:
+      groupIdPattern: org.springframework
+      artifactIdPattern: spring-core
+      version: 4.x
 recipeList:
```
nmck257, line 36: I'd like to move this precondition to a narrower subrecipe -- right now, it's fully halting the "chain" of each upgrade recipe calling the previous, It also means that recipes such `SpringBoot2BestPractices` are no longer run for projects starting on newer Spring versions, which is perhaps unexpected and I'd say undesirable.

Maybe I move the precondition to `MigrateHibernateConstraintsToJavax`? Any concerns @BoykoAlex ?

BoykoAlex: I think that still makes the unit test that I've added to pass. Perhaps we can narrow down the check as well to `jakarta.validation-api` transient dependency presence.

Issue #541, opened 2024-06-20
Maven dependency added with invalid classifier when using LocalServerPort


eocantu:
## What version of OpenRewrite are you using?

I am using

- OpenRewrite v8.28.1
- rewrite-spring v5.13.2
- Recipe: org.openrewrite.java.spring.boot3.UpgradeSpringBoot_3_2

## How are you running OpenRewrite?

I am using the Maven plugin, and my project is a multi-module project: https://github.com/eocantu/spring-boot-demo

## What is the smallest, simplest way to reproduce the problem?

The project referenced above is a minimalistic multi-module project using `LocalServerPort`. The error can be recreated as follows:
```bash
git clone https://github.com/eocantu/spring-boot-demo
cd spring-boot-demo
./mvnw rewrite:run
```

## What did you expect to see?
I expected to not see a dependency added, or if it was added, not added with a classifier.

## What did you see instead?
```xml
        <dependency>
            <groupId>org.springframework.boot</groupId>
            <artifactId>spring-boot-starter-web</artifactId>
            <version>3.2.7</version>
            <classifier>org.springframework.boot.web.server.LocalServerPort</classifier>
            <scope>test</scope>
        </dependency>
```

## What is the full stack trace of any errors you encountered?
No errors observed

timtebeek:
Thanks for the quick reproducer! Looking at this recipe I think the arguments have shifted such that we accidentally pass in `LocalServerPort` as a classifier. Likely an easy fix. Is that something you'd be willing to contribute?
https://github.com/openrewrite/rewrite-spring/blob/24af0a4b05b8620028611bc1dfc15fa981a02eb7/src/main/java/org/openrewrite/java/spring/boot2/MigrateLocalServerPortAnnotation.java#L64-L74

PR #544, opened 2024-06-22
Remove classifier on dependency added migrating LocalServerPort

eocantu:
## What's changed?

Removes classifier setting on dependency added when migrating LocalServerPort

## What's your motivation?

Fixes https://github.com/openrewrite/rewrite-spring/issues/541

### Checklist


timtebeek (approved):
Thanks a lot for making not just the code change, but also adding a test to verify correct handling going forward. I've added some small touch ups to verify the dependency is added, with the correct scope. Good to see this fixed.

PR #553, opened 2024-07-19
fix: Updating docs to explain vulnerabilities

pstreef:


README.md
```diff
@@ -24,6 +24,16 @@ See the full documentation at [docs.openrewrite.org](https://docs.openrewrite.or
 3. Commit & push changes.
 4. Repeat periodically as new minor versions of Spring Boot are released.
 
+## Why do artifact scanners detect vulnerabilities in recipe artifacts/JARs?
+
+In order to modernize and upgrade old or vulnerable code, some OpenRewrite recipe modules bundle copies of old libraries. Libraries bundled into recipe modules are never executed. That being said, these libraries are never executed.
```
timtebeek, line 29: The last two sentences end in `never executed`. Can we rewrite that without that duplication?

README.md
```diff
@@ -24,6 +24,16 @@ See the full documentation at [docs.openrewrite.org](https://docs.openrewrite.or
...
+
+OpenRewrite exercises the Java compiler internally to compile code patterns that exist in these old and/or vulnerable libraries. These patterns are then used to match old or vulnerable code for the sake of modernizing or repairing it.
+
+Using a library in compilation in this way does not trigger class initialization in the way that reflection might, for example. In other words, code paths in libraries used in compilation are never executed.
```
timtebeek, line 33: `never executed` a third time here; Perhaps we could emphasize not vulnerable?

README.md
pstreef: while I agree I just copy pasted the docs here:
https://docs.moderne.io/administrator-documentation/moderne-platform/references/faq#why-do-artifact-scanners-detect-vulnerabilities-in-recipe-artifacts-jars

and rewrote a little as this is in the rewrite-spring context.

pstreef: I removed 1 and added a little more text behind another. Hope thats better?

timtebeek (approved):
Yes thanks! Approved already. Tagging @mike-solomon to see if we should remove the duplication elsewhere too.

PR #555, opened 2024-07-22
Spring Boot 2.0 - moving the precondition to run in more cases

nmck257:
## What's changed?

narrowing the precondition on the SB 2.0 recipe to ensure it runs more consistently.

## What's your motivation?

https://github.com/openrewrite/rewrite-spring/pull/543/files#r1678212731

the new precondition is better, but does have a false positive if running the 2.2 recipe on a 2.2 project and a false negative if running the 2.7 recipe on a 2.7 project. this is hard to avoid. the javax.validation-api dep is necessary for SB 2.3 (when hibernate-validator 6.1 swaps its transitive dep from javax to jakarta) through SB 2.7 until SB 3.0 (when spring swaps to jakarta). and with this recipe's placement, it has to make an educated guess about the "final" version of SB for the project after all other recipes in the chain complete while only knowing the current version.

@BoykoAlex 

a nice counterbalance to this recipe might be if the `rewrite-migrate-java` javax->jakarta migration removed direct jakarta deps which are also available transitively, after it converts javax deps to jakarta. this would help untangle a pom which previously used `spring-boot-starter-validation` plus `javax.validation-api`. in the current state, the javax->jakarta recipe leaves such a pom with a redundant `jakarta.validation-api` which the starter already has covered.

### Checklist

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -144,6 +144,7 @@ dependencies {
     "testWithSpringBoot_1_5RuntimeOnly"("org.springframework.boot:spring-boot:1.5.+")
     "testWithSpringBoot_1_5RuntimeOnly"("org.springframework.boot:spring-boot-autoconfigure:1.5.+")
     "testWithSpringBoot_1_5RuntimeOnly"("org.springframework.boot:spring-boot-test:1.5.+")
+    "testWithSpringBoot_1_5RuntimeOnly"("org.hibernate.validation:hibernate-validator:1.5.+")
     "testWithSpringBoot_1_5RuntimeOnly"("org.hamcrest:hamcrest:2.2")
     "testWithSpringBoot_1_5RuntimeOnly"("junit:junit:latest.release")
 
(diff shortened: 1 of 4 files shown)


build.gradle.kts
```diff
@@ -144,6 +144,7 @@ dependencies {
     "testWithSpringBoot_1_5RuntimeOnly"("org.springframework.boot:spring-boot:1.5.+")
     "testWithSpringBoot_1_5RuntimeOnly"("org.springframework.boot:spring-boot-autoconfigure:1.5.+")
     "testWithSpringBoot_1_5RuntimeOnly"("org.springframework.boot:spring-boot-test:1.5.+")
+    "testWithSpringBoot_1_5RuntimeOnly"("org.hibernate.validation:hibernate-validator:1.5.+")
```
BoykoAlex, line 147: The build fails. Try:
```
    "testWithSpringBoot_1_5RuntimeOnly"("org.hibernate:hibernate-validator:5.3.+")
```

BoykoAlex (approved):

PR #563, opened 2024-08-12
Separating application.properties into different files based on profile

ishish07:
## What's changed?

I added 1 recipe to split application.properties files into multiple application-profile.properties files if additional profiles are specified and separated by either [#--- or !---](https://docs.spring.io/spring-boot/reference/features/external-config.html#features.external-config.files.multi-document:~:text=For%20application.properties%20files%20a%20special%20%23%2D%2D%2D%20or%20!%2D%2D%2D%20comment%20is%20used%20to%20mark%20the%20document%20splits%3A)

This recipe also deletes the additional profile configurations in the application.properties file itself as it would be replaced by a new file.

## What's your motivation?

During my internship this summer, I had to manually create new application-environment.properties files based on the profiles and/or environments that were specified in application.properties. This was during our migration to Java 21. We were doing this for organizational purposes.

Anything under  #--- or !--- in application.properties belongs to a different document. If there are global properties under #--- or !---, those configurations will not be applied to other profiles and/or environments and could cause bugs. By running this recipe and separating application-env.properties files, it becomes obvious if there are missing global properties in application.properties.

I would like the reviewers to focus on whether the file path to the newly created application-env.properties files is accurate. These new files belong in the same folder that application.properties exists in.

I would also like the reviewers to see if there is any way to add a helpful comment when appending properties into an existing application-profile.properties file. I append new content on line 92 in the appendToExistingPropertiesFile method. I was not sure how to add a comment as we are dealing with Properties.Content instances. It would be nice to add something like 

"# Below are the properties from application.properties!"

The third thing I'd like reviewers to focus on is whether there is an easy way to have the code do nothing if no application.properties file is found. I was trying to use Precondition.check in getScanner but I wasn't able to get it to work. There are a few if statements littered around the code just to deal with the case that the application.properties file doesn't exist, so it would make the code a lot more clean if it could just do nothing when it detects that in the scanner.

Thank you!

## Any additional context

Most likely, respositories will not maintain a profile specific section in application.properties along with an existing application-profile.properties file. However, I felt it would be safe to keep this edge case in mind. I wrote the recipe to safely append content to its corresponding pre-existing application-profile.properties file - otherwise the recipe would run and harmfully delete configurations.  

### Checklist


timtebeek:
Thanks a lot @ishish07 ! I like how you've sketched the context of having had to do this by hand, and now wish to automate this for others going forward. I'll need to take a closer look & go over the focus areas you've identified, but expect we can get this in somewhere this week.

src/main/java/org/openrewrite/java/spring/SeparateApplicationPropertiesByProfile.java
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    private Map<String, List<Properties.Content>> getNewApplicationPropertyFileInfo(List<Properties.Content> contentList) {
+        Map<String, List<Properties.Content>> map = new HashMap<>();
+        int index = 0;
+        while (index < contentList.size()) {
+            if (isSeparator(contentList.get(index))) {
+                List<Properties.Content> newContent = getContentForNewFile(contentList, ++index);
+                map.put("application-" + ((Properties.Entry) newContent.get(0)).getValue().getText() + ".properties",
+                        newContent.subList(1, newContent.size()));
```
timtebeek, line 141: I think we should use the full target file name as key here, since there could be a multi module project that could clash.
```
parent
  - childA
    - src/main/resource/application.properties
  - childB
    - src/main/resource/application.properties
```

timtebeek (commented):
Great start here @ishish07 ! Thanks for your patience in the review. I've pushed up a few polishing commits already, but think we have some work to do still where multi module projects are involved. Are you ok to make that change still?

src/main/java/org/openrewrite/java/spring/SeparateApplicationPropertiesByProfile.java
ishish07: I believe the code already adjusts for files within nested folders like your example (line 206 of the testing file). I also do not think that a project can have multiple application.properties files. 

The method above - getNewApplicationPropertyFileInfo is simply for getting the content of each application-env.properties file from application.properties. Application-env.properties files that don't exist will, by default, be placed in the same folder as the application.properties file.

timtebeek: hi! I've just pushed a test to highlight what I meant; https://github.com/openrewrite/rewrite-spring/pull/563/commits/4f232974640180041c78307cde8c156c1fef9538
The test at line 206 only had a single folder structure, not two separate.

Given the scale at which recipes can run I think we should account for such a scenario.

src/main/java/org/openrewrite/java/spring/SeparateApplicationPropertiesByProfile.java
```diff
@@ -0,0 +1,234 @@
...
+                ((((Properties.Comment) c).getDelimiter() ==
+                        Properties.Comment.Delimiter.valueOf("HASH_TAG")) ||
+
+                        ((Properties.Comment) c).getDelimiter() ==
+                                Properties.Comment.Delimiter.valueOf("EXCLAMATION_MARK"));
+    }
+
+    public static class Accumulator {
+        // Map from module path to its property file info
+        Map<String, ModulePropertyInfo> moduleProperties = new HashMap<>();
```
timtebeek: Making a quick note for myself here: Elsewhere we use `Map<JavaProject, ....>` when working with scanning recipes across multi modules projects, and use the JavaProject marker that's available as opposed to a String key. Welcome to adjust to using that already, or we can try to fit that in with a review.

PR #566, opened 2024-08-14
Include Apache Httpclient4 to 5 in Spring Framework 6 upgrade

Laurens-W:
## What's changed?

Include Apache HttpClient4 to HttpClient5 migration with Spring framework 6 migration 

## What's your motivation?

https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-3.1-Release-Notes#dependency-management-for-apache-httpclient-4

> Support for Apache HttpClient 4 with RestTemplate was [removed in Spring Framework 6](https://github.com/spring-projects/spring-framework/issues/28925), in favor of Apache HttpClient 5. Spring Boot 3.0 includes dependency management for both HttpClient 4 and 5. Applications that continue to use HttpClient 4 can experience [errors when using RestTemplate](https://github.com/spring-projects/spring-boot/issues/33515) that are difficult to diagnose.
> 
> Spring Boot 3.1 removes dependency management for HttpClient 4 to encourage users to move to HttpClient 5 instead.
> 

While technically this would only be required when using Apache HttpClient with RestTemplate, the 3.1 release notes already hint at dependency management being dropped for httpclient 4.x

## Any additional context
https://docs.spring.io/spring-boot/3.3/appendix/dependency-versions/coordinates.html
image

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -112,7 +112,6 @@ val rewriteVersion = rewriteRecipe.rewriteVersion.get()
 
 dependencies {
     implementation(platform("org.openrewrite:rewrite-bom:${rewriteVersion}"))
-    implementation("org.openrewrite:rewrite-apache")
     implementation("org.openrewrite:rewrite-java")
     implementation("org.openrewrite:rewrite-xml")
     implementation("org.openrewrite:rewrite-properties")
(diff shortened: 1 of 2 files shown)


timtebeek (approved):
Nice addition! Unlocks a next step.

Issue #567, opened 2024-08-14
The Junit4to5Migration recipe fails to migrate @ClassRule (of type SpringClassRule)


FieteO:
## What version of OpenRewrite are you using?
2.16.0

## How are you running OpenRewrite?
I am applying the Maven Command Line command on a single module project.

Can you share your configuration so that we can rule out any configuration issues?
I am not explicitly changing any configuration.

Is your project public? If so, can you share a link to it?

## What is the smallest, simplest way to reproduce the problem?

```java
import org.springframework.boot.test.context.SpringBootTest;
import org.springframework.test.context.junit4.rules.SpringClassRule;
import org.springframework.test.context.junit4.rules.SpringMethodRule;

@SpringBootTest
public class SomeTest {

    @ClassRule
    public static final SpringClassRule springClassRule = new SpringClassRule();

    @Rule
    public final SpringMethodRule springMethodRule = new SpringMethodRule();

}
```

## What did you expect to see?

According to [Spring Docs - Spring Extension for JUnit Jupiter](https://docs.spring.io/spring-framework/reference/testing/testcontext-framework/support-classes.html#testcontext-junit-jupiter-extension), `SpringClassRule` and `SpringMethodRule` are replaced by `@ExtendWith(SpringExtension.class)`.
There are two scenarios:
- `@SpringBootTest` is present on the test: `SpringClassRule` and `SpringMethodRule` should be removed. `@ExtendWith(SpringExtension.class)` is not needed since `@SpringBootTest` adds it already
- `@SpringBootTest` is not present on the test: `SpringClassRule` and `SpringMethodRule` should be removed. `@ExtendWith(SpringExtension.class)` should be added.

```java
import org.springframework.boot.test.context.SpringBootTest;

@SpringBootTest
public class SomeTest {

}
```

## What did you see instead?
No change to the class.
```java
import org.springframework.boot.test.context.SpringBootTest;
import org.springframework.test.context.junit4.rules.SpringClassRule;
import org.springframework.test.context.junit4.rules.SpringMethodRule;

@SpringBootTest
public class SomeTest {

    @ClassRule
    public static final SpringClassRule springClassRule = new SpringClassRule();

    @Rule
    public final SpringMethodRule springMethodRule = new SpringMethodRule();

}
```

## Are you interested in [contributing a fix to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#contributing-fixes)?

[...]

PR #571, opened 2024-08-14
JUnit4 SpringClassRule and SpringMethodRule to SpringExtension

Laurens-W:
## What's changed?

Added a recipe to migrate from JUnit 4 Spring Rules to JUnit 5 ExtendWith

## What's your motivation?

- Fixes https://github.com/openrewrite/rewrite-spring/issues/567

@timtebeek @FieteO 

## Any additional context

I'm wondering whether there'd be any harm in annotating the class with `@SpringBootTest` instead of `@ExtendWith(SpringExtension.class)`

### Checklist


FieteO:
Regarding your concern with `@SpringBootTest`:
I think you are right, there is a difference: `@SpringBootTest` influences context scanning, while `@Extend(SpringExtension.class)` doesn't. Having that in mind, using `@ExtendWith(...)` is the safer thing to do.

timtebeek (approved):
Good addition; thanks! Also thanks to @FieteO for the suggestion and helpful context in the review.

Issue #540, opened 2024-06-18
`HttpComponentsClientHttpRequestFactory.setReadTimeout` was removed in Spring 6.1.0


timtebeek:
## What problem are you trying to solve?
`HttpComponentsClientHttpRequestFactory.setReadTimeout` is removed in Spring 6.1.0 (spring boot 3.2.0).

## What precondition(s) should be checked before applying this recipe?
Uses method.

## Describe the situation before applying the recipe
TODO

## Describe the situation after applying the recipe
TODO

## Any additional context
- https://docs.spring.io/spring-framework/docs/6.0.0/javadoc-api/org/springframework/http/client/HttpComponentsClientHttpRequestFactory.html#setReadTimeout(int)
- https://github.com/spring-projects/spring-boot/issues/32857
- https://github.com/spring-projects/spring-boot/commit/e6c37d698f4cece58ca3eb0ae5442a2df32af8a9

PR #572, opened 2024-08-14
Migrate `HttpComponentsClientHttpRequestFactory.setReadTimeout` to `SocketConfig.setSoTimeout`

Laurens-W:
## What's changed?

Added a conservative recipe to migrate from `HttpComponentsClientHttpRequestFactory.setReadTimeout` to `SocketConfig.setSoTimeout`.

## What's your motivation?

- Fixes https://github.com/openrewrite/rewrite-spring/issues/540

Test coverage

@timtebeek 

## Any additional context

For now the recipe assumes a `PoolingHttpClientConnectionManager` is defined within the same class as the `HttpComponentsClientHttpRequestFactory`. It doesn't take into account that a `PoolingHttpClientConnectionManagerBuilder` exists nor does it refactor to this.

### Checklist


src/main/resources/META-INF/rewrite/classpath/httpcore5-5.2.5.jar
timtebeek, line 1: So a thing to note here is that these classpath resources end up in the packaged jar we send out to folks ; right now we're adding four jars, together weighing in at 2.8 MB; there's also alternatives that do not require any addition to the classpath for tests: https://github.com/openrewrite/rewrite-spring/blob/729164fd713cbd0bfce44e582b0d613f4d1965b0/src/test/java/org/openrewrite/java/spring/test/SpringRulesToJUnitExtensionTest.java#L34

Let's strive to be conservative into what we add as jars, and how we add those. Preferably in src/test when we can, and automatically through `parserClasspath` entries and the `./gradlew downloadRecipeDependencies` task, as opposed to manually.

src/test/java/org/openrewrite/java/spring/framework/HttpComponentsClientHttpRequestFactoryReadTimeoutTest.java
```diff
@@ -0,0 +1,110 @@
...
+        );
+    }
+
+    // TODO Additional scenarios not yet covered
+    // - Using BasicHttpClientConnectionManager
+    // - Using PoolingHttpClientConnectionManagerBuilder
+    // - No HttpClientConnectionManager at all
+    // - No intermediate variable for connectionManager
+    // - setReadTimeout called with local variable not accessible near connection manager
+    // - there already is a setDefaultSocketConfig call
```
timtebeek, line 108: As indicated here the recipe right now is very specific to a particular test scenario; there's various others that we might want to cover as well, and a few that are quite hard to cover well. Let's do a best effort pass through these to see which we can still reasonably implement, and then merge this first iteration.

src/main/java/org/openrewrite/java/spring/framework/HttpComponentsClientHttpRequestFactoryReadTimeout.java
```diff
@@ -0,0 +1,83 @@
...
+                    public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx) {
+                        if (METHOD_MATCHER.matches(method)) {
+                            Expression expression = method.getArguments().get(0);
+                            doAfterVisit(new JavaIsoVisitor<ExecutionContext>() {
+                                @Override
+                                public J.Block visitBlock(J.Block block, ExecutionContext ctx) {
+                                    for (Statement statement : block.getStatements()) {
+                                        if (statement instanceof J.VariableDeclarations &&
+                                            TypeUtils.isAssignableTo(POOLING_HTTP_CLIENT_CONNECTION_MANAGER,
+                                                    ((J.VariableDeclarations) statement).getTypeAsFullyQualified())) {
```
timtebeek, line 61: As indicated on our call we're aware that the `PoolingHttpClientConnectionManager` changed here might have nothing to do with the `HttpComponentsClientHttpRequestFactory`  we're modifying. In theory the `ConnectionManager` could even be in a totally different file, not uncovered because this is not a ScanningRecipe. It'll be hard to cover all cases well, so this should be seen as an optimistic first step.

src/main/java/org/openrewrite/java/spring/framework/HttpComponentsClientHttpRequestFactoryReadTimeout.java
```diff
@@ -42,35 +50,38 @@ public String getDescription() {
...
                             public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx) {
                                 if (SET_READ_TIMEOUT_METHOD_MATCHER.matches(method)) {
                                     Expression expression = method.getArguments().get(0);
                                     if (expression instanceof J.Literal || expression instanceof J.FieldAccess) {
                                         readTimeout.set(expression);
                                     }
+
+                                    String message = " Manual migration to `SocketConfig.Builder.setSoTimeout(Timeout)` necessary; see: " +
+                                                     "https://docs.spring.io/spring-framework/docs/6.0.0/javadoc-api/org/springframework/http/client/HttpComponentsClientHttpRequestFactory.html#setReadTimeout(int)";
+                                    if (method.getComments().stream().noneMatch(c -> c.printComment(getCursor()).contains(message))){
```
Laurens-W, line 70: Shouldn't we be checking this when we match the `SET_DEFAULT_SOCKET_CONFIG`?

Laurens-W: Wait never mind, we're leaving the timeout in tact in most edge cases and instructing via this comment how to resolve

PR #579, opened 2024-08-20
rearranging springdoc/openapi recipes. individual springdoc upgrades …

nmck257:
…now occur with relevant spring boot upgrades. SB 2.6 now includes migration from springfox (which was popular for a while and then abandoned in ~2020) to springdoc. this bumps the swagger 2 -> openapi 3 migration to be a little sooner. the springdoc 2 upgrade with SB 3 now better-handles versions for renamed modules, and includes a swap to jakarta swagger artifacts (if apps are referencing those directly)

## What's your motivation?

springfox stops working for the most part on spring boot 2.6:
https://stackoverflow.com/questions/70178343/springfox-3-0-0-is-not-working-with-spring-boot-2-6-0
meanwhile springdoc is the de facto replacement.
even a partial migration from springfox to springdoc at 2.6 is probably helpful and leaves you no-worse-off.

## Any additional context

a future iteration could try to be more clever about whether to add the SpringDoc `webflux` starter instead of the `webmvc` starter; I started on it then set it aside when I realized I would want a precondition which looks for the *non-presence* of `spring-webmvc`, which I didn't see an obvious way to do.

a future iteration could also tackle the final code changes in the "migrating from springfox" guide: https://springdoc.org/#migrating-from-springfox

### Checklist

### DIFF
--- a/src/main/resources/META-INF/rewrite/spring-boot-22.yml
+++ b/src/main/resources/META-INF/rewrite/spring-boot-22.yml
@@ -53,6 +53,10 @@ recipeList:
   - org.openrewrite.gradle.UpdateGradleWrapper:
       version: ^4.10
       addIfMissing: false
+  - org.openrewrite.java.dependencies.UpgradeDependencyVersion:
+      groupId: org.springdoc
+      artifactId: "*"
+      newVersion: 1.3.x
 
   # Use recommended replacements for deprecated APIs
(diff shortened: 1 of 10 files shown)


timtebeek (approved):
Thanks a lot for rearranging these! Appreciate the effort taken to make incremental upgrades work as well.

Issue #531, opened 2024-05-22
Migrate `WebMvcTags` to `ServerRequestObservationConvention`


timtebeek:
## What problem are you trying to solve?
[WebMvcTags was deprecated in 3.0](https://docs.spring.io/spring-boot/docs/3.1.0/api/org/springframework/boot/actuate/metrics/web/servlet/WebMvcTags.html), for removal in 3.2 in favor of [ServerRequestObservationConvention](https://docs.spring.io/spring-framework/docs/6.0.9/javadoc-api/org/springframework/http/server/observation/ServerRequestObservationConvention.html).
The [3.0 release notes](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-3.0-Release-Notes) merely mention:
> RestTemplateExchangeTagsProvider, WebClientExchangeTagsProvider, WebFluxTagsProvider, WebMvcTagsProvider and related classes have been replaced with ObservationConvention equivalents.

The before/after of this change is still a bit unclear to me; hoping someone can provide additional samples of how this API was used, and how the replacements should look in 3.2+.

bsmahi:
Hi @timtebeek 

I could find some more information on this topic

https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-3.0-Migration-Guide#tag-providers-and-contributors-migration

Hope this helps. I will further explore on it.

Thanks,
Mahi

timtebeek:
Thanks for the helpful link @bsmahi ; looks quite involved at first glance, especially as the methods in [DefaultServerRequestObservationConvention](https://docs.spring.io/spring-framework/docs/6.0.9/javadoc-api/org/springframework/http/server/observation/DefaultServerRequestObservationConvention.html#method(org.springframework.http.server.observation.ServerRequestObservationContext)) are non static, whereas `WebMvcTags` did have static methods. That might complicate a migration, automated or manual.

timtebeek:
- Fixed in https://github.com/openrewrite/rewrite-spring/pull/586

Swachithreddy:
Hi all,I'm currently migrating to springboot 3.2.9 and found webmvtags was depreacted I'm using this methods.
webmvctags.method(request),
webmvctags.exception(exception) 
webmvctgas.uri(request ,response).
I ran the recipe but changes didn't get reflected.Can i k kw the replacement was on the same boat from few days.TIA

Laurens-W:
[...]

PR #586, opened 2024-09-04
Migrate `WebMvcTags` to `ServerRequestObservationConvention`

Laurens-W:
## What's your motivation?

- https://github.com/openrewrite/rewrite-spring/issues/531

@timtebeek 

## Any additional context

https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-3.0-Migration-Guide#tag-providers-and-contributors-migration

### Checklist


src/testWithSpringBoot_3_0/java/org/openrewrite/java/spring/boot3/MigrateWebMvcTagsToObservationConventionTest.java
```diff
@@ -0,0 +1,100 @@
...
+                public KeyValues getLowCardinalityKeyValues(ServerRequestObservationContext context) {
+                    HttpServletRequest request = context.getCarrier();
+                    KeyValues values = super.getLowCardinalityKeyValues(context);
+            
+                    String customHeader = request.getHeader("X-Custom-Header");
+                    if (customHeader != null) {
+                        values.and(KeyValue.of("custom.header", customHeader));
+                    }
+                    values.and(KeyValue.of("a", "b"), KeyValue.of("c", "d"));
+                    for (Tag tag : staticTags) {
```
Laurens-W, line 91: In the actual the preceding space for `staticTags` is gone, no idea why

src/main/java/org/openrewrite/java/spring/boot3/MigrateWebMvcTagsToObservationConvention.java
```diff
@@ -0,0 +1,242 @@
...
+                    return JavaTemplate.builder("values.and(" + string + ")")
+                            .javaParser(JavaParser.fromJavaVersion().classpath("micrometer-commons"))
+                            .build()
+                            .apply(getCursor(), a.getCoordinates().replace(), createKeys.toArray());
+                } else if (TAGS_AND_TAG_ARRAY.matches(init)) {
+                    J.Identifier iterable = (J.Identifier) init.getArguments().get(0);
+                    String template = "for (Tag tag : #{any()}) {\n" +
+                                      "    values.and(KeyValue.of(tag.getKey(), tag.getValue()));\n" +
+                                      "}\n";
+                    return JavaTemplate.builder(template)
```
Laurens-W, line 217: See template here

build.gradle.kts
```diff
@@ -114,6 +114,8 @@ recipeDependencies {
 
     parserClasspath("org.apache.httpcomponents.core5:httpcore5:5.1.+")
     parserClasspath("org.apache.httpcomponents.client5:httpclient5:5.1.+")
+
+    parserClasspath("org.apache.tomcat.embed:tomcat-embed-core:10.1.+")
```
timtebeek, line 118: Is there a smaller tomcat library we could add here? One with just the classes we need for type attribution, not everything needed to run tomcat?

Laurens-W: We only really need the jakarta servlet api, will update this

src/main/java/org/openrewrite/java/spring/boot3/MigrateWebMvcTagsToObservationConvention.java
```diff
@@ -0,0 +1,285 @@
...
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return Preconditions.check(new UsesType<>(WEBMVCTAGSPROVIDER_FQ, true), new JavaVisitor<ExecutionContext>() {
+            @Override
+            public J.ClassDeclaration visitClassDeclaration(J.ClassDeclaration classDecl, ExecutionContext ctx) {
+                List<Statement> getTagsBodyStatements = new ArrayList<>();
+                for (Statement stmt : classDecl.getBody().getStatements()) {
+                    if (stmt instanceof J.MethodDeclaration) {
+                        J.MethodDeclaration md = (J.MethodDeclaration) stmt;
+                        if (md.getSimpleName().equals("getTags") && md.getBody() != null) {
```
knutwannheden, line 76: Not sure if we should use a method matcher to be on the safe side. Moving the `md` variable to the top-level scope of this method would be good, so that we can use it later on.

src/main/java/org/openrewrite/java/spring/boot3/MigrateWebMvcTagsToObservationConvention.java
```diff
@@ -0,0 +1,285 @@
...
+                maybeAddImport(KEYVALUE_FQ);
+                maybeAddImport(KEYVALUES_FQ);
+                maybeAddImport(SERVERREQUESTOBSERVATIONCONVENTION_FQ);
+                maybeRemoveImport(HTTPSERVLETREQUEST_FQ);
+                maybeRemoveImport(HTTPSERVLETRESPONSE_FQ);
+                maybeRemoveImport(TAG_FQ);
+                maybeRemoveImport(TAGS_FQ);
+                maybeRemoveImport(WEBMVCTAGS_FQ);
+                maybeRemoveImport(WEBMVCTAGSPROVIDER_FQ);
+                return (J.ClassDeclaration) super.visitClassDeclaration(newClassDeclaration, ctx);
```
knutwannheden, line 127: With the super call being down here, I think we should update the cursor with `newClassDeclaration`, just so that there isn't any surprises if anyone in another visit method calls `getParent()` and expects to find the updated class declaration.

Alternatively we could maybe do the super call first?

src/main/java/org/openrewrite/java/spring/boot3/MigrateWebMvcTagsToObservationConvention.java
```diff
@@ -0,0 +1,285 @@
...
+                        .build()
+                        .apply(getCursor(), classDecl.getCoordinates().replace());
+
+                J.ClassDeclaration finalNewClassDeclaration = newClassDeclaration;
+                newClassDeclaration = newClassDeclaration
+                        .withLeadingAnnotations(classDecl.getLeadingAnnotations())
+                        .withModifiers(classDecl.getModifiers())
+                        .withPrefix(classDecl.getPrefix())
+                        .withBody(newClassDeclaration.getBody().withStatements(ListUtils.map(classDecl.getBody().getStatements(), stmt -> {
+                            if (stmt instanceof J.MethodDeclaration && ((J.MethodDeclaration) stmt).getSimpleName().equals("getTags")) {
```
knutwannheden, line 109: Here we could check if it is referentially equal to that `md` we previously found.

src/main/java/org/openrewrite/java/spring/boot3/MigrateWebMvcTagsToObservationConvention.java
```diff
@@ -0,0 +1,285 @@
...
+            public J.MethodDeclaration visitMethodDeclaration(J.MethodDeclaration method, ExecutionContext ctx) {
+                J.MethodDeclaration m = (J.MethodDeclaration) super.visitMethodDeclaration(method, ctx);
+                J.VariableDeclarations methodParam = (J.VariableDeclarations) m.getParameters().get(0);
+                J.Identifier methodParamIdentifier = methodParam.getVariables().get(0).getName();
+                Boolean addHttpServletResponse = getCursor().pollMessage("addHttpServletResponse");
+                Boolean addHttpServletRequest = getCursor().pollMessage("addHttpServletRequest");
+                if (Boolean.TRUE.equals(addHttpServletResponse) && m.getBody() != null) {
+                    m = JavaTemplate.builder("HttpServletResponse response = #{any()}.getResponse();")
+                            .imports(HTTPSERVLETRESPONSE_FQ, SERVERREQUESTOBSERVATIONCONVENTION_FQ)
+                            .contextSensitive()
```
knutwannheden, line 140: Does this template need to be context sensitive. It doesn't look like it, but this is really hard to tell...

build.gradle.kts
```diff
@@ -114,6 +114,10 @@ recipeDependencies {
 
     parserClasspath("org.apache.httpcomponents.core5:httpcore5:5.1.+")
     parserClasspath("org.apache.httpcomponents.client5:httpclient5:5.1.+")
+
+    parserClasspath("jakarta.servlet:jakarta.servlet-api:6.1.+")
+    parserClasspath("io.micrometer:micrometer-commons:1.11.+")
+    parserClasspath("io.micrometer:micrometer-core:1.11.12")
```
knutwannheden, line 120: Should we use `1.11.+` here as well?

knutwannheden (approved):

src/main/java/org/openrewrite/java/spring/boot3/MigrateWebMvcTagsToObservationConvention.java
```diff
@@ -0,0 +1,291 @@
...
+                    m = JavaTemplate.builder("HttpServletResponse response = #{any()}.getResponse();")
+                            .imports(HTTPSERVLETRESPONSE_FQ, SERVERREQUESTOBSERVATIONCONVENTION_FQ)
+                            .javaParser(JavaParser.fromJavaVersion().classpathFromResources(ctx, "jakarta.servlet-api", "spring-web-6.+", "micrometer-commons", "micrometer-observation"))
+                            .build()
+                            .apply(updateCursor(m), m.getBody().getCoordinates().firstStatement(), methodParamIdentifier);
+                }
+                if (Boolean.TRUE.equals(addHttpServletRequest) && m.getBody() != null) {
+                    m = JavaTemplate.builder("HttpServletRequest request = #{any()}.getCarrier();")
+                            .imports(HTTPSERVLETREQUEST_FQ, SERVERREQUESTOBSERVATIONCONVENTION_FQ, "io.micrometer.observation.transport.ReceiverContext")
+                            .javaParser(JavaParser.fromJavaVersion().classpathFromResources(ctx, "jakarta.servlet-api", "spring-web-6.+", "micrometer-commons", "micrometer-observation"))
```
timtebeek, line 155: Here you reference `micrometer-observation` but I don't see any matching jar in `/src/main/resources/META-INF/rewrite/classpath` or `build.gradle.kts` > `parserClasspath("...")`. Could you explain that part to me?

Laurens-W: Most likely it's transitively provided by spring-web, I've looked at these two templates again and was able to reduce the imports and required classpath, but micrometer-observations remains, would you like me to explicitly add it?

src/main/java/org/openrewrite/java/spring/boot3/MigrateWebMvcTagsToObservationConvention.java
```diff
@@ -0,0 +1,292 @@
...
+                            }
+
+                        }
+                    }
+                }
+
+                String tmpl = "class " + classDecl.getSimpleName() + " extends DefaultServerRequestObservationConvention {\n" +
+                              "    @Override\n" +
+                              "    public KeyValues getLowCardinalityKeyValues(ServerRequestObservationContext context) {\n" +
+                              "        KeyValues values = super.getLowCardinalityKeyValues(context);\n" +
```
timtebeek, line 98: Did you explore whether to use [low cardinality versus high cardinality](https://docs.micrometer.io/micrometer/reference/observation/introduction.html#:~:text=High%20cardinality%20means%20that%20a,bounded%20number%20of%20possible%20values.)? It seems more risky to assume low cardinality. Any thoughts on that subject?

timtebeek (changes requested):
Added a test that fails just from having another method without any arguments, as that leads to class cast exception. Likely you'll need some form of limiting when to modify methods.

timtebeek (approved):
Great to see; I understand this one wasn't easy with the lack of public examples and the number of options that needed to be converted. Thanks for diving in! Let's see how it holds up in practice with the coming release.

Issue #554, opened 2024-07-19
Spring Framework 6.0 - ResponseStatusException method changes


nmck257:
## What problem are you trying to solve?

In Spring Framework 5.3, `ResponseStatusException` had methods `getRawStatusCode()` and `getStatus()`:
https://docs.spring.io/spring-framework/docs/5.3.37/javadoc-api/org/springframework/web/server/ResponseStatusException.html

In Spring Framework 6.0, these are removed. The `getStatusCode()` method from superclass `ErrorResponseException` can be used as a direct replacement for `getStatus()`, and `getStatusCode().value()` can replace `getRawStatusCode()`:
https://docs.spring.io/spring-framework/docs/6.0.0/javadoc-api/org/springframework/web/server/ResponseStatusException.html

## What precondition(s) should be checked before applying this recipe?

None

## Describe the situation before applying the recipe

```java
class A {
    void foo(ResponseStatusException e) {
        int i = e.getRawStatusCode();
        HttpStatus s = e.getStatus();
    }
}
```

## Describe the situation after applying the recipe

```java
class A {
    void foo(ResponseStatusException e) {
        int i = e.getStatusCode().value();
        HttpStatus s = e.getStatusCode();
    }
}
```

## Are you interested in [contributing this recipe to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#adding-new-recipes)?

Not urgent for me; probably a good first issue?

timtebeek:
Thanks for logging the suggested improvement; I've had a quick look at the `getRawStatusCode` case using the new trait API. I'm running up against some test class path issues that stop me from adding this as is, but figured log this here as a starter for folks that want to help out.

```java
package org.openrewrite.java.spring.framework;

import org.openrewrite.ExecutionContext;
import org.openrewrite.Recipe;
import org.openrewrite.TreeVisitor;
import org.openrewrite.java.JavaParser;
import org.openrewrite.java.JavaTemplate;
import org.openrewrite.java.trait.Traits;
import org.openrewrite.java.tree.J;
import org.openrewrite.java.tree.MethodCall;

public class MigrateResponseStatusException extends Recipe {
    @Override
    public String getDisplayName() {
        return "Migrate `ResponseStatusException#getRawStatusCode()` to `getStatusCode().value()`";
    }

    @Override
    public String getDescription() {
[...]

PR #589, opened 2024-09-09
Update `HttpStatus` to `HttpStatusCode` in `ResponseEntityExceptionHandler` implementations

Laurens-W:
## What's your motivation?

- Fixes https://github.com/openrewrite/rewrite-spring/issues/369

## Anyone you would like to review specifically?

@timtebeek 

## Any additional context

- https://github.com/spring-projects/spring-framework/issues/28214
- https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/http/HttpStatus.html
- https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/http/HttpStatusCode.html

### Checklist
- [ ] I've added unit tests to cover both positive and negative cases
- [x] I've read and applied the [recipe conventions and best practices](https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices)
- [x] I've used the IntelliJ IDEA auto-formatter on affected files

PR #598, opened 2024-10-03
Add recipe to migrate `ResponseStatusException` changes

mslowiak:
## What's changed?
Add recipe to migrate breaking changes introduced in Spring Framework 6.0 touching `ResponseStatusException` class. 

## What's your motivation?
- https://github.com/openrewrite/rewrite-spring/issues/554

Was not sure where to put the recipes - I added them to `spring-framework-60.yml` as another group (? - let me know correct naming here)

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/spring/framework/MigrateResponseStatusExceptionGetRawStatusCodeMethod.java
+++ b/src/main/java/org/openrewrite/java/spring/framework/MigrateResponseStatusExceptionGetRawStatusCodeMethod.java
@@ -42,7 +42,6 @@ public class MigrateResponseStatusExceptionGetRawStatusCodeMethod extends Recipe
                     MethodCall tree = mc.getTree();
                     if (tree instanceof J.MethodInvocation) {
                         return JavaTemplate.builder("#{any()}.getStatusCode().value()")
-                                .contextSensitive()
                                 .javaParser(JavaParser.fromJavaVersion().classpathFromResources(ctx, "spring-core-6", "spring-beans-6", "spring-web-6"))
                                 .build().apply(mc.getCursor(), tree.getCoordinates().replace(), ((J.MethodInvocation) tree).getSelect());
                     }
--- a/src/main/java/org/openrewrite/java/spring/framework/MigrateResponseStatusExceptionGetRawStatusCodeMethod.java
+++ b/src/main/java/org/openrewrite/java/spring/framework/MigrateResponseStatusExceptionGetRawStatusCodeMethod.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2024 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package org.openrewrite.java.spring.framework;
 
 import org.openrewrite.ExecutionContext;
--- a/src/main/java/org/openrewrite/java/spring/framework/MigrateResponseStatusExceptionGetRawStatusCodeMethod.java
+++ b/src/main/java/org/openrewrite/java/spring/framework/MigrateResponseStatusExceptionGetRawStatusCodeMethod.java
@@ -1,0 +1,37 @@
+package org.openrewrite.java.spring.framework;
+
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.Recipe;
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.java.JavaParser;
+import org.openrewrite.java.JavaTemplate;
+import org.openrewrite.java.trait.Traits;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.java.tree.MethodCall;
+
+public class MigrateResponseStatusExceptionGetRawStatusCodeMethod extends Recipe {
+    @Override
+    public String getDisplayName() {
+        return "Migrate `ResponseStatusException#getRawStatusCode()` to `getStatusCode().value()`";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Migrate Spring Framework 5.3's `ResponseStatusException` method `getRawStatusCode()` to Spring Framework 6's `getStatusCode().value()`.";
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return Traits.methodAccess("org.springframework.web.server.ResponseStatusException getRawStatusCode()")
+                .asVisitor((mc, ctx) -> {
+                    MethodCall tree = mc.getTree();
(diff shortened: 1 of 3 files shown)


timtebeek:
@Laurens-W any input from your side here seeing that you recently added this similar
- https://github.com/openrewrite/rewrite-spring/pull/589 ?

src/main/java/org/openrewrite/java/spring/framework/MigrateResponseStatusExceptionGetRawStatusCodeMethod.java
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    @Override
+    public String getDescription() {
+        return "Migrate Spring Framework 5.3's `ResponseStatusException` method `getRawStatusCode()` to Spring Framework 6's `getStatusCode().value()`.";
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return Traits.methodAccess("org.springframework.web.server.ResponseStatusException getRawStatusCode()")
```
timtebeek, line 25: I love that you've adopted the Traits API here; seems like a perfect match for the replacement here.

src/main/java/org/openrewrite/java/spring/framework/MigrateResponseStatusExceptionGetRawStatusCodeMethod.java
```diff
@@ -0,0 +1,37 @@
...
+                .asVisitor((mc, ctx) -> {
+                    MethodCall tree = mc.getTree();
+                    if (tree instanceof J.MethodInvocation) {
+                        return JavaTemplate.builder("#{any()}.getStatusCode().value()")
+                                .contextSensitive()
+                                .javaParser(JavaParser.fromJavaVersion().classpathFromResources(ctx, "spring-core-6", "spring-beans-6", "spring-web-6"))
+                                .build().apply(mc.getCursor(), tree.getCoordinates().replace(), ((J.MethodInvocation) tree).getSelect());
+                    }
+                    return tree;
```
timtebeek, line 34: Would it make sense to cover method reference here as well? Not expecting a ton of those, but it seems that would be easy enough to add just in case.

src/main/resources/META-INF/rewrite/spring-framework-60.yml
```diff
@@ -31,7 +31,16 @@ recipeList:
   - org.openrewrite.apache.httpclient5.UpgradeApacheHttpClient_5
   - org.openrewrite.java.spring.framework.HttpComponentsClientHttpRequestFactoryReadTimeout
   - org.openrewrite.java.spring.framework.MigrateResponseEntityExceptionHandlerHttpStatusToHttpStatusCode
-
+---
+type: specs.openrewrite.org/v1beta/recipe
+name: org.openrewrite.java.spring.framework.MigrateResponseStatusException
```
timtebeek, line 36: Indeed best to reference this `MigrateResponseStatusException` from `UpgradeSpringFramework_6_0` still, otherwise folks would need to run it explicitly before seeing those changes.

timtebeek (commented):
This is about as perfect of a pull request as I have ever seen; thanks a lot @mslowiak ! Just a small final question/suggestion before we merge, which we can also pick up on our end if needed.

src/main/java/org/openrewrite/java/spring/framework/MigrateResponseStatusExceptionGetRawStatusCodeMethod.java
timtebeek: Hmm; might be more involved as it likely needs `org.openrewrite.java.spring.util.concurrent.MemberReferenceToMethodInvocation` as well; not worth it for now. Lets postpone that additional work until we get reports that it's actually needed.

src/test/java/org/openrewrite/java/spring/framework/MigrateResponseStatusExceptionTest.java
```diff
@@ -0,0 +1,92 @@
...
+                      HttpStatus i = e.getStatus();
+                  }
+              }
+              """,
+            """
+              import org.springframework.http.HttpStatus;
+              import org.springframework.web.server.ResponseStatusException;
+              class A {
+                  void foo(ResponseStatusException e) {
+                      HttpStatus i = e.getStatusCode();
```
timtebeek, line 57: I think this ought to be `HttpStatusCode` instead of `HttpStatus`, as per:
https://docs.spring.io/spring-framework/docs/6.0.0/javadoc-api/org/springframework/web/ErrorResponseException.html#getStatusCode()

timtebeek: @Laurens-W is that something you can work in based on your earlier work on #589 ?

src/main/java/org/openrewrite/java/spring/framework/MigrateResponseStatusExceptionGetRawStatusCodeMethod.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return Traits.methodAccess("org.springframework.web.server.ResponseStatusException getRawStatusCode()")
+                .asVisitor((mc, ctx) -> {
+                    MethodCall tree = mc.getTree();
+                    if (tree instanceof J.MethodInvocation) {
+                        return JavaTemplate.builder("#{any()}.getStatusCode().value()")
+                                .contextSensitive()
```
knutwannheden, line 45: I am surprised that this template needs to be context-sensitive. What error is there if we remove this? Note that this can make a huge performance difference, as the result of a context-free template can be cached.

timtebeek: Good point; indeed not necessary here and removed.

src/test/java/org/openrewrite/java/spring/framework/MigrateResponseStatusExceptionTest.java
timtebeek: Merging this iteration already as it's an improvement already; we'd need to change the test in a follow up PR to expect the variable and import change, and create a new dedicated recipe to handle the `ResponseStatusException getStatus()`.

src/main/resources/META-INF/rewrite/spring-framework-60.yml
```diff
@@ -31,7 +31,17 @@ recipeList:
...
   - org.openrewrite.java.spring.framework.MigrateResponseEntityExceptionHandlerHttpStatusToHttpStatusCode
-
+  - org.openrewrite.java.spring.framework.MigrateResponseStatusException
+---
+type: specs.openrewrite.org/v1beta/recipe
+name: org.openrewrite.java.spring.framework.MigrateResponseStatusException
+displayName: Migrate breaking changes in `ResponseStatusException`
+description: Migrate Spring Framework 5.3's `ResponseStatusException` method `getRawStatusCode()` to Spring Framework 6's `getStatusCode().value()` and `ResponseStatusException` method `getStatus()` to Spring Framework 6's `getStatusCode()` .
+recipeList:
+  - org.openrewrite.java.spring.framework.MigrateResponseStatusExceptionGetRawStatusCodeMethod
```
Laurens-W, line 41: Perhaps once the recipe for `getStatus` to `getStatusCode` is in place these could be included below line 33, or the recipe on line 33 could be pulled into here and this sub recipe could be transformed into a `HttpStatus` to `HttpStatusCode` recipe

PR #603, opened 2024-10-14
Include Apache commons-lang3 migration with Spring Boot 2.0 ugprade

Laurens-W:

@timtebeek 

## Any additional context

Spring manages the version since boot2: https://github.com/spring-projects/spring-boot/blob/2.0.x/spring-boot-project/spring-boot-dependencies/pom.xml#L50
And not in 1.5: https://github.com/spring-projects/spring-boot/blob/1.5.x/spring-boot-dependencies/pom.xml

### Checklist


timtebeek (approved):
Thanks! Good to see this included; about time folks moved over.

Issue #475, opened 2024-01-12
`StringUtils#isEmpty` is not equivalent to `ObjectUtils#isEmpty` for collections


timtebeek:
The replacement done on these lines is incorrect when it concerns empty collections, is what I'm told on Discord:
https://github.com/openrewrite/rewrite-spring/blob/71fbf7c27536877c47b28e85a13f0f4629d6b5de/src/main/resources/META-INF/rewrite/spring-framework-53.yml#L47-L54

> earlier a List<Object> list was passed, which was coming as null from a rest API call. Then the REST call started returning an emptyList() instead which was not null. But the stricter check in ObjectUtils now checks if its any collection and the collection is empty or not.
Its an idiosyncracy from our codebase, but the expectation from rewrite plugin was the code changes would be one-to-one, just a null check like the last impl of StringUtils would have been fine. Making it more stricter broke the expected flow
Ideally if you are not passing any sort of Collection to StringUtils.isEmpty() you are good to replace

PR #605, opened 2024-10-16
Replace conversion to `ObjectUtils.isEmpty` with `!StringUtils.hasLength`

dimijdriven:
## What's changed?

`StringUtils.isEmpty()` is deprecated currently openrewrite replaces it with `ObjectUtils.isEmpty()` which is stricter
and led to this issue:
- https://github.com/openrewrite/rewrite-spring/issues/475

According to the official suggestions: https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/util/StringUtils.html#isEmpty(java.lang.Object)

I replaced it with `!StringUtils.hasLength`
## What's your motivation?

- Fixes https://github.com/openrewrite/rewrite-spring/issues/475

@Laurens-W 

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/spring/StringOptimization.java
+++ b/src/main/java/org/openrewrite/java/spring/StringOptimization.java
@@ -1,0 +1,27 @@
+package org.openrewrite.java.spring;
+
+import com.google.errorprone.refaster.annotation.AfterTemplate;
+import com.google.errorprone.refaster.annotation.BeforeTemplate;
+import org.openrewrite.java.template.RecipeDescriptor;
+import org.springframework.util.StringUtils;
+
+@SuppressWarnings("ALL")
+public class StringOptimization {
+
+    @RecipeDescriptor(
+            name = "Use `!StringUtils.hasLength`",
+            description = "Replace usage of deprecated `StringUtils.isEmpty` with `!StringUtils.hasLength` https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/util/StringUtils.html#isEmpty(java.lang.Object).")
+    static class ReplaceDeprecatedStringUtilsIsEmpty {
+
+
+        @BeforeTemplate
+        boolean before(String s) {
+            return StringUtils.isEmpty(s);
+        }
+
+        @AfterTemplate
+        boolean after(String s) {
+            return !StringUtils.hasLength(s);
+        }
+    }
+}
--- a/src/test/java/org/openrewrite/java/spring/ReplaceDeprecatedStringUtilsIsEmptyTest.java
+++ b/src/test/java/org/openrewrite/java/spring/ReplaceDeprecatedStringUtilsIsEmptyTest.java
@@ -1,0 +1,45 @@
+package org.openrewrite.java.spring;
+
+import org.junit.jupiter.api.Test;
+import org.openrewrite.DocumentExample;
+import org.openrewrite.InMemoryExecutionContext;
+import org.openrewrite.java.JavaParser;
+import org.openrewrite.test.RecipeSpec;
+import org.openrewrite.test.RewriteTest;
+
+import static org.openrewrite.java.Assertions.java;
+
+class ReplaceDeprecatedStringUtilsIsEmptyTest implements RewriteTest {
+
+    @Override
+    public void defaults(RecipeSpec spec){
+        spec.recipe(new StringOptimizationRecipes()).parser(JavaParser.fromJavaVersion().classpathFromResources(new InMemoryExecutionContext(),"spring-core-5.+"));
+    }
+
+    @DocumentExample
+    @Test
+    @SuppressWarnings("deprecation")
+    void replaceStringUtilsIsEmpty() {
+        rewriteRun(
+          //language=java
+          java(
+            """
+              import org.springframework.util.StringUtils;
+              class Test {
+                    void test(String s) {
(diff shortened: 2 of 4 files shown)


src/main/java/org/openrewrite/java/spring/StringOptimization.java
```diff
@@ -0,0 +1,27 @@
+package org.openrewrite.java.spring;
+
+import com.google.errorprone.refaster.annotation.AfterTemplate;
+import com.google.errorprone.refaster.annotation.BeforeTemplate;
+import org.openrewrite.java.template.RecipeDescriptor;
+import org.springframework.util.StringUtils;
+
+@SuppressWarnings("ALL")
```
Laurens-W, line 8: This might be a bit broad

src/test/java/org/openrewrite/java/spring/ReplaceDeprecatedStringUtilsIsEmptyTest.java
```diff
@@ -0,0 +1,45 @@
...
+    @SuppressWarnings("deprecation")
+    void replaceStringUtilsIsEmpty() {
+        rewriteRun(
+          //language=java
+          java(
+            """
+              import org.springframework.util.StringUtils;
+              class Test {
+                    void test(String s) {
+                      return StringUtils.isEmpty(s);
```
Laurens-W, line 30: Let's also add a testcase where `!StringUtils.isEmpty(s);` is used

src/main/java/org/openrewrite/java/spring/StringOptimization.java
timtebeek: Also be sure to add a `@RecipeDescriptor` to the outer class, as that'll get turned into a `Recipes` class itself, which needs documentation.

dimijdriven:
> Just passing by, but the new implementation will only deal with `String` compile-time arguments, right?
> 
> I think it makes sense to be conservative indeed, and leave the deprecated call if it can’t be guaranteed to keep the original intention.
> 
> However, the case of `String` compile-time argument was actually the only case where using `ObjectUtils.isEmpty()` would have preserved the semantics, so maybe it’s better than `!StringUtils.hasLength()`? I think the former remains more readable.
> 
> Since `ObjectUtils.isEmpty()` can’t be used, I wonder if there is any choice that can’t be considered as too opinionated though.

I think you are right, also I realized after your comment that probably my PR does not solve the issue, on the contrary it will cause compile errors for this specific user.

But it is a weird case all around, an object was accepted for a String comparison, probably that is why they deprecated this method and the user was putting as input for an object a list!

I replicated and a minimum example would be the following:
ObjectUtils.isEmpty(Collections.EMPTY_LIST); // returns true
StringUtils.isEmpty(Collections.EMPTY_LIST); // returns false

So not sure if it actually makes sense to fix that case.

dimijdriven:
Closed after the comment of @DidierLoiseau

DidierLoiseau:
> But it is a weird case all around, an object was accepted for a String comparison, probably that is why they deprecated this method and the user was putting as input for an object a list!

Yes, I would assume so too. If you check the `StringUtils.isEmpty()` implementation it is `str == null || "".equals(str)`. I suppose the goal of having an `Object` parameter was to avoid checking if some value was actually a `String` upfront.

> I think you are right, also I realized after your comment that probably my PR does not solve the issue, on the contrary it will cause compile errors for this specific user.

Do you think so? As this PR currently stands, it would only do a change when the parameter is actually a `String`, right? So what could be the issue?

Maybe this PR should just remove the migration entirely instead, since it is unsafe?

dimijdriven:
> > But it is a weird case all around, an object was accepted for a String comparison, probably that is why they deprecated this method and the user was putting as input for an object a list!
> 
> Yes, I would assume so too. If you check the `StringUtils.isEmpty()` implementation it is `str == null || "".equals(str)`. I suppose the goal of having an `Object` parameter was to avoid checking if some value was actually a `String` upfront.
> 
> > I think you are right, also I realized after your comment that probably my PR does not solve the issue, on the contrary it will cause compile errors for this specific user.
> 
> Do you think so? As this PR currently stands, it would only do a change when the parameter is actually a `String`, right? So what could be the issue?
The person who opened the ticket, passes a list to the function and that works because it accepts objects. The current migration accepts also objects so it compiles, just does not return the correct result. If it is refactored to what I did it will only accept String as parameters and I expect that this will cause a compilation error because List is not String.

> Maybe this PR should just remove the migration entirely instead, since it is unsafe?
This would solve the issue, but not the deprecation. I would not be sure if the ideal solution is to remove the migration or just accept that the user was an extreme edge case and accept that in this case the migration breaks.

Laurens-W:
> > The person who opened the ticket, passes a list to the function and that works because it accepts objects. The current migration accepts also objects so it compiles, just does not return the correct result. If it is refactored to what I did it will only accept String as parameters and I expect that this will cause a compilation error because List is not String.
> 
> No but with this PR, it would only do the change when the compile-time type of the expression used as argument for `StringUtils.isEmpty()` is a `String`, if my understanding of Refaster is correct.
> 
> > This would solve the issue, but not the deprecation. I would not be sure if the ideal solution is to remove the migration or just accept that the user was an extreme edge case and accept that in this case the migration breaks.
> 
> I don’t think OpenRewrite recipes should be expected to fix all deprecation issues.

This is correct, and even if refaster doesn't cut it you could make the recipe "strongly typed" so to say

PR #607, opened 2024-10-16
adding a test case for MigrateWebMvcTagsToObservationConvention for t…

nmck257:
…he case where the iterable passed to `Tags.and(...)` is *not* an identifier, and partially fixing it, BUT this commit still has a strange test failure with cyclically duplicating http request/response variables

## What's changed?

see subject

## What's your motivation?

currently, the scenario in the test case yields a classcast exception and breaks the recipe process

here's the result of running this test case.

it seems that these lines are getting added again in a second cycle?
```java
        HttpServletRequest request = context.getCarrier();
        HttpServletResponse response = context.getResponse();
```

test result:
```
org.opentest4j.AssertionFailedError: [Expected recipe to complete in 1 cycle, but took at least one more cycle. Between the last two executed cycles there were changes to "CustomWebMvcTagsProvider.java"] 
expected: "import io.micrometer.common.KeyValue;
import io.micrometer.common.KeyValues;
import io.micrometer.core.instrument.Tag;
import io.micrometer.core.instrument.Tags;
import jakarta.servlet.http.HttpServletRequest;
import jakarta.servlet.http.HttpServletResponse;
import org.springframework.http.server.observation.DefaultServerRequestObservationConvention;
import org.springframework.http.server.observation.ServerRequestObservationContext;

class CustomWebMvcTagsProvider extends DefaultServerRequestObservationConvention {

    Tags staticTags = Tags.of("a", "b", "c", "d");
    Tag staticTag = Tag.of("a", "b");

    @Override
    public KeyValues getHighCardinalityKeyValues(ServerRequestObservationContext context) {
        HttpServletRequest request = context.getCarrier();
        HttpServletResponse response = context.getResponse();
        KeyValues values = super.getHighCardinalityKeyValues(context);

        String customHeader = request.getHeader("X-Custom-Header");
        if (customHeader != null) {
            values.and(KeyValue.of("custom.header", customHeader));
        }
        if (response.getStatus() >= 400) {
            values.and(KeyValue.of("error", "true"));
        }
        values.and(KeyValue.of("a", "b"), KeyValue.of("c", "d"));
        values.and(KeyValue.of("a", "b"), KeyValue.of(staticTag.getKey(), staticTag.getValue()));
        for (Tag tag : staticTags) {
            values.and(KeyValue.of(tag.getKey(), tag.getValue()));
        }
        for (Tag tag : methodTags()) {
            values.and(KeyValue.of(tag.getKey(), tag.getValue()));
        }
        return values;
    }

    private Tags methodTags() {
        return staticTags;
    }
}"

 but was:

"import io.micrometer.common.KeyValue;
import io.micrometer.common.KeyValues;
import io.micrometer.core.instrument.Tag;
import io.micrometer.core.instrument.Tags;
import jakarta.servlet.http.HttpServletRequest;
import jakarta.servlet.http.HttpServletResponse;
import org.springframework.http.server.observation.DefaultServerRequestObservationConvention;
import org.springframework.http.server.observation.ServerRequestObservationContext;

class CustomWebMvcTagsProvider extends DefaultServerRequestObservationConvention {

    Tags staticTags = Tags.of("a", "b", "c", "d");
    Tag staticTag = Tag.of("a", "b");

    @Override
    public KeyValues getHighCardinalityKeyValues(ServerRequestObservationContext context) {
        HttpServletRequest request = context.getCarrier();
        HttpServletResponse response = context.getResponse();
        HttpServletRequest request = context.getCarrier();
        HttpServletResponse response = context.getResponse();
        KeyValues values = super.getHighCardinalityKeyValues(context);

        String customHeader = request.getHeader("X-Custom-Header");
        if (customHeader != null) {
            values.and(KeyValue.of("custom.header", customHeader));
        }
        if (response.getStatus() >= 400) {
            values.and(KeyValue.of("error", "true"));
        }
        values.and(KeyValue.of("a", "b"), KeyValue.of("c", "d"));
        values.and(KeyValue.of("a", "b"), KeyValue.of(staticTag.getKey(), staticTag.getValue()));
        for (Tag tag : staticTags) {
            values.and(KeyValue.of(tag.getKey(), tag.getValue()));
        }
        for (Tag tag : methodTags()) {
            values.and(KeyValue.of(tag.getKey(), tag.getValue()));
        }
        return values;
    }

    private Tags methodTags() {
        return staticTags;
    }
}"
```

@Laurens-W 
@timtebeek 

## Any additional context

looks like there's some cursor messaging which tries to control against duplicating these lines; not sure why it's breaking. probably simpler for the original developer to look since it was only a week ago :)

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/spring/boot3/MigrateWebMvcTagsToObservationConvention.java
+++ b/src/main/java/org/openrewrite/java/spring/boot3/MigrateWebMvcTagsToObservationConvention.java
@@ -24,7 +24,7 @@ import org.openrewrite.java.JavaParser;
 import org.openrewrite.java.JavaTemplate;
 import org.openrewrite.java.JavaVisitor;
 import org.openrewrite.java.MethodMatcher;
-import org.openrewrite.java.search.UsesType;
+import org.openrewrite.java.search.FindImplementations;
 import org.openrewrite.java.tree.*;
 
 import java.util.ArrayList;
@@ -68,7 +68,7 @@ public class MigrateWebMvcTagsToObservationConvention extends Recipe {
 
     @Override
     public TreeVisitor<?, ExecutionContext> getVisitor() {
-        return Preconditions.check(new UsesType<>(WEBMVCTAGSPROVIDER_FQ, true), new JavaVisitor<ExecutionContext>() {
+        return Preconditions.check(new FindImplementations(WEBMVCTAGSPROVIDER_FQ), new JavaVisitor<ExecutionContext>() {
             @Override
             public J.ClassDeclaration visitClassDeclaration(J.ClassDeclaration classDecl, ExecutionContext ctx) {
                 J.MethodDeclaration getTagsMethod = null;
--- a/src/main/java/org/openrewrite/java/spring/boot3/MigrateWebMvcTagsToObservationConvention.java
+++ b/src/main/java/org/openrewrite/java/spring/boot3/MigrateWebMvcTagsToObservationConvention.java
@@ -251,7 +251,7 @@ public class MigrateWebMvcTagsToObservationConvention extends Recipe {
                             }
                             return getMultiKeyValueStatement(ctx, coords, args, returnIdentifier);
                         } else if (TAGS_AND_TAG_ITERABLE.matches(init) || TAGS_OF_TAG_ITERABLE.matches(init)) {
-                            J.Identifier iterable = (J.Identifier) init.getArguments().get(0);
+                            Expression iterable = init.getArguments().get(0);
                             String template = "for (Tag tag : #{any()}) {\n" +
                                               "    #{any()}.and(KeyValue.of(tag.getKey(), tag.getValue()));\n" +
                                               "}\n";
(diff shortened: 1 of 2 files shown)


Laurens-W:
It's making a second cycle because the precondition still matches. I think changing the precondition to `org.openrewrite.java.search.FindImplementations` would solve it

nmck257:
yep, passing tests with that change. thanks!
...though, does that imply there's a (known) gap in this recipe where it should ideally update `DeclaringType` of all the class' declared methods after refactoring to a different superclass?

Laurens-W:
> yep, passing tests with that change. thanks! ...though, does that imply there's a (known) gap in this recipe where it should ideally update `DeclaringType` of all the class' declared methods after refactoring to a different superclass?

If we want the LST to be 100% correct maybe we should.
What's odd to me is that it isn't preventing the import from being removed.
I'll take a look at the (unrelated) failing test as I've probably introduced that one

nmck257:
Import removal probably does not consult implicit usages, which feels correct, since we only need imports for types explicitly named in the source code

timtebeek:
Thanks for the fixes here both! Didn't get to a review with travel, but certainly agree with the change seeing it now. :)

PR #612, opened 2024-10-28
Replace @Required on setters with @Autowired

nielsdebruin:
## What's changed?

The deprecated [@Required](https://docs.spring.io/spring-framework/docs/5.1.8.RELEASE_to_5.1.9.RELEASE/Spring%20Framework%205.1.8.RELEASE/org/springframework/beans/factory/annotation/Required.html) annotation can be applied to setters of field, to ensure a bean is injected. However, the annotation is now deprecated. This recipe will replace the `@Required` annotation with `@Autowired`.

@timtebeek  @Laurens-W 

### Checklist


src/main/java/org/openrewrite/java/spring/ReplaceRequiredAnnotationOnSetterWithAutowired.java
```diff
@@ -0,0 +1,70 @@
...
+import org.openrewrite.Recipe;
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.java.*;
+import org.openrewrite.java.search.FindAnnotations;
+import org.openrewrite.java.search.UsesType;
+import org.openrewrite.java.tree.J;
+
+import java.util.Comparator;
+
+public class ReplaceRequiredAnnotationOnSetterWithAutowired extends Recipe {
```
timtebeek, line 29: Instead of a recipe would this also be possible with a declarative recipe `ChangeType`? Or are there edge cases that won't cover?

Whenever possible try to do declaratively what we can; that helps reduce what we need to maintain going forward.

timtebeek (commented):
Also: your recipe is not yet hooked into any of the Spring Boot migrations, meaning folks would have to run this one manually; please hook it into the appropriate place in a larger migration.

timtebeek (approved):

src/main/java/org/openrewrite/java/spring/ReplaceRequiredAnnotationOnSetterWithAutowired.java
nielsdebruin: Good suggestion, Thanks!

Laurens-W (approved):

PR #614, opened 2024-10-30
Migrate AuditorAware to an Optional<> for spring boot 1 to 2 migration

### LINKED ISSUE #613: spring-data-commons changed AuditorAware from 1.x to 2.x
Migrating between Spring boot 1 and 2, the implementation of AuditorAware changed. 

see [the code](https://github.com/spring-projects/spring-data-commons/blob/1.13.23.RELEASE/src/main/java/org/springframework/data/domain/AuditorAware.java#L31)
```java
public class AuditorAwareImpl implements AuditorAware<String> {
    public String getCurrentAuditor() {
        return "Some code here";
    }
}
```

see [the code](https://github.com/spring-projects/spring-data-commons/blob/2.0.0.RELEASE/src/main/java/org/springframework/data/domain/AuditorAware.java#L33)
```java
public class AuditorAwareImpl implements AuditorAware<String> {
    public Optional<String> getCurrentAuditor() {
        return Optional.ofNullable("Some code here");
    }
}
```

This should be an easy one. Will see, if no one picks it up sooner.

Jenson3210:
- This closes #613


Jenson3210:
@timtebeek I've added 4 `//TODO Question for TIM:`'s. 
If you could have a look and help me out in understanding the issue? 🙏 

Thanks!

Jenson3210:
@timtebeek Can we also go over the changes of the bot once? 
I am seeing stuff I would not perse just commit as suggestion.

For example: where did the maybeAddImport go? 
Screenshot 2024-10-30 at 22 09 42

timtebeek:
> @timtebeek Can we also go over the changes of the bot once? I am seeing stuff I would not perse just commit as suggestion.
> 
> For example: where did the maybeAddImport go?

Hi yes unfortunately [the .patch file is correct](https://github.com/openrewrite/rewrite-spring/actions/runs/11601658865/job/32304917675?pr=614#step:12:1), but once converted into comments by https://github.com/googleapis/code-suggester it's sometimes a little off when crossing multiple lines. I have yet to fix that external tool, or write my own. 

You can also run the recipes directly if you prefer, possibly through the Moderne CLI to make it most convenient:
https://docs.openrewrite.org/recipes/recipes/openrewritebestpractices

Jenson3210:
Okay. Makes sense. 

On my private pc I do not have moderne cli set up yet as I do not have the access token etc. 

Would that be possible for the open source projects (like open rewrite itself)?

timtebeek:
> Okay. Makes sense.
> 
> On my private pc I do not have moderne cli set up yet as I do not have the access token etc.
> 
> Would that be possible for the open source projects (like open rewrite itself)?

Yes for OSS you would not even need a Moderne license; Just a Moderne public instance personal access token to download recipes and LSTs.

src/test/java/org/openrewrite/java/spring/data/MigrateAuditorAwareToOptionalTest.java
```diff
@@ -0,0 +1,362 @@
...
+          java(
+            """
+              package sample;
+              
+              import org.springframework.data.domain.AuditorAware;
+              
+              public class MyAuditorAware implements AuditorAware<String> {
+                  @Override
+                  public String getCurrentAuditor() {
+                      return "admin";
```
timtebeek, line 54: Most of the tests seem to use String literals; would it make sense have a unit test a little more reflective of expected use, like method invocations [seen here](https://github.com/timtebeek/spring-security-samples/blob/2e6d81211abb8ddae5589e41c2bf5d477143b2d7/audit-spring-data-entities/src/main/java/com/jdriven/AuditSecurityConfiguration.java#L20-L27)?

src/test/java/org/openrewrite/java/spring/data/MigrateAuditorAwareToOptionalTest.java
```diff
@@ -0,0 +1,362 @@
...
+          //language=java
+          java(
+            """
+              package sample;
+              
+              import org.springframework.data.domain.AuditorAware;
+              
+              public class Configuration {
+                  public AuditorAware<String> auditorAware() {
+                      return this::getCurrentAuditor;
```
timtebeek, line 212: Optionally there's this utility that for now was tucked away; might make this easier if we leverage that util.
https://github.com/openrewrite/rewrite-spring/blob/298ba3aeffc5cf1035a0e27c85caf5ed2f3ec0b5/src/main/java/org/openrewrite/java/spring/util/concurrent/MemberReferenceToMethodInvocation.java#L31

src/main/java/org/openrewrite/java/spring/data/MigrateAuditorAwareToOptional.java
```diff
@@ -0,0 +1,180 @@
...
+                    return return_.withExpression(implementationVisitor.visitNewClass((J.NewClass) expression, ctx));
+                } else if (expression instanceof J.MemberReference) {
+                    J.MemberReference memberReference = (J.MemberReference) expression;
+                    JavaType.Method methodType = memberReference.getMethodType();
+                    if (methodType == null || isOptional.matches(methodType.getReturnType())) {
+                        return return_;
+                    }
+                    Expression containing = memberReference.getContaining();
+                    //TODO Question for TIM: If I use #{any()} for the method name, as getName returns a String, I get a java.lang.ClassCastException: class java.lang.String cannot be cast to class org.openrewrite.java.tree.J
+                    JavaTemplate template = JavaTemplate.builder("() -> Optional.ofNullable(#{any()}." + methodType.getName() + "())")
```
timtebeek, line 170: As indicated; if we reuse MemberReferenceToMethodInvocation here I think this will be easier to tackle.

src/test/java/org/openrewrite/java/spring/data/MigrateAuditorAwareToOptionalTest.java
```diff
@@ -0,0 +1,362 @@
...
+          .parser(JavaParser.fromJavaVersion().classpath("spring-data-commons"))
+          .recipe(new MigrateAuditorAwareToOptional());
+    }
+
+    @DocumentExample
+    @Test
+    void rewriteImplementation() {
+        //TODO Question for TIM: how to get rid of the types? I have the imports.
+        //-    public Optional<String> getCurrentAuditor() {
+        //+    public java.util.Optional<java.lang.String> getCurrentAuditor() {
```
timtebeek, line 42: I've added `ShortenFullyQualifiedTypeReferences` to shorten the Optional import. That intentionally does not shorten any java.lang references, as it possible those were intentional to avoid conflicts with for instance `java.lang.Record`. The `String` would thus have to be shortened on the generated template here, or kept as is.
```suggestion
```

src/test/java/org/openrewrite/java/spring/data/MigrateAuditorAwareToOptionalTest.java
```diff
@@ -0,0 +1,362 @@
...
+    void rewriteLambdaLiteral() {
+        //TODO Question for TIM:
+        //sample/Configuration.java:7: error: incompatible types: bad return type in lambda expression
+        //       return () -> "admin";
+        //                    ^
+        //   String cannot be converted to Optional<String>
+        //
+        //LST contains missing or invalid type information
+        //MethodInvocation->Lambda->Return->Block->MethodDeclaration->Block->ClassDeclaration->CompilationUnit
+        // *~~(MethodInvocation type is missing or malformed)~~>*/Optional.ofNullable("admin")
```
timtebeek, line 85: No quick fix here just yet; it reads like we might need a classpath from resources for the `AuditorAware` method signature with a different return type, and use that in the JavaTemplates instead of the older version with a non-Optional return type. Not entirely sure if that would fix it here, but it might be what contributing to the invalid types.

src/main/java/org/openrewrite/java/spring/data/MigrateAuditorAwareToOptional.java
```diff
@@ -0,0 +1,180 @@
...
+
+    private JavaIsoVisitor<ExecutionContext> implementationVisitor() {
+        return new JavaIsoVisitor<ExecutionContext>() {
+
+            @Override
+            public J.CompilationUnit visitCompilationUnit(J.CompilationUnit cu, ExecutionContext ctx) {
+                maybeAddImport("java.util.Optional");
+                return super.visitCompilationUnit(cu, ctx);
+            }
+
```
timtebeek, line 70: Let's not unconditionally add a visitor that might add the import, and instead move this method invocation further down where we actually might introduce the need for the import.
```suggestion
```

src/main/java/org/openrewrite/java/spring/data/MigrateAuditorAwareToOptional.java
```diff
@@ -0,0 +1,180 @@
...
+        return "As of Spring boot 2.0, the `AuditorAware.getCurrentAuditor` method should return an `Optional`. " +
+               "This recipe will update the implementations of this method to return an `Optional` using the `ofNullable`.";
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        JavaIsoVisitor<ExecutionContext> implementationVisitor = implementationVisitor();
+        JavaIsoVisitor<ExecutionContext> functionalVisitor = functionalVisitor(implementationVisitor);
+
+        //TODO the other visitors for new AuditorAware() {...} and method references.
```
timtebeek, line 45: What additional cases would you like to cover? Or is this left over from earlier?
```suggestion
```

src/main/java/org/openrewrite/java/spring/data/MigrateAuditorAwareToOptional.java
```diff
@@ -0,0 +1,180 @@
...
+
+        return new TreeVisitor<Tree, ExecutionContext>() {
```
timtebeek, line 47: Perhaps good to add a Precondition here on UsesType for that AuditorAware. That would help to quickly reject unrelated classes.

timtebeek (commented):
Great to see how much of this you've covered already; I've applied some fixes already and left a few comments for other work left to do. Didn't quite finish the last failing test, but we can revisit that once the other items are crossed off.

src/main/java/org/openrewrite/java/spring/data/MigrateAuditorAwareToOptional.java
```diff
@@ -0,0 +1,180 @@
...
+               "This recipe will update the implementations of this method to return an `Optional` using the `ofNullable`.";
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        JavaIsoVisitor<ExecutionContext> implementationVisitor = implementationVisitor();
+        JavaIsoVisitor<ExecutionContext> functionalVisitor = functionalVisitor(implementationVisitor);
+
+        //TODO the other visitors for new AuditorAware() {...} and method references.
+
```
Jenson3210, line 46:
```suggestion
```

src/test/java/org/openrewrite/java/spring/data/MigrateAuditorAwareToOptionalTest.java
timtebeek: If that helps here then we can make it public and pull it up; I'd intentionally written it as a reusable component, that I hadn't proactively pulled up just yet until there was another use case for it.

src/test/java/org/openrewrite/java/spring/data/MigrateAuditorAwareToOptionalTest.java
Jenson3210: well... at least it seems to have remove the cannot be converted error. 

But still seeing on 3 tests:
```
LST contains missing or invalid type information
MethodInvocation->Lambda->Return->Block->MethodDeclaration->Block->ClassDeclaration->CompilationUnit
/*~~(MethodInvocation type is missing or malformed)~~>*/Optional.ofNullable("admin")
```

src/test/java/org/openrewrite/java/spring/data/MigrateAuditorAwareToOptionalTest.java
Jenson3210: should we pull it up to rewrite-java?

timtebeek: I'd say not yet. Both uses are in rewrite-spring for now

Jenson3210:
Hi @timtebeek , 

I've pushed some comments you made in the past 3 rounds of rewiew. 
There are still 2 tests which I cannot get fixed with the Missing information:
- org.openrewrite.java.spring.data.MigrateAuditorAwareToOptionalTest#dontRewriteLambdaBlock
- org.openrewrite.java.spring.data.MigrateAuditorAwareToOptionalTest#dontRewriteLambdaLiteral

Strangly, both are in the don't -> lambda part. So it looks like the lambda in the test code can't be parsed. 
The other negated entries are fine :/

timtebeek (approved):
Great to see! The last remaining failing tests were using a method signature from 2.x as opposed to 1.x, so changing the parser classpath for those tests while still asserting no change is made helped get those to pass.

There might be some scenarios not covered like a class defined in a separate file outside of an immediate return, but we can tackle those separately if they even occur in your wider scan there.

Thanks again!

PR #617, opened 2024-11-06
Add BeanMethodReturnNull recipe

SiBorea:
## What's changed?
In SpringBoot 2, it is allowed to define a bean method returning void. Such trick is not allowed in SpringBoot 3.

### Checklist
- [✅] I've added unit tests to cover both positive and negative cases
- [✅] I've read and applied the [recipe conventions and best practices](https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices)
- [✅] I've used the IntelliJ IDEA auto-formatter on affected files


timtebeek (approved):
Great to see @SiBorea ; Would you mind adding a link here detailing when this was changed? I couldn't immediately find a reference, but it'd be good to have that here in case there's any questions in the future.

Apart from that I've applied some light refactoring and also changed the recipe you had likely looked to for inspiration. With that we're good to merge. Thanks again!

Issue #620, opened 2024-11-08
Custom HandlerInterceptor with Spring Boot 3 migration does not compile


jevanlingen:
## What is the smallest, simplest way to reproduce the problem?
```java
import javax.servlet.http.*;

class MyInterceptor extends MySuperInterceptor {
    @Override
    public boolean preHandle(HttpServletRequest request, HttpServletResponse response, Object handler) {
        return super.preHandle(request, response, handler);
    }
}
```

```java
import javax.servlet.http.*;
import org.springframework.web.servlet.HandlerInterceptor;

class MySuperInterceptor extends HandlerInterceptorAdapter {
    @Override
    public boolean preHandle(HttpServletRequest request, HttpServletResponse response, Object handler) {
        return super.preHandle(request, response, handler);
    }
}
```

## What did you expect to see?

```java
import javax.servlet.http.*;

class MyInterceptor extends MySuperInterceptor {
    @Override
    public boolean preHandle(HttpServletRequest request, HttpServletResponse response, Object handler) {
        return super.preHandle(request, response, handler);
    }
}
```

```java
import javax.servlet.http.*;
import org.springframework.web.servlet.HandlerInterceptor;

class MySuperInterceptor implements HandlerInterceptor {
    @Override
    public boolean preHandle(HttpServletRequest request, HttpServletResponse response, Object handler) {
        return HandlerInterceptor.super.preHandle(request, response, handler);
    }
}
```

## What did you see instead?
```java
import javax.servlet.http.*;

class MyInterceptor extends MySuperInterceptor {
    @Override
    public boolean preHandle(HttpServletRequest request, HttpServletResponse response, Object handler) {
        return HandlerInterceptor.super.preHandle(request, response, handler);
    }
}
```

```java
import javax.servlet.http.*;
import org.springframework.web.servlet.HandlerInterceptor;

class MySuperInterceptor implements HandlerInterceptor {
    @Override
    public boolean preHandle(HttpServletRequest request, HttpServletResponse response, Object handler) {
        return HandlerInterceptor.super.preHandle(request, response, handler);
    }
}
```

## What is the full stack trace of any errors you encountered?
The code does not even compile, because the import `HandlerInterceptor` import does not even exist for the subclass. But even if the import does exist, only classes which extends the HandlerInterceptor immediately should be changed!

jevanlingen:
[...]

PR #622, opened 2024-11-08
MigrateHandlerInterceptor recipe migrates too much

jevanlingen:
## What's changed?
Improve/Bugfix the `MigrateHandlerInterceptor`recipe to only migrate classes that directly extends the HandlerInterceptorAdapter.

## What's your motivation?
Migration is targeted broadly, so it can migrate `super` calls when it is not intended to. This can lead to bugs, see 
- fixes https://github.com/openrewrite/rewrite-spring/issues/620

for example.

## Any additional context
The migration does now convert `super.preHandle(..)` to `HandlerInterceptor.super.preHandle(..)`. According to mr [unconditional](https://stackoverflow.com/users/4571544/unconditional), you should migrate to `return true`. See [this](https://stackoverflow.com/questions/72152105/migrate-handlerinterceptor-to-spring-boot-2-6) SO post for more information.

_Also [this](https://sbrta.medium.com/spring-boot-2-to-3-a-migration-journey-92d5fb3a5b5b#7a20) blog seems to suggest that this is the right move (as in, when you move from SB 2 to 3, you likely have a `return true` kind of implementation)._

There are two other methods avaliable in the HandlerInterceptorAdapter/HandlerInterceptor classes: `postHandle` and `afterCompletion`. Both of them have empty default implementation, So maybe the recipe should be doing something else:

- **preHandle**: `super.preHandle(..)` -> `return true`
- **postHandle**: `super.postHandle(..)` -> `<remove super call>`
 - **afterCompletion**: `super.afterCompletion(..)` ->  `<remove super call>`

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/spring/boot2/MigrateDatabaseCredentials.java
+++ b/src/main/java/org/openrewrite/java/spring/boot2/MigrateDatabaseCredentials.java
@@ -83,7 +83,7 @@ public class MigrateDatabaseCredentials extends Recipe {
                 public Yaml visitDocuments(Yaml.Documents documents, ExecutionContext ctx) {
                     if (FindProperty.find(documents, "spring." + tool + ".username", true).isEmpty() &&
                             FindProperty.find(documents, "spring." + tool + ".password", true).isEmpty()) {
-                        doAfterVisit(new FindProperty("spring." + tool + ".url", true).getVisitor());
+                        doAfterVisit(new FindProperty("spring." + tool + ".url", true, null).getVisitor());
                     }
                     return documents;
                 }
--- a/src/main/java/org/openrewrite/java/spring/framework/MigrateHandlerInterceptor.java
+++ b/src/main/java/org/openrewrite/java/spring/framework/MigrateHandlerInterceptor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2024 the original author or authors.
+ * Copyright 2021 the original author or authors.
  * <p>
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/src/main/java/org/openrewrite/java/spring/framework/MigrateHandlerInterceptor.java
+++ b/src/main/java/org/openrewrite/java/spring/framework/MigrateHandlerInterceptor.java
@@ -15,12 +15,14 @@
  */
 package org.openrewrite.java.spring.framework;
 
+import org.jspecify.annotations.Nullable;
 import org.openrewrite.ExecutionContext;
 import org.openrewrite.Preconditions;
 import org.openrewrite.Recipe;
 import org.openrewrite.TreeVisitor;
 import org.openrewrite.java.JavaTemplate;
 import org.openrewrite.java.JavaVisitor;
+import org.openrewrite.java.MethodMatcher;
 import org.openrewrite.java.search.UsesType;
 import org.openrewrite.java.tree.J;
 import org.openrewrite.java.tree.JavaType;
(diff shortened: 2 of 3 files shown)


src/main/java/org/openrewrite/java/spring/framework/MigrateHandlerInterceptor.java
```diff
@@ -30,6 +29,9 @@
 
 public class MigrateHandlerInterceptor extends Recipe {
 
+    public static final String HANDLER_INTERCEPTOR_ADAPTER = "org.springframework.web.servlet.handler.HandlerInterceptorAdapter";
+    public static final String HANDLER_INTERCEPTOR = "org.springframework.web.servlet.HandlerInterceptor";
```
timtebeek, line 33: There's no need for these to be public right?
```suggestion
    private static final String HANDLER_INTERCEPTOR_ADAPTER = "org.springframework.web.servlet.handler.HandlerInterceptorAdapter";
    private static final String HANDLER_INTERCEPTOR = "org.springframework.web.servlet.HandlerInterceptor";
```

src/testWithSpringBoot_2_4/java/org/openrewrite/java/spring/MigrateHandlerInterceptorTest.java
```diff
@@ -66,4 +67,54 @@ public boolean preHandle(HttpServletRequest request, HttpServletResponse respons
...
+              """,
+            """
+              import javax.servlet.http.*;
+
+              import org.springframework.web.servlet.HandlerInterceptor;
+
+              class MySuperInterceptor implements HandlerInterceptor {
+                  @Override
+                  public boolean preHandle(HttpServletRequest request, HttpServletResponse response, Object handler) {
+                      return HandlerInterceptor.super.preHandle(request, response, handler);
```
timtebeek, line 98: In your comments you pointed out that this should be 
```suggestion
                      return true;
```
And looking at [HandlerInterceptor.java#L103-L107](https://github.com/spring-projects/spring-framework/blob/80b72151a7984b5151311d93c59e0f46b61c2c50/spring-webmvc/src/main/java/org/springframework/web/servlet/HandlerInterceptor.java#L103-L107) I tend to agree; what made you not make that change just yet?

jevanlingen: There is a difference between bugfixing and changing the intention of things. I tend to separate the two, unless discussed. As you agree that the intention of this recipe should be changed, I will make a change accordingly.

src/testWithSpringBoot_2_4/java/org/openrewrite/java/spring/MigrateHandlerInterceptorTest.java
```diff
@@ -114,6 +134,25 @@ public boolean preHandle(HttpServletRequest request, HttpServletResponse respons
...
+                      return super.test();
+                  }
+              }
+
+              class A {
+                  public String test() {
+                      return "test";
+                  }
+              }
+              """
```
timtebeek: Just because there's an import does not mean the type is in use on any LST elements. As such the precondition will fail here, and hence no changes will be applied, whereas you probably intended to test that no changes are made even when the precondition matches; can you either remove this block or ensure the precondition matches and no change is made?

src/main/java/org/openrewrite/java/spring/boot2/MigrateDatabaseCredentials.java
```diff
@@ -83,7 +83,7 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
                 public Yaml visitDocuments(Yaml.Documents documents, ExecutionContext ctx) {
                     if (FindProperty.find(documents, "spring." + tool + ".username", true).isEmpty() &&
                             FindProperty.find(documents, "spring." + tool + ".password", true).isEmpty()) {
-                        doAfterVisit(new FindProperty("spring." + tool + ".url", true).getVisitor());
+                        doAfterVisit(new FindProperty("spring." + tool + ".url", true, null).getVisitor());
```
jevanlingen, line 86: This change was needed to make it compile again. The `FindProperty` classes has just been changed by: 
- https://github.com/openrewrite/rewrite/pull/4657

src/testWithSpringBoot_2_4/java/org/openrewrite/java/spring/MigrateHandlerInterceptorTest.java
jevanlingen: Ah good one, changed it!

src/main/java/org/openrewrite/java/spring/boot2/MigrateDatabaseCredentials.java
timtebeek: That type of change may be pushed to main directly. No need to do that now, but just so you know.

src/main/java/org/openrewrite/java/spring/framework/MigrateHandlerInterceptor.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2021 the original author or authors.
+ * Copyright 2024 the original author or authors.
```
timtebeek, line 2: No need to update these copyright headers; only for new recipes do we want those to reflect the year they were added.

timtebeek (approved):

PR #626, opened 2024-11-14
Drop the old runtime Spring Core dependency brought in through spring-data-mongodb:2.+

timtebeek:
As that's blocked due to vulnerabilities in some places, but was proving troublesome when installing recipes jars if the excluded dependency poms can not be found. This seemed the quickest way forward after also exploring the downloads in openrewrite/rewrite but finding no easy way forward there.


timtebeek:
@Laurens-W does the failing test here and in main depend on 
- https://github.com/openrewrite/rewrite/pull/4670 ?

PR #630, opened 2024-11-18
RenameBean Scanning Recipe Refactor

ryan-hudson:
## What's changed?
Refactored the RenameBean recipe into a scanning recipe that accumulates the necessary bean renames in the scanning phase then performs these changes across all files during the visiting phase. 

## What's your motivation?
This change allows the RenameBean recipe to rename usages of the bean that occur outside of the file the bean is declared in. This is a bug in the current version of this recipe that is resolved with this change. (See added test cases for examples.)


timtebeek (approved):
Thanks @ryan-hudson ! Should indeed help not to break tests that do not use the context setup.

timtebeek:
Apologies, wasn't aware, and had discussed this with Nick as well; happy to bring those back; let me know if you'd want me to do so or if you're already working on it.

ryan-hudson:
I've gone ahead and opened another PR restoring them [here](https://github.com/openrewrite/rewrite-spring/pull/640)

Issue #265, opened 2023-01-10
Is the best place to SpringBatch 5.0 migration recipes ?


desprez:
Springboot 3.x [Spring-Boot-3.0-Release-Notes](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-3.0-Release-Notes) comes with new version 5.0 of Spring-batch and lots of deprecated stuff and changes.
Do  you plan to add recipes for this migration, here or in annoter repository ?

timtebeek:
This repository would probably be the best place for such migration recipes, similar to how we've included recipes for Spring Data migrations here as well. I imagine a new package `org.openrewrite.java.spring.batch` should be enough to get started for now. Would you be willing to contribute migration recipes to kick this off?

I'll link the [Spring Batch v5.0.0 Release](https://github.com/spring-projects/spring-batch/releases/tag/v5.0.0) and [Migration Guide](https://github.com/spring-projects/spring-batch/wiki/Spring-Batch-5.0-Migration-Guide) here for others as well.

Quickly looking through the migration guide there's: 
- [ ] [Dependency upgrades](https://github.com/spring-projects/spring-batch/wiki/Spring-Batch-5.0-Migration-Guide#dependencies-upgrade) such as Spring Framework 6, Jakarta EE 9, Hibernate 6.
  - There are already existing efforts for these migrations that we can link to in a Spring Batch 5 migration
- [ ] Database schema updates
  - Not clear to me if these would require a recipe.
- [ ] [Job repository/explorer configuration updates](https://github.com/spring-projects/spring-batch/wiki/Spring-Batch-5.0-Migration-Guide#job-repositoryexplorer-configuration-updates)
- [ ] [Transaction manager bean exposure/configuration](https://github.com/spring-projects/spring-batch/wiki/Spring-Batch-5.0-Migration-Guide#transaction-manager-bean-exposureconfiguration)
- [ ] [JobBuilderFactory and StepBuilderFactory bean exposure/configuration
](https://github.com/spring-projects/spring-batch/wiki/Spring-Batch-5.0-Migration-Guide#jobbuilderfactory-and-stepbuilderfactory-bean-exposureconfiguration)
- [ ] [Data types updates](https://github.com/spring-projects/spring-batch/wiki/Spring-Batch-5.0-Migration-Guide#data-types-updates)
[...]

PR #631, opened 2024-11-19
Add ChangeMethodParameter for modify parameters in method declaration

ckcd:
Continue to complete https://github.com/openrewrite/rewrite/pull/4683

@timtebeek as we discussed, I put the new PR here, please take a look at this if you are available, thanks a lot.

## What's changed?

Add a `ChangeMethodParameter` recipe for modify parameters in method declaration.
For now we have [AddMethodParameter](https://github.com/openrewrite/rewrite/blob/main/rewrite-java/src/main/java/org/openrewrite/java/AddMethodParameter.java) which is for add new parameter in method declaration, but didn't have the one to modify current parameters.

@timtebeek as we discussed in slack, I tried to add this new `ChangeMethodParameter` recipe, please take a look at this if you are available, thanks a lot.

## What's your motivation?

I want to find any recipe in java can replace a parameter's type, for example:
```java
// From:
public void m(int a) {}

// To:
public void m(long a) {}
```
but currently we don't have. So I try to add this new recipe.

### Checklist


timtebeek:
Thanks a lot @ckcd ! Would it make sense to immediately add recipes for some of those updated Spring Batch data types?
- https://github.com/openrewrite/rewrite-spring/issues/265

That way we have the context of why this was added and can try it out with an implementation.

ckcd:
@timtebeek Thanks for your reply!

I add a `UpgradeSkipPolicyParameterType` in `spring-batch-5.0.yml`, also I add the test case for it. PTAL again, thanks.

ckcd:
@timtebeek Thanks for your fix!

For now I saw all checks are passed, is there anything I need to do?

timtebeek (approved):
Much appreciated the careful planning ahead to make this recipe already support types not yet used, and then putting it to work here for Spring Batch as well to prove it works. 🙏🏻

ckcd:
@timtebeek Thanks a lot!

PR #632, opened 2024-11-20
Fix incorrect path computation for Spring API endpoints

adrilo:
## What's changed?
The current logic did not cover the case where a `RequestMapping` with a path was defined on the Controller class, and no paths were defined in the `*Mapping` at the method level.

### Checklist


timtebeek:
Thanks a lot for the fix here @adrilo ; Out of curiosity: did you already run this recipe through the Moderne CLI against local projects?

I'll fix the test failures separately before we merge; Those came in after
- https://github.com/openrewrite/rewrite/pull/4688

timtebeek (approved):

Issue #634, opened 2024-11-22
Provide inline comment for deprecated and removed properties


ashakirin:
Currently, OpenRewrite provides a way to generate recipes for renaming properties when a property key has changed, using:
`Run [GeneratePropertiesMigratorConfiguration.java]`.

However, when a property is removed or deprecated without an alternative, this change is not reflected in the migrated properties.

Automatically removing such properties could lead to unexpected issues because users:

- May lose important values stored in these properties.
- Might still reference these properties in custom code unrelated to Spring.

_Proposed Solution:_
I suggest adding inline comments for removed or deprecated properties (without alternatives) during the migration process. This would make the deprecation or removal clear to users, allowing them to review and decide whether to manually remove these properties as needed.

timtebeek:
Thanks again @ashakirin !

PR #635, opened 2024-11-22
feat: Added recipe to comment deprecated and removed properties witho…

ashakirin:
Added recipe to comment deprecated and removed properties without alternative
- Refs: #634

### DIFF
--- a/src/main/java/org/openrewrite/java/spring/InlineCommentSpringProperties.java
+++ b/src/main/java/org/openrewrite/java/spring/InlineCommentSpringProperties.java
@@ -17,7 +17,6 @@ package org.openrewrite.java.spring;
 
 import lombok.EqualsAndHashCode;
 import lombok.Value;
-import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.openrewrite.*;
 
--- a/src/main/java/org/openrewrite/java/spring/InlineCommentSpringProperties.java
+++ b/src/main/java/org/openrewrite/java/spring/InlineCommentSpringProperties.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2021 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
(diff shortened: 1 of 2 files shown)


timtebeek (commented):
Thanks for the addition already @ashakirin ! I spoke to @BoykoAlex last Thursday who had been looking for something similar for removed properties. I think this could factor in to the solution there.

As a bit of background we automatically generate recipes for properties that have moved, based on meta data available in the Spring Boot jars, as seen here:
https://github.com/openrewrite/rewrite-spring/blob/2eebf1d8ebbfaa7063e381d9235c2ea2fcc3778a/src/test/java/org/openrewrite/java/spring/internal/GeneratePropertiesMigratorConfiguration.java#L116-L139

I think we can similarly generate recipes that add comments for meta data without a listed replacement, as for instance seen here:
https://github.com/spring-projects/spring-boot/blob/32433e84f3048c397aef3c3a8d63a0ae1ca7d5fc/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json#L351-L359

If we are to go that route then there is no need for this recipe to include a `List<String>`, as matching a single key is good enough when we generate the recipes.

@BoykoAlex what are your thoughts on this?

BoykoAlex:
@timtebeek @ashakirin I think I'm inclined to look at the recipe generation first and then come up with the appropriate shape of the recipe putting a comment. I don't know if Spring Boot Properties metadata json contains info about removed properties... Perhaps we'd need to compute removed properties based on X.Y and X.Y+1 metadata comparison? Not sure... Anyway, if it is simpler to consume removed properties one-by-one from the generator code then a single property key parameter. If it comes in as a list/collection already then a list of keys is good. To be honest a list parameter has wider usage options and likely more performant  so I'd say I'm fine with its current shape... This recipe feels like its home should be `rewrite-properties` project under `rewrite` repo...

timtebeek:
Thanks for the reply! I'd prefer to keep it here for now, as it uses `ChangeSpringPropertyValue` which has some Spring specific matching, and works across both Yaml and Properties.

As for removed properties I _think_ those are called out as errors without replacements, sometimes with a reason, as shown on [the link above](https://github.com/spring-projects/spring-boot/blob/32433e84f3048c397aef3c3a8d63a0ae1ca7d5fc/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json#L351-L359). When there's a reason we can add that inline; if there's none then the error at least indicates a user should have a look to this likely removal.

ashakirin:
I would like to split it in some small steps:
1) add recipes for inline comments (current PR)
2) add inline comment for deprecated properties without replacements, like @timtebeek suggested (seems to be easy)
3) analyse options for removed properties: when we can recognize them using deprecations with errors, should be trivial as well. For removed properties we can even comment the property itself + add inline comment or just leave inline comment in other format, WDYT?

timtebeek:
- Superseded by https://github.com/openrewrite/rewrite-spring/pull/639

Thanks again! Let's continue there.

PR #640, opened 2024-11-26
Restore Existing Static RenameBean Constructors

ryan-hudson:
## What's changed?
Brief PR to restore the removed `fromDeclaration` static methods to allow for programmatic usage in custom recipes


timtebeek (approved):

timtebeek:
Thanks for the quick restore; sorry I hadn't registered that these were actually in use before! :)

PR #641, opened 2024-11-26
fix test dontEnableVirtualThreadsIfDisabled

sullis:


timtebeek:
Thanks for the quick fix! Now I'm wondering why it hadn't failed on main before.. 🤔

PR #644, opened 2024-12-04
Add logic to quote yaml scalar value

tkvangorder:
## What's changed?
Ran into an edge case when attempting to change the value of `management.endpoints.web.exposure.include` to `*` and the current implementation does not quote the value when properties are expressed as YAML. Simply added logic that is similar to the logic used in AddSpringProperty to quote the scalar value.


timtebeek (approved):
Looks good! Thanks for the quick fix here.

PR #645, opened 2024-12-05
Remove BeanMethodsNotPublic from SpringBoot2BestPractices (again)

nmck257:
- See #304 for exactly the same conversation :)

That said, getting philosophical:
I think we typically have a design pattern where "upgrades" and "best practices" are kept separate, and "best practices" allows for more opinionated/noisy changes. That's what the JUnit 5 recipes look like, iirc.

This file violates that pattern by having the 2.0 upgrade recipe include the best practices recipe. Buuuut, it's been that way for over a year, and the other included recipes aren't nearly as disruptive as this one.

I could imagine a spectrum with one extreme having only the most-strictly necessary changes, and the other end having the most noisy and least-popularly-agreed-upon changes, and we slice that spectrum into sections like:
- 1: "upgrades"
- 2: "best practices"
- 3: "standalone recipes"
- 4: "not even recipes"

And in that model, maybe the other recipes in `SpringBoot2BestPractices` happen to fall closer to section 1, and `BeanMethodsNotPublic` is in section 3, and the composite recipe happens to be named for section 2 and then is included in the section 1 `UpgradeSpringBoot_2_0` recipe. So it's easy for someone to think "this is a section 2 recipe, lemme add it to that bucket" and then it sneaks into section 1 and that feels bad. But it's all fuzzy.

### DIFF
--- a/src/main/resources/META-INF/rewrite/best-practices.yml
+++ b/src/main/resources/META-INF/rewrite/best-practices.yml
@@ -15,6 +15,23 @@
 #
 ---
 type: specs.openrewrite.org/v1beta/recipe
+name: org.openrewrite.java.spring.boot2.SpringBoot2BestPractices
+displayName: Spring Boot 2.x best practices
+description: Applies best practices to Spring Boot 2 applications.
+tags:
+  - spring
+  - boot
+recipeList:
+  # Note that we do not upgrade to Spring Boot 2.x here, as the 2.0 recipe includes SpringBoot2BestPractices itself
+  - org.openrewrite.java.spring.NoRequestMappingAnnotation
+  - org.openrewrite.java.spring.ImplicitWebAnnotationNames
+  - org.openrewrite.java.spring.boot2.UnnecessarySpringExtension
+  - org.openrewrite.java.spring.NoAutowiredOnConstructor
+  - org.openrewrite.java.spring.boot2.RestTemplateBuilderRequestFactory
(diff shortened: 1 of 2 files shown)


timtebeek:
Thanks for the nuanced discussion here @nmck257 ; sorry for any disruption this might have caused. It seems to have slipped in
- https://github.com/openrewrite/rewrite-spring/pull/617

Would you be ok with moving that recipe to the Spring Boot 3.x best practices? That one seemed closer aligned with 2. above,
and is _not_ hooked into any of the upgrades by default. We similarly moved the virtual threads recipe there as well in https://github.com/openrewrite/rewrite-spring/commit/c3a79c92e44a92717148cb7fead82fdf3eede791, which is also opinionated and not a perfect fit everywhere.

[The full 2.0 best practices recipe](https://github.com/openrewrite/rewrite-spring/blob/a6c60335578f4f6b2fa9cdf9255388c084e4f462/src/main/resources/META-INF/rewrite/spring-boot-20.yml#L171-L186) can move to `src/main/resources/META-INF/rewrite/best-practices.yml` as well, probably with a note to prevent another regression.

timtebeek:
I've applied my suggestions from above; hope you agree! Figured with the note and state goals this would now be clear.

nmck257:
sorry for the delay -- yeah, no objections to your changes. thanks for moving this along!
I might draft a PR later that goes a step further to lift some of the existing SB2 "best practices" recipes up to the upgrade recipe itself, and remove "best practices" from the upgrade recipe; haven't decided myself yet

Issue #642, opened 2024-11-28
Recipes to upgrade to Spring Boot 3.4


timtebeek:
We want to support folks looking to migrate to Spring Boot 3.4
- https://spring.io/blog/2024/11/21/spring-boot-3-4-0-available-now
- https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-3.4-Release-Notes

Here's an incomplete list of tasks related to this migration
- [x] https://github.com/openrewrite/rewrite-spring/issues/627
- [ ] Generate Spring Boot property migration recipes ([runs on a schedule](https://github.com/openrewrite/rewrite-spring/blob/03bf6f1865355e294b52d38fc864ca29a1f90100/.github/workflows/properties.yml#L2-L7))
- [x] WebJars Locator Integration: instead of `org.webjars:webjars-locator-core` depend on `org.webjars:webjars-locator-lite`
- [x] OkHttp Dependency Management Removed
  - verify change parent already adds the explicit version where that was managed before
- [ ] Netty in Native Image: upgrade the GraalVM reachability metadata version
- [x] https://github.com/openrewrite/rewrite-spring/pull/646
- [x] Deprecations in `RestTemplateBuilder`
  - requestFactory -> requestFactoryBuilder
  - setConnectTimeout ->connectTimeout
  - setReadTimeout -> readTimeout
  - setSslBundle -> sslBundle
- [x] Bean configuration properties need to be marked with `@Valid` where previously only `@Validated` was required [see](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-3.4-Release-Notes#bean-validation-of-configuration-properties)

123Haynes:
I created a recipe for the last point (Replacing @MockBean and @SpyBean) which I can contribute.  😃 
Probably needs some cleanup though since this is my first recipe. I'll try to create a pr for it tomorrow after work.

123Haynes:
I also found one more point for the list. The RestTemplateBuilder has several deprecations with Spring Boot 3.4  
https://docs.spring.io/spring-boot/api/java/org/springframework/boot/web/client/RestTemplateBuilder.html

requestFactory -> requestFactoryBuilder
setConnectTimeout ->connectTimeout
setReadTimeout -> readTimeout
setSslBundle -> sslBundle


@timtebeek Can you add that as one TODO item on the list so it isn't forgotten? 😃

123Haynes:
[...]

PR #647, opened 2024-12-06
Add OkHttp Recipe

123Haynes:
## What's changed?

This PR adds a recipe that explicitly adds OkHttp as a dependency if it's used in the codebase.  
I copied the OkHttp version from the `spring-boot-dependencies.pom` of Spring Boot 3.3.6.

## What's your motivation?

- Part of https://github.com/openrewrite/rewrite-spring/issues/642

I'm unsure if adding the okhttp jar directly to the classpath folder was the correct way to handle the requirement in the test case.

### Checklist


timtebeek:
Thanks for kicking this off! I think the issue is slightly different, and likely already handled by ChangeParentPom.

Previously a project using say the 3.3 parent from Spring Boot would not have to define a version for Okhttp, as that was a managed dependency. With 3.4 that managed dependency was removed, so folks would have to add an explicit version going forward. I believe the change parent pom recipe already takes care of that, so then we can keep this test around but likely do not require a dedicated recipe beyond changing the parent pom version.

123Haynes:
@timtebeek Thanks for pointing that out. I didn't know the ChangeParentPom recipe handled that case 😄   
I'm fine with closing this PR then. I'll just treat it as a learning experience. 

Maybe we should mark the TODO for OkHttp as completed in https://github.com/openrewrite/rewrite-spring/issues/642 then and add a comment that this will be handled by `org.openrewrite.maven.UpgradeParentVersion `when the final upgrade recipe is created.  
looks like ChangeParentPom is not used directly in this repo so far, but instead UpgradeParentVersion is used which invokes it 😄

timtebeek:
Good call; I've updated the TODO list, and added a reminder to check, possibly with the test you've added here. Thanks again!

PR #652, opened 2024-12-23
Keep annotations for merged classes for WebSecurityConfigurerAdapter recipe (Spring Security 5.4)

jevanlingen:
## What's changed?
Currently al annotations on the class, when converting

```java
@Configuration
@AllOtherAnnotations
public static class SomeSecurityConfigurerAdapter extends WebSecurityConfigurerAdapter { .. }
```

to 

```java
@Bean
SecurityFilterChain someSecurityFilterChain(HttpSecurity http) throws Exception { .. }
```

, are dropped.

This PR lets them keep them:

```java
@Bean
@AllOtherAnnotations
SecurityFilterChain someSecurityFilterChain(HttpSecurity http) throws Exception { .. }
```

## What's your motivation?
Just dropping the annotations can lead to serious bugs. For example, if a `@ConditionalOnProperty` annotation has been used, dropping it means different behaviour from the application.

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/spring/security5/WebSecurityConfigurerAdapter.java
+++ b/src/main/java/org/openrewrite/java/spring/security5/WebSecurityConfigurerAdapter.java
@@ -32,7 +32,6 @@ import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
 
 import static java.util.Collections.emptyList;
-import static java.util.stream.Collectors.toList;
 
 /**
  * @author Alex Boyko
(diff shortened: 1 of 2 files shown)


src/main/java/org/openrewrite/java/spring/security5/WebSecurityConfigurerAdapter.java
```diff
@@ -144,15 +143,7 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
                         }
                         // only applicable to former subclasses of WebSecurityConfigurerAdapter - other classes won't be flattened
                         classesToFlatten.add(classDecl);
-                        // Remove imports for annotations being removed together with class declaration
-                        // It is impossible in the general case to tell whether some of these annotations might apply to the bean methods
```
jevanlingen, line 148: Though this comment is true, choosing to not do anything with the annotations, is a wrong move. This can lead to bugs, as I have shown in the description of this PR. It's better/safer to copy them and hope for the best 🤞🏻.

timtebeek (approved):
Great to see this added & resolved quickly!

PR #653, opened 2024-12-30
Replace Assert.notEmpty(Collection<?> collection) with Assert.notEmpty(Collection<?> collection, String)

EricleeCrystal:
## What's changed?
The Assert.notEmpty(Collection<?> collection) was deprecated of 4.3.7, and removed in 6.1 in favor of Assert.notEmpty(Collection, String).

## What's your motivation?
To prevent potential compilation errors during the upgrade to Spring Boot 3.3.X.

No.

No.

I don't think there is other workaround because this method will be removed in 6.1.

## Any additional context
No.

### Checklist
No.


timtebeek (approved):
Thanks @EricleeCrystal !

PR #663, opened 2025-01-11
SimplifyWebTestClientCalls does not work on Kotlin code

barbulescu:
## What's changed?
Add support for both int (Java) and long (Kotlin) status code types.

## What's your motivation?
Have this recipe work on Kotlin code.

## Any additional context
Followed the [instructions](https://discord.com/channels/996548279191998535/996548280118935674/threads/1327645523288657973) on Discord.

Wanted to add support for HttpStatus argument but one extra step is to remove the import if not used anywhere else.
Decided to have a follow-up PR for this fix.

### Checklist


timtebeek:
Thanks a lot @barbulescu ; you've stumbled upon an issue with the underlying KotlinParser, which I've logged here:
- https://github.com/openrewrite/rewrite-kotlin/pull/624

That meant that literals of the type `int` where returning a `Long` value as you've discovered, which then needed the casts seen here. Ideally we fix this upstream; I'll try my hand at that first.

timtebeek:
You've uncovered another imperfection with the way we represent types that come from Java libraries in Kotlin:
- https://github.com/openrewrite/rewrite-kotlin/pull/626

That'll negatively affect our ability to match methods used in Kotlin code, such that existing recipes might not match where they should. We should take a closer look in rewrite-kotlin in how that type information is built up, and ensure Java types are used, or adjust the matcher to match both Java and Kotlin types for `int` and kotlin.Int` and others.

timtebeek (approved):
Thanks for the necessary adjustments to make this work for Kotlin as well, and helping uncover two imperfections upstream in rewrite-kotlin, and the help in getting one of those cleared out already. We'll merge this iteration and can separately look to cover the HttpStatus case you're after indeed.

PR #676, opened 2025-02-12
Adopt TypeTable

timtebeek:
## What's changed?
- [x] Remove packaged resource jars and replace those with TypeTable

Needs a new
- [x] [release of openrewrite](https://github.com/openrewrite/rewrite/releases) v8.45.3+
- [x] [rewrite-build-gradle-plugin](https://github.com/openrewrite/rewrite-build-gradle-plugin/releases) v1.126.2

## What's your motivation?
Packaged resource jars could trip up scanners, even if the jars are not the classpath.

Over time we have also added test resource jars; these are typically used only to compile the before-test-sources, and not needed on the runtime classpath, because there JavaTemplate only needs the after-jar-versions. I think I'd prefer to keep that this way, and not unnecessarily inflate the `.tsv` file further; especially seeing how we also have to deflate that on startup, which could cause a bit of an IO spike.

## Any additional context
- https://github.com/openrewrite/rewrite-build-gradle-plugin/pull/75
- https://github.com/openrewrite/rewrite/pull/4932
- https://docs.openrewrite.org/reference/faq#why-do-artifact-scanners-detect-vulnerabilities-in-recipe-artifactsjars


sambsnyd (approved):

timtebeek:
Ok, next fun edge case discovered on a failing test here: meta annotations.
https://github.com/openrewrite/rewrite-spring/blob/70153755a1077235c1064e01b0703898ce0a650d/src/test/java/org/openrewrite/java/spring/RenameBeanTest.java#L1034
```
~/.rewrite/classpath/.tt/org/springframework/spring-context/6.+/org/springframework/stereotype$ javap Service.class 
public interface org.springframework.stereotype.Service extends java.lang.annotation.Annotation {
  public abstract java.lang.String value();
}
```

timtebeek:
Not seeing the same failure locally; this is the decompiled class as extracted:
```
~/.rewrite/classpath/.tt/org/springframework/spring-web/5.+/org/springframework/web/server$ javap ResponseStatusException.class 
```
```java
public class org.springframework.web.server.ResponseStatusException extends org.springframework.core.NestedRuntimeException {
  public org.springframework.web.server.ResponseStatusException(org.springframework.http.HttpStatus);
  public org.springframework.web.server.ResponseStatusException(org.springframework.http.HttpStatus, java.lang.String);
  public org.springframework.web.server.ResponseStatusException(org.springframework.http.HttpStatus, java.lang.String, java.lang.Throwable);
  public org.springframework.web.server.ResponseStatusException(int, java.lang.String, java.lang.Throwable);
  public org.springframework.http.HttpStatus getStatus();
  public int getRawStatusCode();
  public java.util.Map<java.lang.String, java.lang.String> getHeaders();
  public org.springframework.http.HttpHeaders getResponseHeaders();
  public java.lang.String getReason();
  public java.lang.String getMessage();
}
```

timtebeek:
@sambsnyd Given the limitations seen here and the potentially continued use of (some) packaged resource jars, how would you feel about [swapping the order of these two lines](https://github.com/openrewrite/rewrite/blob/7e6da9ae40b83f54a6597bef008c4f21c827b14c/rewrite-java/src/main/java/org/openrewrite/java/JavaParser.java#L130-L131) such that packaged jars take precedent? That would make it a lot easier to add a type table for all, while still keeping around the odd jar where needed. Now it's a bit awkward having to download recipe jars for a select few only by commenting out the rest, and then creating a type table with the inverse such to the jars are not ignored because of fault entries in type tables.

timtebeek:
Endless fun to be had! [TypeTable uses getResourceAsStream](https://github.com/openrewrite/rewrite/blob/f79215982fdb9182c33cce6c2b273b7fb5d95895/rewrite-java/src/main/java/org/openrewrite/java/internal/parser/TypeTable.java#L83-L87), which is documented as [unstable ordering when multiple such resources exist](https://github.com/openjdk/jdk/blob/fa1bd2344e60163bf247c668b94f98c50c72855a/src/java.base/share/classes/java/lang/ClassLoader.java#L1378-L1380).

> Where several modules are defined to the same class loader, and where more than one module contains a resource with the given name, then the ordering that modules are searched is not specified and may be very unpredictable.

`rewrite-migrate-java` contains a `META-INF/rewrite/classpath.tsv.zip`, as does `rewrite-spring` itself, and both are in the same classloader. We're seeing here that the tsv from `rewrite-migrate-java` is picked up in `RestTemplateBuilderRequestFactoryVisitor` from `rewrite-spring`.
https://github.com/openrewrite/rewrite-spring/blob/4626502424d3e4c8c084eb403af716da39cd5e1d/src/main/java/org/openrewrite/java/spring/boot2/RestTemplateBuilderRequestFactory.java#L65-L66

timtebeek:
Next issue: we're storing in the `.tsv.zip` and inside `.tt` the requested `spring-context-5.+`, which then fails to match say `spring-context-5.3.*`. We likely should adjust TypeTable to use the actual version, not the pattern that was requested.

timtebeek:
Perfect, just what I was hoping for: a quick apply the plugin, run the task, and off we go!

PR #680, opened 2025-02-19
Upgrade spring-security-oauth2-authorization-server

Laurens-W:
## What's changed?

Made `spring-security-oauth2-authorization-server` part of spring security upgrades

@timtebeek 

## Any additional context

https://github.com/spring-projects/spring-authorization-server/releases

### Checklist

### DIFF
--- a/src/main/resources/META-INF/rewrite/spring-security-63.yml
+++ b/src/main/resources/META-INF/rewrite/spring-security-63.yml
@@ -16,25 +16,24 @@
 
 ---
 type: specs.openrewrite.org/v1beta/recipe
-name: org.openrewrite.java.spring.security6.UpgradeSpringSecurity_6_2
-displayName: Migrate to Spring Security 6.2
+name: org.openrewrite.java.spring.security6.UpgradeSpringSecurity_6_3
+displayName: Migrate to Spring Security 6.3
 description: >-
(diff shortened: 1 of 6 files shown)


src/main/resources/META-INF/rewrite/spring-security-63.yml
```diff
@@ -0,0 +1,39 @@
...
+  - org.openrewrite.java.dependencies.UpgradeDependencyVersion:
+      groupId: org.springframework.security
+      artifactId: "*"
+      newVersion: 6.3.x
+      overrideManagedVersion: false
+  - org.openrewrite.java.dependencies.UpgradeDependencyVersion:
+      groupId: org.springframework.security
+      artifactId: spring-security-oauth2-authorization-server
+      newVersion: 1.3.x
+      overrideManagedVersion: false
```
timtebeek, line 39: I'm wondering about the ordering here and above; I think we validate the versions we migrate to actually exist so it should not be an issue that `org.springframework.security:*:6.3.x` also matches `spring-security-oauth2-authorization-server`, but perhaps it can't hurt to be sure. Did you check that part?

Laurens-W: Did not verify that, let me do that!

PR #682, opened 2025-03-03
Adjust expectation as seen in CI

timtebeek:
This seems wrong; we're not seeing a matching import, as that import is qualified; Either the import should change too, or this PR should not be necessary. Any insights as to recent related changes @knutwannheden ? I think I saw something come by.


timtebeek:
Specifically this change seems related;
- https://github.com/openrewrite/rewrite/pull/5080

And the failures seem to have started after that was merged:
https://github.com/openrewrite/rewrite-spring/actions

timtebeek:
This has been applied by @sambsnyd in https://github.com/openrewrite/rewrite-spring/commit/0ce0e771f6de5179bab528bfd93326732a2f1645; we'll likely want the upstream fix still, but for now can close this PR.

sambsnyd:
I could probably be talked into either `ChangeType` behavior. Fine with reverting my change if we land somewhere else on what its behavior should be.

timtebeek:
We had a similar issue with rewrite-logging-frameworks, which was fixed by changing the type pattern used there:
https://github.com/openrewrite/rewrite-logging-frameworks/commit/74b90e59b5b2ac5ebbaa7b1ce6baa47dcdcef6d0

PR #683, opened 2025-03-05
Add KafkaFailureCallback handler

SiBorea:
## What's changed?
Handle a corner case of using `KafkaFailureCallback` in `ListenableFuture`

### Checklist


timtebeek (approved):
Thanks for the added case here @SiBorea ! I've added some light polish to fix a few of the type issues; the remaining ones are a bit puzzling but not worth more time right now.

Issue #627, opened 2024-11-14
Recipes to upgrade to Spring Framework 6.2


timtebeek:
We want to support migrations to Spring Framework 6.2, as part of a migration to Spring Boot 3.4.

Here's a quick link to the release notes: https://github.com/spring-projects/spring-framework/wiki/Spring-Framework-6.2-Release-Notes

And here's an incomplete list of necessary steps for this migration:
- [x] Deprecated APIs removed in https://github.com/spring-projects/spring-framework/issues/30608
  - https://github.com/openrewrite/rewrite-spring/pull/693
  - https://github.com/openrewrite/rewrite-spring/pull/689
- [x] Additional APIs removed in https://github.com/spring-projects/spring-framework/issues/33123
- [ ] WebContentGenerator removed in https://github.com/spring-projects/spring-framework/issues/31492
- [ ] Migrating from HtmlUnit 2.x.x to HtmlUnit 3.x.x as per https://htmlunit.sourceforge.io/migration.html
  - [ ] ~https://github.com/openrewrite/rewrite-testing-frameworks/pull/643~
  - [ ] Include the HtmlUnit migrations with the Spring Framework 6.2 upgrade: https://docs.openrewrite.org/recipes/java/testing/htmlunit/upgradehtmlunit_3

We welcome anyone to pick off individual items here in pull requests, even if it's just a draft PR with unit tests added.

timtebeek:
hi! Those recipes first have to be written and contributed to the project; I've logged a small break down of tasks for Spring Boot:
- https://github.com/openrewrite/rewrite-spring/issues/642

Welcome to help out on either the steps here or on that other issue.

timtebeek:
This PR from @wapkch helpfully includes some of the steps above already:
- https://github.com/openrewrite/rewrite-spring/pull/693

PR #689, opened 2025-03-27
Add Recipe to migrate `org.springframework.util.Base64Utils` to `java.util.Base64`

wapkch:
## What's changed?

Migrate `org.springframework.util.Base64Utils` to `java.util.Base64`

## What's your motivation?

- Addresses part of https://github.com/openrewrite/rewrite-spring/issues/627

### Checklist


src/main/java/org/openrewrite/java/spring/framework/MigrateBase64Utils.java
```diff
@@ -0,0 +1,114 @@
...
+
+    private static final JavaTemplate decodeFromString = JavaTemplate.builder("Base64.getDecoder().decode(#{any(String)})").imports(JAVA_UTIL_BASE_64).build();
+
+    private static final JavaTemplate encodeUrlSafe = JavaTemplate.builder("Base64.getUrlEncoder().encode(#{anyArray(byte)})").imports(JAVA_UTIL_BASE_64).build();
+
+    private static final JavaTemplate decodeUrlSafe = JavaTemplate.builder("Base64.getUrlDecoder().decode(#{anyArray(byte)})").imports(JAVA_UTIL_BASE_64).build();
+
+    private static final JavaTemplate encodeToUrlSafeString = JavaTemplate.builder("Base64.getUrlEncoder().encodeToString(#{anyArray(byte)})").imports(JAVA_UTIL_BASE_64).build();
+
+    private static final JavaTemplate decodeFromUrlSafeString = JavaTemplate.builder("Base64.getUrlDecoder().decode(#{any(String)})").imports(JAVA_UTIL_BASE_64).build();
```
greg-at-moderne, line 64: When I saw this and the code for `visitMethodInvocation` I thought maybe one could put these matcher-template pairs into some kind of a map. And then just apply them generically in `visitMethodInvocation`. But not sure if this would be more readable though. So just a thought.

greg-at-moderne (approved):
Thanks for this nice contribution!

src/main/java/org/openrewrite/java/spring/framework/MigrateBase64Utils.java
```diff
@@ -0,0 +1,81 @@
...
+            SPRING_TO_JDK_MAPPINGS.put(new MethodMatcher(SPRING_BASE_64_UTILS + " decodeFromString(String)"), JavaTemplate.builder("Base64.getDecoder().decode(#{any(String)})").imports(JAVA_UTIL_BASE_64).build());
+            SPRING_TO_JDK_MAPPINGS.put(new MethodMatcher(SPRING_BASE_64_UTILS + " encodeUrlSafe(byte[])"), JavaTemplate.builder("Base64.getUrlEncoder().encode(#{anyArray(byte)})").imports(JAVA_UTIL_BASE_64).build());
+            SPRING_TO_JDK_MAPPINGS.put(new MethodMatcher(SPRING_BASE_64_UTILS + " decodeUrlSafe(byte[])"), JavaTemplate.builder("Base64.getUrlDecoder().decode(#{anyArray(byte)})").imports(JAVA_UTIL_BASE_64).build());
+            SPRING_TO_JDK_MAPPINGS.put(new MethodMatcher(SPRING_BASE_64_UTILS + " encodeToUrlSafeString(byte[])"), JavaTemplate.builder("Base64.getUrlEncoder().encodeToString(#{anyArray(byte)})").imports(JAVA_UTIL_BASE_64).build());
+            SPRING_TO_JDK_MAPPINGS.put(new MethodMatcher(SPRING_BASE_64_UTILS + " decodeFromUrlSafeString(String)"), JavaTemplate.builder("Base64.getUrlDecoder().decode(#{any(String)})").imports(JAVA_UTIL_BASE_64).build());
```
greg-at-moderne, line 64: Can we make it static or lazy initialised?

The consequence of the current code is that it runs JavaTemplate a few times on every creation of the recipe object. And JavaTemplate is a fairly expensive operation - i.e. it instantiates and runs a parser for Java code under the hood.

src/main/java/org/openrewrite/java/spring/framework/MigrateBase64Utils.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    private static class UpdateDeprecatedMethods extends JavaIsoVisitor<ExecutionContext> {
+
+        private static final String JAVA_UTIL_BASE_64 = "java.util.Base64";
+
+        private static final Map<MethodMatcher, JavaTemplate> SPRING_TO_JDK_MAPPINGS = new HashMap<>();
+
+        static {
+            SPRING_TO_JDK_MAPPINGS.put(new MethodMatcher(SPRING_BASE_64_UTILS + " encode(byte[])"), JavaTemplate.builder("Base64.getEncoder().encode(#{anyArray(byte)})").imports(JAVA_UTIL_BASE_64).build());
```
timtebeek, line 57: @knutwannheden I think we wouldn't want to have statically initialized JavaTemplates here right? As that would mean we'll create those eagerly, even when not used. Whereas the caching we have on templates if they are used ought to be enough I think. Correct?

timtebeek (approved):
Thanks a lot @wapkch ; ties in nicely with #693 to now add this recipe to 6.2 as well. 🙏🏻

PR #690, opened 2025-03-27
Add Recipe to migrate `org.springframework.web.reactive.HandlerResult.setExceptionHandler` method

wapkch:
## What's your motivation?

- Addresses part of https://github.com/openrewrite/rewrite-spring/issues/627

### Checklist


src/main/java/org/openrewrite/java/spring/framework/MigrateHandlerResultSetExceptionHandlerMethod.java
```diff
@@ -0,0 +1,95 @@
...
+
+@NoArgsConstructor
+@AllArgsConstructor
+public class MigrateHandlerResultSetExceptionHandlerMethod extends Recipe {
+
+    private static final String HandlerResult = "org.springframework.web.reactive.HandlerResult";
+
+    private static final MethodMatcher METHOD_MATCHER = new MethodMatcher(HandlerResult + " setExceptionHandler(java.util.function.Function)");
+
+    private String exceptionHandlerFieldName = "exceptionHandler";
```
greg-at-moderne, line 41: I think instead of making the variable name a field (or optional parameter with hard-coded default) it would be better to generate the variable name. There's `VariableNameUtils.generateVariableName()` for that purpose, which is smart enough to avoid name clashes, etc.

src/test/java/org/openrewrite/java/spring/framework/MigrateHandlerResultSetExceptionHandlerMethodTest.java
```diff
@@ -0,0 +1,106 @@
...
+                """
+                    import org.springframework.web.reactive.HandlerResult;
+                    import reactor.core.publisher.Mono;
+
+                    import java.util.function.Function;
+
+                    class MyHandler {
+                        void configureHandler(HandlerResult result) {
+                            Function<Throwable, Mono<HandlerResult>> func = (ex -> Mono.empty());
+                            result.setExceptionHandler((exchange, ex) -> func.apply(ex));
```
greg-at-moderne, line 64: I wonder if we could inline the `func` variable here.
I might be missing some context here in terms of this variable having some significance beyond this line, but I am not sure what it would be as it's a local variable (not a field).

wapkch: Oh, I hadn't realized that.

greg-at-moderne (approved):
Looks good now!
Thanks for the contribution, incl. the fixes.

PR #693, opened 2025-03-30
Add Recipe to migrate `org.springframework.web.reactive.HandlerResult.hasExceptionHandler` to `getExceptionHandler()` method

wapkch:
…andler` method to `getExceptionHandler()`

## What's your motivation?

Addresses part of https://github.com/openrewrite/rewrite-spring/issues/627

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/spring/framework/MigrateHandlerResultHasExceptionHandlerMethod.java
+++ b/src/main/java/org/openrewrite/java/spring/framework/MigrateHandlerResultHasExceptionHandlerMethod.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright 2025 the original author or authors.
+ * <p>
+ * Licensed under the Moderne Source Available License (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://docs.moderne.io/licensing/moderne-source-available-license
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.java.spring.framework;
+
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.Preconditions;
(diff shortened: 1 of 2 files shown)


src/main/java/org/openrewrite/java/spring/framework/MigrateHandlerResultHasExceptionHandlerMethod.java
```diff
@@ -0,0 +1,62 @@
...
+import org.openrewrite.Preconditions;
+import org.openrewrite.Recipe;
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.java.JavaTemplate;
+import org.openrewrite.java.JavaVisitor;
+import org.openrewrite.java.MethodMatcher;
+import org.openrewrite.java.search.UsesType;
+import org.openrewrite.java.tree.J;
+
+public class MigrateHandlerResultHasExceptionHandlerMethod extends Recipe {
```
timtebeek, line 28: Great to see this recipe; should we hook it into any of the larger declarative composite recipe like a Spring Boot upgrade? Otherwise folks would have to run this in isolation, whereas presumably it can be made part of an upgrade to run by default.

wapkch: Yes, this should be part of the [Recipes to upgrade to Spring Framework 6.2](https://github.com/openrewrite/rewrite-spring/issues/627).  Do you think I should file another PR to upgrade to Spring Framework 6.2 and combine these changes?

timtebeek: That would be much appreciated, yes! Should ideally be relatively straightforward looking at the 6.1 migration as an example:
https://github.com/openrewrite/rewrite-spring/blob/main/src/main/resources/META-INF/rewrite/spring-framework-61.yml

Could even be part of this pull request if it's of a similar size as the 6.1 case.

wapkch: I found that the case for 6.2 is much larger than 6.1’s, so I addressed part of it first. I may then file another PR to cover the rest.

timtebeek (approved):
Thanks a lot @wapkch ! Great to have the 6.2 recipes included here as well.

PR #698, opened 2025-04-18
[draft] Remove Spring annotations if they repeating in subclasses

arodionov:
### Checklist


timtebeek:
- blocked by https://github.com/openrewrite/rewrite/pull/5337

timtebeek:
As discussed we'll park this for now; no need to implement a partial solution when we're blocked on a full solution.

PR #705, opened 2025-05-05
Removing SpringExtension.class from  ExtendWith annotation

arodionov:
- added new tests and fix recipe behaviour

### DIFF
--- a/src/main/java/org/openrewrite/java/spring/boot2/UnnecessarySpringExtension.java
+++ b/src/main/java/org/openrewrite/java/spring/boot2/UnnecessarySpringExtension.java
@@ -19,6 +19,7 @@ import org.openrewrite.ExecutionContext;
 import org.openrewrite.Preconditions;
 import org.openrewrite.Recipe;
 import org.openrewrite.TreeVisitor;
+import org.openrewrite.java.AnnotationMatcher;
 import org.openrewrite.java.JavaIsoVisitor;
 import org.openrewrite.java.RemoveAnnotation;
 import org.openrewrite.java.search.FindAnnotations;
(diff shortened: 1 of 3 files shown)


src/main/java/org/openrewrite/java/spring/boot2/UnnecessarySpringExtension.java
```diff
@@ -92,13 +100,41 @@ public J.ClassDeclaration visitClassDeclaration(J.ClassDeclaration classDecl, Ex
...
+                        if (EXTEND_WITH_MATCHER.matches(annotation)) {
+
+                            Expression expression = annotation.getArguments().get(0);
+
+                            if (expression instanceof J.NewArray) {
+                                List<Expression> collected = ((J.NewArray) expression).getInitializer().stream()
+                                        .filter(e -> !isSpringExtension(e))
+                                        .collect(Collectors.toList());
+                                Expression newArgument = ((J.NewArray) expression).withInitializer(collected);
+                                J.Annotation newAnnotation = annotation.withArguments(Collections.singletonList(newArgument));
```
timtebeek, line 125: can we change this to use `ListUtils.map()`, with a mapper that uses `return null` where an element should be dropped?

arodionov: https://github.com/openrewrite/rewrite-spring/pull/705/commits/3c975cb63f2bc7c39ff8012a6acf4bd83be7903f

timtebeek (approved):
Thanks for the help! In reviewing I discovered that we can drop the explicit list of spring boot test annotations by looking for meta annotations instead. That simplifies things quite a bit. 

Also there was a bit of an odd handling of class body, and your addition to conditionally call `super.visitClassDeclaration` to jump into `visitAnnotation`, so figured clear all of that out as well to arrive at a single visit.

timtebeek:
Hmm; the failures we're now seeing are when we upgrade from an older Spring Boot version where those `SpringBootTest` annotations are not yet meta annotated, and the type information is not reflected when we upgrade to a newer Spring Boot version.

timtebeek (approved):

PR #708, opened 2025-05-12
Fixed the ClassCastException in MergeBootstrapYamlWithApplicationYaml recipe

Jenson3210:
## What's changed?
There was a ClassCastException in this recipe when running on platform against some spring projects. 

## What's your motivation?
The existing test was failing when I undo the disabling. Fixed the test, added some more tests + adapted the behaviour to correspond with the tests. 

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/spring/boot2/MergeBootstrapYamlWithApplicationYaml.java
+++ b/src/main/java/org/openrewrite/java/spring/boot2/MergeBootstrapYamlWithApplicationYaml.java
@@ -22,13 +22,14 @@ import org.openrewrite.internal.ListUtils;
 import org.openrewrite.java.spring.ExpandProperties;
 import org.openrewrite.yaml.CoalescePropertiesVisitor;
 import org.openrewrite.yaml.MergeYamlVisitor;
+import org.openrewrite.yaml.YamlParser;
 import org.openrewrite.yaml.search.FindProperty;
 import org.openrewrite.yaml.tree.Yaml;
 
 import java.nio.file.Path;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.List;
+import java.util.Optional;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 public class MergeBootstrapYamlWithApplicationYaml extends ScanningRecipe<MergeBootstrapYamlWithApplicationYaml.Accumulator> {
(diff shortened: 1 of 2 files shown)


src/main/java/org/openrewrite/java/spring/boot2/MergeBootstrapYamlWithApplicationYaml.java
```diff
@@ -58,12 +59,10 @@ public TreeVisitor<?, ExecutionContext> getScanner(Accumulator acc) {
                 }
                 SourceFile source = (SourceFile) tree;
                 Path sourcePath = source.getSourcePath();
-                if (acc.getBootstrapYaml() == null && source instanceof Yaml.Documents &&
-                        PathUtils.matchesGlob(sourcePath, "**/main/resources/bootstrap.yml")) {
-                    acc.setBootstrapYaml((Yaml.Documents) source);
-                } else if (acc.getApplicationYaml() == null &&
-                        PathUtils.matchesGlob(sourcePath, "**/main/resources/application.yml")) {
-                    acc.setApplicationYaml((Yaml.Documents) source);
+                if (acc.getBootstrapYaml() == null && PathUtils.matchesGlob(sourcePath, "**/main/resources/bootstrap.y?ml")) {
```
greg-at-moderne, line 62: Is `y?ml` glob supposed to match `yml`?

Jenson3210: Let me add a test for that... So that I don't do the same as the previous recipe: do it for one and not for the other

Jenson3210: changed to `*`

greg-at-moderne (approved):

timtebeek:
Did we look at the underlying issue why these yaml files now fail to parse and are turned into Quarks?
Could be good to run FindParseFailures on the same repositories and log findings as an issue.

Jenson3210:
> Did we look at the underlying issue why these yaml files now fail to parse and are turned into Quarks? Could be good to run FindParseFailures on the same repositories and log findings as an issue.

I tried doing this but there are no Quarks (`org.openrewrite.FindQuarks`) and FindParseFailures. 
I ran them against netflix and spring organization. 
According to both these recipes, no quarks are present in these organizations. However, the failure report reported only this one error.
Also no commits happened on this repo that I investigated the issue for, so decided to just write more tests and make sure that we do not try to handle a Quark in that recipe.

PR #711, opened 2025-05-16
`MigrateJobBuilderFactory` requires Spring Batch 5.1.+ specifically

timtebeek:
## What's changed?
Update TypeTable & JavaTemplate to match required version.

## What's your motivation?
When using 5.2.+ we see this failure, so then better to be explicit and match the `parserClasspath` entry.

I'd been curious to know if we see any other failures when updating the type table; this ought to flush out any such issues.

@bmuschko


src/main/java/org/openrewrite/java/spring/batch/MigrateJobBuilderFactory.java
```diff
@@ -64,7 +64,7 @@ public J visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx)
                             .builder("new JobBuilder(#{any(java.lang.String)}, jobRepository)")
                             .contextSensitive()
                             .javaParser(JavaParser.fromJavaVersion()
-                                    .classpathFromResources(ctx, "spring-batch-core-5.+"))
```
bmuschko, line 67: Can you explain, why the notation `spring-batch-core-5.+` does not match `spring-batch-core:5.1.+` in the `parserClasspath` definition? If the syntax doesn't cover other minor versions, I think upgrading to `spring-batch-core:5.2.+` or higher in the future (and not changing the implementation again) may become pretty brittle.

timtebeek: Oh it does match; but if we were to use 5.2 in parserClasspath this test breaks, so then I'd think it better to make it clear here that we need specifically 5.1 here.

bmuschko: OK. Sounds like we just forgot to update the type table when we upgraded the dependency version?

timtebeek: More the other way around: I should have changed the usage here when I pinned the `parserClasspath` entry in
- https://github.com/openrewrite/rewrite-spring/pull/676/files
- https://github.com/openrewrite/rewrite-spring/pull/676

timtebeek: With that context out of the way I think we're good to merge here; we're not changing anything to spring-batch usage just yet in the classpath.tsv.zip, just a refresh of unpinned dependencies, and making explicit the pinned version range here.

PR #715, opened 2025-05-23
Add some repository-scoped preconditions to speed up recipes by reduc…

sambsnyd:
…ing unnecessary execution of old/redundant upgrade steps.

### Checklist

### DIFF
--- a/src/main/resources/META-INF/rewrite/spring-boot-20.yml
+++ b/src/main/resources/META-INF/rewrite/spring-boot-20.yml
@@ -28,6 +28,11 @@ description: >-
 tags:
   - spring
   - boot
+preconditions:
+  - org.openrewrite.java.dependencies.search.RepositoryHasDependency:
+      groupIdPattern: org.springframework
+      artifactIdPattern: "*"
+      version: 1-5.1
 recipeList:
   # Upgrade 2.0.x from 1.x
   - org.openrewrite.java.dependencies.UpgradeDependencyVersion:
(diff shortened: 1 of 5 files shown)


jkschneider (approved):

nmck257:
Last time I did profiling, the most expensive (and redundant) part of these recipes was the dependency updates (especially glob string comparisons to check each project dep against group/artifact patterns).

Meanwhile, other included sub recipes can be desirable to re-run, for reasons like below:
- some migrations, if missed in initial upgrade, may not lead to an immediate issue, but would still be beneficial to catch later
- deprecations and later removals (where the same migration might be relevant at multiple versions)
- new sub recipes being added in the history which may apply to "modern-version" projects

For a very visible example, I still regularly see projects working on spring boot 3 upgrades which still use junit 4, which is part of the 2.4 (iirc) recipe.

Some alternative ideas:
- slice up the recipe declarations so that these new preconditions apply only to spring dep mgmt recipes
- in-memory caching for glob matcher results (and then re-profile)

jkschneider:
> some migrations, if missed in initial upgrade, may not lead to an immediate issue, but would still be beneficial to catch later

This is why the preconditions introduce a version range which still has some slack in it.

nmck257:
The version range slack helps; if I'm reading right, each recipe would expect to include its N-1 recipe in a one-minor-version upgrade scenario (or N-2 if the range syntax is inclusive). That probably catches most deprecate-remove scenarios, but perhaps not all.

But my two cents is still that the deep history is a strength, and I *expect* (but don't have stats) that avoiding just the repeated "update spring GAVs" part of each recipe would yield most of the same perf benefits while preserving the recipes' diligence.

sambsnyd:
Increasing the maven metadata cache expiration time in the agent has taken 10m timeouts down to 30s successes, so this doesn't feel necessary anymore

nmck257:
Excellent news; thanks for being open to other avenues here :)

Issue #718, opened 2025-06-01
MigrateWebMvcTagsToObservationConvention Recipe crashes


123Haynes:
## What version of OpenRewrite are you using?
I am using

- OpenRewrite v8.54.0
- Maven/Gradle plugin v6.10.0
- rewrite-spring v6.8.2

## How are you running OpenRewrite?

I am using the Maven plugin, and my project is a single module project.
```xml
            <plugin>
                <groupId>org.openrewrite.maven</groupId>
                <artifactId>rewrite-maven-plugin</artifactId>
                <version>6.10.0</version>
                <configuration>
                    <exportDatatables>true</exportDatatables>
                    <activeRecipes>
                        <recipe>org.openrewrite.java.spring.boot3.MigrateWebMvcTagsToObservationConvention</recipe>
                    </activeRecipes>
                </configuration>
                <dependencies>
                    <dependency>
                        <groupId>org.openrewrite.recipe</groupId>
                        <artifactId>rewrite-spring</artifactId>
                        <version>6.8.2</version>
                    </dependency>
                </dependencies>
            </plugin>
```

## What is the smallest, simplest way to reproduce the problem?

```java
import javax.servlet.http.HttpServletRequest;
import javax.servlet.http.HttpServletResponse;

import org.springframework.boot.actuate.metrics.web.servlet.WebMvcTagsProvider;

import io.micrometer.core.instrument.Tag;
import io.micrometer.core.instrument.Tags;
import lombok.AllArgsConstructor;

@AllArgsConstructor
public class WebMvcExample implements WebMvcTagsProvider {

    /**
     * Instance of {@link WebMvcTagsProvider}.
     */
    private final WebMvcTagsProvider webMvcTagsProvider;

    @Override
    public Iterable<Tag> getTags(final HttpServletRequest request, final HttpServletResponse response, final Object handler, final Throwable exception) {
        return Tags.of(webMvcTagsProvider.getTags(request, response, handler, exception)).and("ABC", "ABC");
    }

    @Override
    public Iterable<Tag> getLongRequestTags(final HttpServletRequest request, final Object handler) {
        return Tags.of(webMvcTagsProvider.getLongRequestTags(request, handler)).and("ABC", "ABC");
    }
}
```

## What did you expect to see?

I expect to either see the working migration or that nothing changes and openrewrite continues to the next recipe.

## What did you see instead?

Openrewrite logs an exception and ends without applying any migrations.

[...]

PR #720, opened 2025-06-02
Added a precondition to MigrateWebMvcTagsToObservationConvention

123Haynes:
## What's changed?
Currently the recipe only supports classes that don't declare the getLongRequestTags method.
The recipe will now safely back off instead of crashing if that method exists.

## What's your motivation?
For now, prevent the recipe from crashing if it can't handle the existing code.
Partly adresses https://github.com/openrewrite/rewrite-spring/issues/718

I considered changeing the recipe to also handle `getLongRequestTags` and handle `.and(..)` in the fluent api.  
However I'm still pretty new at writing java recipes and it would probably take me quite a while.  
So for now I've taken the simplest route to at least avoid crashing the process if the recipe can't handle the existing code.

### Checklist


timtebeek (approved):
Thanks for the quick workaround here to back off from making incorrect changes.

PR #727, opened 2025-06-05
Migrate Docket to GroupedOpenApi

JohannisK:
## What's changed?
Added a recipe to migrate Swagger2 Docker beans to GroupedOpenApi beans

## What's your motivation?
Recipe was missing from the more general migration recipe.


src/test/java/org/openrewrite/java/spring/doc/MigrateDocketBeanToGroupedOpenApiBeanTest.java
```diff
@@ -0,0 +1,562 @@
...
+                class ApplicationConfiguration {
+                }
+                """
+            )
+          )
+        );
+    }
+
+    @Test
+    void rewriteDocketWithApplicationProperties() {
```
jevanlingen: The names of the test `rewriteDocketWithApplicationProperties` and `rewriteDocketWithoutProperties` do show what's going on. Can you add a cause to tell the reader why this happens? So something like:

- rewriteDocketWithApplicationProperties => rewriteSingleDocketToApplicationProperties

That way you show not only what happens, but also why it happens.

src/main/java/org/openrewrite/java/spring/doc/MigrateDocketBeanToGroupedOpenApiBean.java
```diff
@@ -0,0 +1,297 @@
...
+        public boolean hasProperties = false;
+        public List<DocketDefinition> docketDefinitions = new ArrayList<>();
+
+        @Value
+        static class DocketDefinition {
+            Expression groupName;
+            Expression paths;
+            Expression apis;
+
+            static class Builder {
```
jevanlingen: Is it not possible to use the Lombok @Builder here?

JohannisK: I don't think si, because I want to keep track weather or not the builder is valid.

src/main/java/org/openrewrite/java/spring/doc/MigrateDocketBeanToGroupedOpenApiBean.java
```diff
@@ -0,0 +1,298 @@
...
+                        }
+                        return result;
+                    }
+                }
+                return super.visit(tree, executionContext, parent);
+            }
+        });
+    }
+
+    private static @NotNull TreeVisitor<?, ExecutionContext> isDocketJavaBeanConfiguration() {
```
jevanlingen:
```suggestion
    private static TreeVisitor<?, ExecutionContext> isDocketJavaBeanConfiguration() {
```

And add a @NullMarked-`package-info.java` file.

src/test/java/org/openrewrite/java/spring/doc/MigrateDocketBeanToGroupedOpenApiBeanTest.java
JohannisK: Nice one :) better!

src/main/java/org/openrewrite/java/spring/doc/MigrateDocketBeanToGroupedOpenApiBean.java
JohannisK: Done

src/main/java/org/openrewrite/java/spring/doc/MigrateDocketBeanToGroupedOpenApiBean.java
```diff
@@ -0,0 +1,304 @@
...
+import org.openrewrite.marker.SearchResult;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.Consumer;
+
+import static org.openrewrite.Preconditions.and;
+import static org.openrewrite.Preconditions.or;
+
+public class MigrateDocketBeanToGroupedOpenApiBean extends ScanningRecipe<MigrateDocketBeanToGroupedOpenApiBean.DocketBeanAccumulator> {
```
timtebeek, line 39: Which composite recipe should now include `org.openrewrite.java.spring.doc.MigrateDocketBeanToGroupedOpenApiBean` ?
Because we do not want users to have to run this as a one-off recipe separately.

timtebeek: I'm guessing it's this one, but you'll want to look a little more closely perhaps:
https://github.com/openrewrite/rewrite-spring/blob/1cd94adaa20767263778ae6984123d229dfc357a/src/main/resources/META-INF/rewrite/springdoc.yml#L75-L81

src/main/java/org/openrewrite/java/spring/doc/MigrateDocketBeanToGroupedOpenApiBean.java
```diff
@@ -0,0 +1,304 @@
...
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getScanner(DocketBeanAccumulator acc) {
+        return Preconditions.check(or(isDocketJavaBeanConfiguration(), new IsPossibleSpringConfigFile()), new TreeVisitor<Tree, ExecutionContext>() {
+            @Override
+            public Tree visit(Tree tree, ExecutionContext ctx) {
+                if (tree instanceof J.CompilationUnit) {
+                    return new JavaBeanConfigurationScanner(acc).visit(tree, ctx);
+                } else if (tree != null && new IsPossibleSpringConfigFile().visit(tree, ctx).getMarkers().findFirst(SearchResult.class).isPresent()) {
```
timtebeek: `SearchResult` is not very specific; would cursor messaging be a better option here instead?

src/main/java/org/openrewrite/java/spring/doc/MigrateDocketBeanToGroupedOpenApiBean.java
```diff
@@ -0,0 +1,304 @@
...
+                                maybeRemoveImport("springfox.documentation.builders.PathSelectors");
+                                maybeRemoveImport("springfox.documentation.builders.RequestHandlerSelectors");
+                                maybeRemoveImport("springfox.documentation.spi.DocumentationType");
+                                maybeRemoveImport("springfox.documentation.spring.web.plugins.Docket");
+                                if (canConfigureInProperties(acc, docketDefinition)) {
+                                    return null;
+                                } else {
+                                    String methodTemplate = "@Bean\n" +
+                                            "public GroupedOpenApi " + method.getSimpleName() + "() {\n" +
+                                            "    return GroupedOpenApi.builder()\n";
```
timtebeek: I'm usually somewhat suspicious when JavaTemplates contain spaces for indentation; can we instead use `autoFormat` to ensure we match the indentation style & level the user is expecting?

build.gradle.kts
```diff
@@ -128,6 +131,10 @@ recipeDependencies {
     parserClasspath("io.micrometer:micrometer-commons:1.11.+")
     parserClasspath("io.micrometer:micrometer-core:1.11.+")
     parserClasspath("io.micrometer:micrometer-observation:1.11.+")
+
+    parserClasspath("io.springfox:springfox-core:3.+")
+    parserClasspath("io.springfox:springfox-spring-web:3.+")
+    parserClasspath("io.springfox:springfox-spi:3.+")
```
timtebeek: Only the above two are used to create new LST elements:
https://github.com/openrewrite/rewrite-spring/blob/84f2801a341a3ab88def38c320bb240fd6c62b23/src/main/java/org/openrewrite/java/spring/doc/MigrateDocketBeanToGroupedOpenApiBean.java#L120

In tests you'll want to use regular `testRuntime` dependencies, and `.classpath(...)` instead of `classpathFromResources(ctx, ...)`. No need to inflate our runtime TypeTables with types we only need to compile an test input text block.
```suggestion
```

src/main/java/org/openrewrite/java/spring/doc/MigrateDocketBeanToGroupedOpenApiBean.java
```diff
@@ -0,0 +1,304 @@
...
+        }
+    }
+
+    private static class ArgumentExtractor extends JavaIsoVisitor<Consumer<Expression>> {
+
+        @Override
+        public @Nullable J visit(@Nullable Tree tree, Consumer<Expression> consumer) {
+            return super.visit(tree, consumer);
+        }
+
```
timtebeek: This doesn't really do anything?
```suggestion
```

src/main/java/org/openrewrite/java/spring/doc/MigrateDocketBeanToGroupedOpenApiBean.java
```diff
@@ -0,0 +1,304 @@
...
+        Expression apis;
+
+        static class Builder {
+            @Getter
+            private boolean valid = true;
+
+            private Expression paths = null;
+            private Expression apis = null;
+
+            private Expression groupName = null;
```
timtebeek: You'll want to annotate these fields as `@Nullable` when they can be null.

src/main/java/org/openrewrite/java/spring/doc/MigrateDocketBeanToGroupedOpenApiBean.java
timtebeek: Here and throughout you can reduce the warnings (ideally zero!) by using `visitNonNull` instead of `visit` where applicable.

src/main/java/org/openrewrite/java/spring/doc/MigrateDocketBeanToGroupedOpenApiBean.java
jevanlingen: You need to add it to the `SwaggerToSpringDoc` recipe:
https://github.com/openrewrite/rewrite-spring/blob/1cd94adaa20767263778ae6984123d229dfc357a/src/main/resources/META-INF/rewrite/springdoc.yml#L34-L41

src/main/java/org/openrewrite/java/spring/doc/MigrateDocketBeanToGroupedOpenApiBean.java
```diff
@@ -0,0 +1,304 @@
...
+                        return result;
+                    }
+                }
+                return super.visit(tree, ctx, parent);
+            }
+        });
+    }
+
+    private static boolean canConfigureInProperties(DocketBeanAccumulator acc, DocketDefinition docketDefinition) {
+        return acc.hasProperties && (docketDefinition.groupName == null || docketDefinition.groupName instanceof J.Literal) && (docketDefinition.apis == null || docketDefinition.apis instanceof J.Literal) && (docketDefinition.paths == null || docketDefinition.paths instanceof J.Literal);
```
jevanlingen:
```suggestion
        return acc.hasProperties &&
                (docketDefinition.groupName == null || docketDefinition.groupName instanceof J.Literal) && 
                (docketDefinition.apis == null || docketDefinition.apis instanceof J.Literal) && 
                (docketDefinition.paths == null || docketDefinition.paths instanceof J.Literal);
```

src/main/java/org/openrewrite/java/spring/doc/MigrateDocketBeanToGroupedOpenApiBean.java
```diff
@@ -0,0 +1,304 @@
...
+                                maybeRemoveImport("springfox.documentation.spi.DocumentationType");
+                                maybeRemoveImport("springfox.documentation.spring.web.plugins.Docket");
+                                if (canConfigureInProperties(acc, docketDefinition)) {
+                                    return null;
+                                } else {
+                                    String methodTemplate = "@Bean\n" +
+                                            "public GroupedOpenApi " + method.getSimpleName() + "() {\n" +
+                                            "    return GroupedOpenApi.builder()\n";
+                                    if (docketDefinition.groupName == null) {
+                                        methodTemplate = methodTemplate + "            .group(\"public\")\n";
```
jevanlingen:
```suggestion
                                        methodTemplate += "            .group(\"public\")\n";
```

Also use the `+=` operator for the lines beneath.

src/main/java/org/openrewrite/java/spring/doc/MigrateDocketBeanToGroupedOpenApiBean.java
```diff
@@ -0,0 +1,304 @@
...
+                                            .javaParser(JavaParser.fromJavaVersion().classpathFromResources(ctx, "spring-context", "springdoc-openapi-starter-common"))
+                                            .imports("org.springdoc.core.models.GroupedOpenApi", "org.springframework.context.annotation.Bean").build()
+                                            .apply(getCursor(), method.getCoordinates().replace());
+                                }
+                            }
+                            return md;
+                        }
+                    }.visit(tree, ctx);
+                } else if (new IsPossibleSpringConfigFile().visit(tree, ctx).getMarkers().findFirst(SearchResult.class).isPresent()) {
+                    if (canConfigureInProperties(acc, docketDefinition)) {
```
jevanlingen: this if-statement can be merged with the else-if on the line above.

src/main/java/org/openrewrite/java/spring/doc/MigrateDocketBeanToGroupedOpenApiBean.java
```diff
@@ -0,0 +1,304 @@
...
+                            ARGUMENT_EXTRACTOR.visit(mi.getArguments(), builder::addPath);
+                        } else if (mi.getSelect() != null &&
+                                APISELECTORBUILDER_TYPEMATCHER.matches(mi.getSelect().getType()) &&
+                                "apis".equals(mi.getSimpleName())) {
+                            ARGUMENT_EXTRACTOR.visit(mi.getArguments(), builder::addApi);
+                        } else if (mi.getSelect() != null &&
+                                DOCKET_TYPEMATCHER.matches(mi.getSelect().getType()) &&
+                                "groupName".equals(mi.getSimpleName())) {
+                            builder.setGroup(mi.getArguments().get(0));
+                        }
```
jevanlingen:
```suggestion
                        if (!builder.isValid() || mi.getSelect() == null) {
                            return mi;
                        }
                        if (APISELECTORBUILDER_TYPEMATCHER.matches(mi.getSelect().getType()) && "paths".equals(mi.getSimpleName())) {
                            ARGUMENT_EXTRACTOR.visit(mi.getArguments(), builder::addPath);
                        } else if (APISELECTORBUILDER_TYPEMATCHER.matches(mi.getSelect().getType()) && "apis".equals(mi.getSimpleName())) {
                            ARGUMENT_EXTRACTOR.visit(mi.getArguments(), builder::addApi);
                        } else if (DOCKET_TYPEMATCHER.matches(mi.getSelect().getType()) && "groupName".equals(mi.getSimpleName())) {
                            builder.setGroup(mi.getArguments().get(0));
                        }
```

src/main/java/org/openrewrite/java/spring/doc/MigrateDocketBeanToGroupedOpenApiBean.java
```diff
@@ -0,0 +1,304 @@
...
+
+                    @Override
+                    public J.NewClass visitNewClass(J.NewClass newClass, DocketDefinition.Builder builder) {
+                        J.NewClass nc = super.visitNewClass(newClass, builder);
+                        if (!builder.isValid()) {
+                            return nc;
+                        }
+                        if (!DOCKET_TYPEMATCHER.matches(newClass.getType()) ||
+                                !DOCUMENTATIONTYPE_TYPEMATCHER.matches(newClass.getArguments().get(0).getType()) ||
+                                !"SWAGGER_2".equals(((J.FieldAccess) newClass.getArguments().get(0)).getName().getSimpleName())) {
```
jevanlingen:
```suggestion
                        if (builder.isValid() &&
                                (!DOCKET_TYPEMATCHER.matches(newClass.getType()) ||
                                !DOCUMENTATIONTYPE_TYPEMATCHER.matches(newClass.getArguments().get(0).getType()) ||
                                !"SWAGGER_2".equals(((J.FieldAccess) newClass.getArguments().get(0)).getName().getSimpleName()))) {
```

src/main/java/org/openrewrite/java/spring/doc/MigrateDocketBeanToGroupedOpenApiBean.java
```diff
@@ -0,0 +1,304 @@
...
+                paths = expression;
+            }
+
+            public void setGroup(Expression expression) {
+                groupName = expression;
+            }
+
+            public void addApi(Expression expression) {
+                apis = expression;
+            }
```
jevanlingen: Remove this and just use normal setters (with @Setter).

src/main/java/org/openrewrite/java/spring/doc/MigrateDocketBeanToGroupedOpenApiBean.java
```diff
@@ -0,0 +1,304 @@
...
+                                            .imports("org.springdoc.core.models.GroupedOpenApi", "org.springframework.context.annotation.Bean").build()
+                                            .apply(getCursor(), method.getCoordinates().replace());
+                                }
+                            }
+                            return md;
+                        }
+                    }.visit(tree, ctx);
+                } else if (new IsPossibleSpringConfigFile().visit(tree, ctx).getMarkers().findFirst(SearchResult.class).isPresent()) {
+                    if (canConfigureInProperties(acc, docketDefinition)) {
+                        Tree result = new AddSpringProperty("springdoc.api-docs.path", "/v3/api-docs", null, null).getVisitor().visit(tree, ctx);
```
jevanlingen: 💡 As you use the `new AddSpringProperty` that often, you might want to create a utility method for it.

src/main/java/org/openrewrite/java/spring/doc/MigrateDocketBeanToGroupedOpenApiBean.java
JohannisK: Changed this to a more explicit check

src/main/java/org/openrewrite/java/spring/doc/MigrateDocketBeanToGroupedOpenApiBean.java
JohannisK: Removed the spaces, works even without auto format.

build.gradle.kts
JohannisK: Thnx!

src/main/java/org/openrewrite/java/spring/doc/MigrateDocketBeanToGroupedOpenApiBean.java
JohannisK: Done

src/main/java/org/openrewrite/java/spring/doc/MigrateDocketBeanToGroupedOpenApiBean.java
JohannisK: There is only one left, but I'm not sure how to fix that one, because I need to remove a method with that one.

src/main/java/org/openrewrite/java/spring/doc/MigrateDocketBeanToGroupedOpenApiBean.java
JohannisK: I changed this to a StringBuilder

src/main/java/org/openrewrite/java/spring/doc/MigrateDocketBeanToGroupedOpenApiBean.java
JohannisK: thnx!

src/main/java/org/openrewrite/java/spring/doc/MigrateDocketBeanToGroupedOpenApiBean.java
JohannisK: Done

build.gradle.kts
```diff
@@ -116,7 +116,7 @@ recipeDependencies {
 //    parserClasspath("org.springframework.cloud:spring-cloud-sleuth-instrumentation:3.1.+")
 //    parserClasspath("org.springframework.cloud:spring-cloud-sleuth-brave:3.1.+")
 
-    parserClasspath("org.springframework.plugin:spring-plugin-core:2.+")
+    //parserClasspath("org.springframework.plugin:spring-plugin-core:2.+")
```
jevanlingen: Don't think this was your intention? Please revert this.

jevanlingen (approved):

src/main/java/org/openrewrite/java/spring/doc/MigrateDocketBeanToGroupedOpenApiBean.java
```diff
@@ -0,0 +1,319 @@
...
+                                }
+                                if (docketDefinition.paths == null && docketDefinition.apis == null) {
+                                    methodTemplateBuilder.append(".pathsToMatch(\"/**\")\n");
+                                } else {
+                                    if (docketDefinition.paths != null) {
+                                        String paths = docketDefinition.paths instanceof J.Literal ? "\"" + docketDefinition.paths + "\"" : docketDefinition.paths.toString();
+                                        methodTemplateBuilder.append(".pathsToMatch(").append(paths).append(")\n");
+                                    }
+                                    if (docketDefinition.apis != null) {
+                                        String apis = docketDefinition.apis instanceof J.Literal ? "\"" + docketDefinition.apis + "\"" : docketDefinition.apis.toString();
```
timtebeek: Try not to call `toString()` on LST elements when you want to pass them into a new JavaTemplate, as that causes the type information to be lost. Instead, add a placeholder like `#{any()}` and then pass in the original LST element as an argument when you `apply` the JavaTemplate.

src/main/java/org/openrewrite/java/spring/doc/MigrateDocketBeanToGroupedOpenApiBean.java
```diff
@@ -0,0 +1,319 @@
...
+                                        methodTemplateBuilder.append(".packagesToScan(").append(apis).append(")\n");
+                                    }
+                                }
+                                methodTemplateBuilder.append(".build();\n")
+                                        .append("}");
+
+                                maybeAddImport("org.springdoc.core.models.GroupedOpenApi", false);
+                                maybeAddImport("org.springframework.context.annotation.Bean", false);
+                                return JavaTemplate.builder(methodTemplateBuilder.toString())
+                                        .contextSensitive()
```
timtebeek: Likely when you follow the directives above not to use `toString()` but passing in existing LST elements you can remove the need for `contextSensitive` here, allowing the template to be cached.
```suggestion
```

src/main/java/org/openrewrite/java/spring/doc/MigrateDocketBeanToGroupedOpenApiBean.java
```diff
@@ -0,0 +1,319 @@
...
+                    return result;
+                }
+                return super.visit(tree, ctx, parent);
+            }
+        });
+    }
+
+    private static boolean isApplicationProperties(@Nullable Tree tree) {
+        return (tree instanceof Properties.File && "application.properties".equals(((Properties.File) tree).getSourcePath().getFileName().toString())) ||
+                (tree instanceof Yaml.Documents && ((Yaml.Documents) tree).getSourcePath().getFileName().toString().matches("application\\.ya*ml"));
```
timtebeek, line 116: Would we want to glance at or even edit any [profile specific properties files](https://docs.spring.io/spring-boot/reference/features/external-config.html#features.external-config.files.profile-specific)?

src/main/java/org/openrewrite/java/spring/doc/MigrateDocketBeanToGroupedOpenApiBean.java
```diff
@@ -0,0 +1,320 @@
...
+                                    }
+                                    methodTemplateBuilder.append(".build();\n")
+                                            .append("}");
+
+                                    maybeAddImport("org.springdoc.core.models.GroupedOpenApi", false);
+                                    maybeAddImport("org.springframework.context.annotation.Bean", false);
+                                    return JavaTemplate.builder(methodTemplateBuilder.toString())
+                                            .javaParser(JavaParser.fromJavaVersion().classpathFromResources(ctx, "spring-context", "springdoc-openapi-starter-common"))
+                                            .imports("org.springdoc.core.models.GroupedOpenApi", "org.springframework.context.annotation.Bean").build()
+                                            .apply(getCursor(), method.getCoordinates().replace(), args.toArray(new Object[0]));
```
jevanlingen:
```suggestion
                                            .apply(getCursor(), method.getCoordinates().replace(), args.toArray());
```

The `toArray()` already returns an `Object[]`. The construct like `toArray(new Foo[0])` is only there to keep the generic type correct. In other words:

```java
Object[] x = List.of("A").toArray();
String[] x = List.of("A").toArray(new String[0]);
```

src/main/java/org/openrewrite/java/spring/doc/MigrateDocketBeanToGroupedOpenApiBean.java
```diff
@@ -0,0 +1,338 @@
...
+import static org.openrewrite.Preconditions.or;
+
+public class MigrateDocketBeanToGroupedOpenApiBean extends ScanningRecipe<MigrateDocketBeanToGroupedOpenApiBean.DocketBeanAccumulator> {
+
+    private static final AnnotationMatcher BEAN_ANNOTATIONMATCHER = new AnnotationMatcher("@org.springframework.context.annotation.Bean");
+    private static final TypeMatcher DOCKET_TYPEMATCHER = new TypeMatcher("springfox.documentation.spring.web.plugins.Docket");
+    private static final TypeMatcher DOCUMENTATIONTYPE_TYPEMATCHER = new TypeMatcher("springfox.documentation.spi.DocumentationType");
+    private static final TypeMatcher APISELECTORBUILDER_TYPEMATCHER = new TypeMatcher("springfox.documentation.spring.web.plugins.ApiSelectorBuilder");
+    private static final ArgumentExtractor REQUESTHANDLERSELECTORS_ARGUMENT_EXTRACTOR = new ArgumentExtractor(new TypeMatcher("springfox.documentation.builders.RequestHandlerSelectors"));
+    private static final ArgumentExtractor PATHSELECTOR_ARGUMENT_EXTRACTOR = new ArgumentExtractor(new TypeMatcher("springfox.documentation.builders.PathSelectors"));
```
timtebeek: I see there's both a `PathSelectors.ant` as  well as `PathSelectors.regex`; do we really support both?

src/main/java/org/openrewrite/java/spring/doc/MigrateDocketBeanToGroupedOpenApiBean.java
```diff
@@ -0,0 +1,337 @@
...
+        return properties;
+    }
+
+    private static Tree addSpringProperty(ExecutionContext ctx, Tree properties, String property, String value) {
+        return new AddSpringProperty(property, value, null, null).getVisitor().visitNonNull(properties, ctx);
+    }
+
+    private static TreeVisitor<?, ExecutionContext> isDocketJavaBeanConfiguration() {
+        return and(new UsesType<>("springfox.documentation.spring.web.plugins.Docket", true), new UsesType<>("org.springframework.context.annotation.Bean", true));
+    }
```
jevanlingen:
```suggestion
```

Obsolete method by now.

src/main/java/org/openrewrite/java/spring/doc/MigrateDocketBeanToGroupedOpenApiBean.java
JohannisK: Did look into this, no we don't. Shall I make the adjustments?

timtebeek: It seems like currently we'd treat both the same, and pass those into the singular replacement. That seems unlikely to work in both cases, so best addressed indeed.

timtebeek: Fine to just support one of those two, depending on what the replacement allows for, and otherwise make no changes at all(!)

JohannisK: The regex version does not have an OpenAPI equivalent, so I'll ignore that case. Rewriting regex to ant notation will probably not be very trustworthy

timtebeek (approved):
Great to see this covered!

Issue #231, opened 2022-10-30
Add Java visitors to `ChangeSpringPropertyKey` to change keys referenced in source code.


tkvangorder:
See this issue in spring-boot-migrator: https://github.com/spring-projects-experimental/spring-boot-migrator/issues/471

It should be straightforward to add additional functionality to `ChangeSpringPropertyKey` to rename keys in : 

- [x] `@Value` string literals
- [x] `@ConditionalOnProperty`
- [ ] `@PropertySource`
- [ ] `@TestPropertySource`

knutwannheden:
The recipe now at least supports `@Value`. I suggest we close this issue for now. Supporting `@PropertySource` and `@TestPropertySource` would as I understand it require loading the corresponding files and do substitutions in there. That seems more esoteric and definitely more complicated.

PR #728, opened 2025-06-05
Support `@Value` and `@ConditionalOnProperty` in `ChangeSpringPropertyKey`

knutwannheden:
Also support `except` option and default values (as in `${foo:bar}`), including escaping (e.g. `${foo:\{bar\}}`).

Issues:
 - #231

### DIFF
--- a/src/main/java/org/openrewrite/java/spring/ChangeSpringPropertyKey.java
+++ b/src/main/java/org/openrewrite/java/spring/ChangeSpringPropertyKey.java
@@ -109,7 +109,7 @@ public class ChangeSpringPropertyKey extends Recipe {
     private String exceptRegex() {
         return except == null || except.isEmpty() ?
                 "" :
-                "(?!\\.(?:" + String.join("|", except) + "))";
+                "(?!\\.(?:" + String.join("|", except) + ")\\b)";
     }
 
     private class JavaPropertyKeyVisitor extends JavaIsoVisitor<ExecutionContext> {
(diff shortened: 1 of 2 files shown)


src/main/java/org/openrewrite/java/spring/ChangeSpringPropertyKey.java
```diff
@@ -96,4 +107,62 @@ private String exceptRegex() {
...
+
+    private J.Annotation mapArgs(J.Annotation a) {
+        if (a.getArguments() != null) {
+            a = a.withArguments(ListUtils.map(a.getArguments(), arg -> {
+                if (arg instanceof J.Literal) {
+                    J.Literal literal = (J.Literal) arg;
+                    if (literal.getValue() instanceof String) {
+                        String value = (String) literal.getValue();
+                        if (value.contains(oldPropertyKey)) {
+                            Pattern pattern = Pattern.compile("\\$\\{(" + quote(oldPropertyKey) + "(?:\\.[^.}:]+)*)(((?:\\\\.|[^}])*)\\})");
```
timtebeek: I'd sort of expected to see reuse of the `exceptRegex` here as well; was that a conscious change here?

knutwannheden: Good call!

knutwannheden:
I removed the `annotated()` trait again, as I also wanted to support another annotation.

PR #746, opened 2025-06-23
Shorten MaintainTrailingSlashURLMappings code by using an inline visitor

jevanlingen:
## What's changed?
Nothing, just refactoring

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/spring/boot3/MaintainTrailingSlashURLMappings.java
+++ b/src/main/java/org/openrewrite/java/spring/boot3/MaintainTrailingSlashURLMappings.java
@@ -57,11 +57,18 @@ public class MaintainTrailingSlashURLMappings extends ScanningRecipe<AtomicBoole
     public TreeVisitor<?, ExecutionContext> getScanner(AtomicBoolean acc) {
         return new JavaIsoVisitor<ExecutionContext>() {
             @Override
-            public J.CompilationUnit visitCompilationUnit(J.CompilationUnit cu, ExecutionContext ctx) {
-                if (!acc.get()) {
-                    acc.set(FindWebConfigurer.find(cu));
+            public J.ClassDeclaration visitClassDeclaration(J.ClassDeclaration classDecl, ExecutionContext ctx) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/openrewrite/java/spring/boot3/MaintainTrailingSlashURLMappings.java
```diff
@@ -57,11 +57,18 @@ public AtomicBoolean getInitialValue(ExecutionContext ctx) {
...
-                if (!acc.get()) {
-                    acc.set(FindWebConfigurer.find(cu));
+            public J.ClassDeclaration visitClassDeclaration(J.ClassDeclaration classDecl, ExecutionContext ctx) {
+                if (!acc.get() && classDecl.getImplements() != null) {
+                    for (TypeTree impl : classDecl.getImplements()) {
+                        JavaType.FullyQualified fullyQualified = TypeUtils.asFullyQualified(impl.getType());
+                        if (fullyQualified != null &&
+                                (WEB_MVC_CONFIGURER.equals(fullyQualified.getFullyQualifiedName()) ||
+                                        WEB_FLUX_CONFIGURER.equals(fullyQualified.getFullyQualifiedName()))) {
+                            acc.set(true);
```
Jenson3210, line 67: We could use `break` to 100% match existing behaviour and to save some iterations.
```suggestion
                            acc.set(true);
                            break;
```

jevanlingen: This is unnecessary optimisation (which is why I also remove the extra return statement, check the original code 😉 ). As we do already the `acc.get()` check, there is a possibility of 1 time the code will match the for-loop per repository. Then the class should also have multiple implements to do the other iterations, which is also rare. Thus in the end you would save at much of 10ms or something.

I tend to keep the code as small as possible, unless this impacts performance a lot. In this case that absolutely untrue.

Jenson3210: If it is removed intentionally, no objections

PR #752, opened 2025-07-02
Migrate `@Query` to `@NativeQuery` in Spring Data JPA 3.4

### LINKED ISSUE #749: Migrate `@Query` to `@NativeQuery` in Spring Data JPA 3.4

It could be interesting to migrate `@Query` to `@NativeQuery` conditionaly if argument `nativeQuery = true`.

The annotation `@NativeQuery` has been introduced in Spring Data JPA 3.4, as an alias for `@Query(nativeQuery=true)`. 
See changelog: https://github.com/spring-projects/spring-data-jpa/releases/tag/3.4.0 .

Code using annotation `org.springframework.data.jpa.repository.Query`.

```java
              import org.springframework.data.jpa.repository.Query;

              interface Test {

                  @Query(value = "select * from foo", nativeQuery = true)
                  void customQuery();
              }
```

```java
              import org.springframework.data.jpa.repository.NativeQuery;

              interface Test {

                  @NativeQuery(value = "select * from foo")
                  void customQuery();
              }
```

## Any additional context

It could be included in a composite recipe like _Migrate to Spring Data 3.4_ (non-existing yet), similar to https://docs.openrewrite.org/recipes/java/spring/data/upgradespringdata_2_7 

### About the naming of _Migrate to Spring Data x.y_

However, I think that _Migrate to Spring Data 3.4_ may be confusing.
In my opinion, a better choice would be:
- Migrate to Spring Data JPA 3.4 
- Migrate to Spring Data BOM 2024.1

It's not clear if _Spring Data x.y_ applies to Spring Data JPA, Spring Data Commons or anything else.
For example, Spring Data BOM 2024.1 contains: 
- Spring Data Commons 3.4     
- Spring Data JPA 3.4     
- Spring Data Neo4j 7.4     
- Spring Data for Apache Cassandra 4.4     
- Spring Data MongoDB 4.4     
- Spring Data KeyValue 3.4     
- Spring Data LDAP 3.4     
- Spring Data REST 4.4     
- Spring Data Redis 3.4     
- Spring Data Elasticsearch 5.4     
- Spring Data Couchbase 5.4     
- Spring Data Relational 3.4     

See also: 
- Spring Data 2024.1 goes GA: https://spring.io/blog/2024/11/15/spring-data-2024-1-goes-ga
- Spring Data 2021.2 (containing Spring Data JPA 2.7): https://spring.io/blog/2022/05/13/spring-data-2021-2-and-2022-0-m4-released

Yes, I can provide a PR.

ghusta:
- Fixes #749


src/main/java/org/openrewrite/java/spring/data/MigrateQueryToNativeQuery.java
```diff
@@ -0,0 +1,157 @@
...
+            Object value = literal.getValue();
+
+            if (value instanceof Boolean) {
+                return (Boolean) value;
+            }
+        }
+
+        // If it's not a literal (e.g. a method call, variable, etc.), we can't resolve statically
+        return null;
+    }
```
timtebeek: We have an Annotated.Matcher trait that makes it easier to extract annotation attribute values. Could be a good option here.

src/main/resources/META-INF/rewrite/spring-data-bom-2024.1.yml
```diff
@@ -0,0 +1,25 @@
...
+
+########################################################################################################################
+# Spring Data 2024.1 (https://spring.io/projects/spring-data)
+---
+type: specs.openrewrite.org/v1beta/recipe
+name: org.openrewrite.java.spring.data.UpgradeSpringData_2024_1
+displayName: Migrate to Spring Data 2024.1
+description: Migrate applications to the latest Spring Data 2024.1 release.
+recipeList:
+  - org.openrewrite.java.spring.data.UpgradeSpringDataJPA_3_4
```
timtebeek: I think I'd prefer to see `UpgradeSpringDataJPA_3_4` included with the Spring Boot upgrade recipe, as opposed to this Spring Data upgrade that folks would need to run separately.

src/main/java/org/openrewrite/java/spring/data/MigrateQueryToNativeQuery.java
```diff
@@ -0,0 +1,157 @@
...
+                    public J.Annotation visitAnnotation(J.Annotation annotation, ExecutionContext ctx) {
+                        J.Annotation a = super.visitAnnotation(annotation, ctx);
+                        if (DATA_JPA_QUERY_ANNOTATION_MATCHER.matches(a) && getCursor().getParentOrThrow().getValue() instanceof J.MethodDeclaration) {
+                            Optional<J.Assignment> nativeQueryArg = findNativeQueryArgument(a);
+                            boolean isNativeQuery = nativeQueryArg
+                                    .map(assignment -> extractBooleanValue(assignment))
+                                    .orElse(NATIVE_QUERY_DEFAULT_VALUE);
+                            if (!isNativeQuery) {
+                                return a;
+                            } else {
```
timtebeek: Let's unwrap the `else` block here, since there's a `return` in the above `if` block.

src/main/java/org/openrewrite/java/spring/data/MigrateQueryToNativeQuery.java
```diff
@@ -0,0 +1,157 @@
...
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+
+public class MigrateQueryToNativeQuery extends Recipe {
+
+    private static final String DATA_JPA_QUERY_FQN = "org.springframework.data.jpa.repository.Query";
+    private static final String DATA_JPA_NATIVE_QUERY_FQN = "org.springframework.data.jpa.repository.NativeQuery";
+
+    private static final AnnotationMatcher DATA_JPA_QUERY_ANNOTATION_MATCHER =
+            new AnnotationMatcher("@" + DATA_JPA_QUERY_FQN);
```
timtebeek: I think we can already make `nativeQuery = true` part of the annotation matching pattern; that way it's a little easier further down.

src/main/java/org/openrewrite/java/spring/data/MigrateQueryToNativeQuery.java
```diff
@@ -0,0 +1,157 @@
...
+
+    private static final boolean NATIVE_QUERY_DEFAULT_VALUE = false;
+
+    private boolean useShorthandStyleWhenSingleArgumentRemaining = true;
```
timtebeek: I'd remove this field and just make it the default that's not configurable.

timtebeek: I'd remove this field and just make it the default that's not configurable.

src/test/java/org/openrewrite/java/spring/data/MigrateQueryToNativeQueryTest.java
```diff
@@ -0,0 +1,176 @@
...
+
+    @Override
+    public void defaults(RecipeSpec spec) {
+        // To be able to compile the before code block in tests, spring-data-jpa-3.4 is added
+        // to src/test/resources/META-INF/rewrite/classpath
+        // See: https://docs.openrewrite.org/authoring-recipes/recipe-testing#specifying-the-classpath-for-dependencies
+        // See also: https://docs.openrewrite.org/authoring-recipes/multiple-versions
+        spec.recipe(new MigrateQueryToNativeQuery())
+          .parser(JavaParser.fromJavaVersion()
+            .classpathFromResources(new InMemoryExecutionContext(), "spring-data-jpa-3.4.+"));
```
timtebeek: Can we rely on a regular non-packaged-jar `.classpath(...)` dependency here instead? That's preferred for tests where possible, as to not make the git repository needlessly larger.

ghusta: Yes, it's also my opinion, it's not a good long-term solution... 
I'll try to find a better one with the Gradle build.

timtebeek: Thanks! I see you've switched to using the type table. We tend to use that only when necessary for JavaTemplate, but I can pick that up if needed. Don't want to cause you rework when I'm reviewing from the train.

timtebeek (approved):
Neat addition! Thanks for getting this started.

Hope you're able to follow along with the changes I made and agree there; I like how they simplify things and ensure we have less to maintain going forward.

timtebeek:
Thanks again for your comments! Took a while longer to circle back, but for the tests we've since reworked those in
 - https://github.com/openrewrite/rewrite-spring/pull/786

I've also updated the recipe display names where appropriate just now to avoid the confusion that you're rightfully pointed out:
 https://github.com/openrewrite/rewrite-spring/commit/5b623ca4f500d022c17c4f42bd7b48d1b386fc30

Hope that resolves everything related to this for now.

PR #757, opened 2025-07-24
`Could not parse as Java` in `UseNewSecurityMatchers`

Jenson3210:
Now here's a weird issue that I am investigating. 
Created a Draft PR to share status and allow others to check the issue also. 

Reproduce: Run the `canBeUpgradedIfAuthenticationManagerBuilderConfigurationPresent` test in `UpgradeSpringSecurityTest` 
Output of this run will be a `java.lang.IllegalArgumentException: Could not parse as Java:`

This is caused by the `JavaTemplateParser` line 196 `compileTemplate(stub)` throwing an error. 
Going into this method is what is triggering weird behaviour. 
When I put a breakpoint after building the `jp` `Parser` (line 273) I can see that only the first invocation of 
`jp.reset().parse(ctx, stub).findFirst()` gives back a J.ParseError. Any subsequent call makes the same call return a `J.CompilationUnit` after which the test is successfully updating the code as expected. 
So By doing `jp.reset().parse(ctx, stub)).findFirst()` once in my debug session before the code does it, I can impact the test to become green. By default the test is failing. 

> I also see a strange stub but verified this is NOT the root cause of this issue by removing the extra template_stops from the stub during debugging. Still wonder if we also need to report/fix that one though.

### DIFF
--- a/src/testWithSpringSecurity_5_8/java/org/openrewrite/spring/security5/UpgradeSpringSecurityTest.java
+++ b/src/testWithSpringSecurity_5_8/java/org/openrewrite/spring/security5/UpgradeSpringSecurityTest.java
@@ -1,0 +1,111 @@
+/*
+ * Copyright 2024 the original author or authors.
+ * <p>
+ * Licensed under the Moderne Source Available License (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * https://docs.moderne.io/licensing/moderne-source-available-license
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.spring.security5;
+
+import org.junit.jupiter.api.Test;
+import org.openrewrite.DocumentExample;
+import org.openrewrite.java.JavaParser;
+import org.openrewrite.test.RecipeSpec;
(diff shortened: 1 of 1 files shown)


Jenson3210:
Fixed [upstream](https://github.com/openrewrite/rewrite/pull/5832) for now with a retry. 
@timtebeek do we want to have this test here as we could not have a test upstream that did not need a bunch of classpath imports.

timtebeek:
Seems fine indeed to have a test verifying the behavior here, also to guard against regressions if someone were to make adjustments upstream. Can we merge this test into an already existing test class and add an `@Issue` that links here?

src/testWithSpringSecurity_5_8/java/org/openrewrite/spring/security5/UpgradeSpringSecurityTest.java
```diff
@@ -0,0 +1,111 @@
...
+class UpgradeSpringSecurityTest implements RewriteTest {
+    @Override
+    public void defaults(RecipeSpec spec) {
+        spec.parser(JavaParser.fromJavaVersion().classpath(
+                "spring-core-5.3.+", "spring-context-5.3.+", "spring-beans-5.3.+", "spring-web-5.3.+", "spring-security-web-5.8.+", "spring-security-config-5.8.+", "spring-security-core-5.8.+", "tomcat-embed"))
+          .recipeFromResources("org.openrewrite.java.spring.security5.UpgradeSpringSecurity_5_8");
+    }
+
+    @Test
+    @Issue("https://github.com/openrewrite/rewrite-spring/pull/757")
```
timtebeek:
```suggestion
    @Issue("https://github.com/openrewrite/rewrite-spring/pull/757")
    @Test
```

timtebeek (approved):

Issue #763, opened 2025-08-14
Recipes to upgrade to Spring Boot 3.5


wapkch:
https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-3.5-Release-Notes

Here's a list of tasks related to this migration

- [ ] Supported values for .enabled properties must now be either true or false.
- [x] Spring Boot property migration recipes. https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-3.5-Configuration-Changelog
  - https://github.com/openrewrite/rewrite-spring/pull/765
- [x] Pushing metrics to a [Prometheus Pushgateway](https://github.com/prometheus/pushgateway) now requires `io.prometheus:prometheus-metrics-exporter-pushgateway` instead of `io.prometheus:simpleclient_pushgateway`.
- [x] New Starter names and new property prefixes have been created for Spring Cloud Gateway to replace the old ones. https://github.com/spring-cloud/spring-cloud-release/releases/tag/v2025.0.0

timtebeek:
Thanks for your continued involvement here @wapkch !

sruffatti:
Please review @timtebeek @wapkch 

https://github.com/openrewrite/rewrite-spring/pull/765

PR #765, opened 2025-08-20
spring boot 3.4 and 3.5 properties

sruffatti:
## What's changed?
Add property changes for Spring Boot 3.4.x and 3.5.x

## What's your motivation?
Contributing to issue #763 

### Checklist


src/main/resources/META-INF/rewrite/spring-boot-34-properties.yml
```diff
@@ -0,0 +1,127 @@
...
+      newPropertyKey: management.endpoint.shutdown.access
+  - org.openrewrite.java.spring.ChangeSpringPropertyKey:
+      oldPropertyKey: management.endpoint.startup.enabled
+      newPropertyKey: management.endpoint.startup.access
+  - org.openrewrite.java.spring.ChangeSpringPropertyKey:
+      oldPropertyKey: management.endpoint.threaddump.enabled
+      newPropertyKey: management.endpoint.threaddump.access
+  - org.openrewrite.java.spring.ChangeSpringPropertyKey:
+      oldPropertyKey: management.endpoints.enabled-by-default
+      newPropertyKey: management.endpoints.access.default
```
timtebeek: So this would break, because the original has a `true/false` argument, whereas the replacement takes different values. 

You'd want to instead move from `false` to `none` and `true` to `read-only`, but that's not expressible with `ChangeSpringPropertyKey`. 

Instead for now I'd propose to [do no harm](https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices#do-no-harm) and remove these incomplete migrations.
```suggestion
```

src/main/resources/META-INF/rewrite/spring-boot-35.yml
```diff
@@ -0,0 +1,29 @@
...
+displayName: Migrate to Spring Boot 3.5
+description: >-
+  Migrate applications to the latest Spring Boot 3.5 release. This recipe will modify an application's build files,
+  make changes to deprecated/preferred APIs.
+tags:
+  - spring
+  - boot
+recipeList:
+  - org.openrewrite.java.spring.boot3.UpgradeSpringBoot_3_4
+  - org.openrewrite.java.spring.boot3.SpringBootProperties_3_5
```
timtebeek: I'd be somewhat hesitant to merge this "Migrate to Spring Boot 3.5" recipe iteration as-is, as in the current form this is likely to lead to confusion / complaints that the recipe isn't yet doing anything other than properties. Can we hold back on this one until there's more to show for that isn't likely to break in a way that gives folks a bad experience?

src/main/resources/META-INF/rewrite/spring-boot-34-properties.yml
sruffatti: Wow, good catch. I apologize for missing that. I used the Gradle plugin to generate this file. I will commit the fix either tonight or tomorrow morning.

src/main/resources/META-INF/rewrite/spring-boot-35.yml
sruffatti: Thats reasonable. What if I remove the `UpgradeSpringBoot_3_5` recipe for now and only commit the `SpringBootProperties_3_5` recipe? I will work on a more robust `UpgradeSpringBoot_3_5` recipe in the mean time.

timtebeek: That would be appreciated, yes. Just looking to avoid a lot of questions, or folks being turned off by seeing partial/broken results, as that reflects on the tech.

src/main/resources/META-INF/rewrite/spring-boot-34-properties.yml
```diff
@@ -24,6 +24,284 @@ tags:
...
+      propertyKey: management.endpoint.threaddump.access
+      newValue: read-only
+      oldValue: true
+  - org.openrewrite.java.spring.ChangeSpringPropertyValue:
+      propertyKey: management.endpoint.threaddump.access
+      newValue: none
+      oldValue: false
+  - org.openrewrite.java.spring.ChangeSpringPropertyKey:
+      oldPropertyKey: management.endpoints.enabled-by-default
+      newPropertyKey: management.endpoints.access.default
```
timtebeek: Would it make sense to branch off these property migrations into a separate Yaml Document ?  (so just below, separated by `---`. That would make them less likely to be overwritten were someone to regenerate the migration properties for newer 3.4 and 3.5 .patch versions.

src/main/resources/META-INF/rewrite/spring-boot-35.yml
timtebeek: Can we pull separate the 3.4 property migrations? Or extract out the 3.5 recipes? That would make it easier to get some of the work you've done here ahead of Wednesday's release.

sruffatti: Yes, removed. I will create another PR for the 3.5 recipes.

src/main/resources/META-INF/rewrite/spring-boot-34-properties.yml
sruffatti: yes, handled. New recipe is called 'org.openrewrite.java.spring.boot3.SpringBootProperties_3_4_EnabledToAccess`

src/main/resources/META-INF/rewrite/spring-boot-35.yml
timtebeek: Perfect, thanks a lot!

timtebeek (approved):
Thanks for the help here @sruffatti ! I've added some tests to guard against regressions, which also helped uncover a case missed before. With that we should be ok to merge this first contribution. 🙏🏻

PR #770, opened 2025-08-25
Spring boot 3.5 and Spring cloud 2025

sruffatti:
## What's changed?
Spring boot 3.5 and Spring cloud 2025

## What's your motivation?
Keep open source alive and current.

@timtebeek 

### Checklist


sruffatti:
@timtebeek  I pushed a change to address your unit test. Thanks for this test! Not sure I would have caught this otherwise. I will write better unit tests going forward.

timtebeek (approved):
Thanks for the help here! Neat to see how much you've covered with just declarative recipes.

timtebeek:
We'll do another release this Wednesday if all goes well that should pick this up. 🙏🏻

PR #771, opened 2025-08-26
Add Functional and self-calls options to `MemberReferenceToMethodInvocation`

Jenson3210:
# Fix MemberReferenceToMethodInvocation for No-Argument Method References

## Summary

This PR enhances the `MemberReferenceToMethodInvocation` visitor to properly handle method references that have no arguments, particularly for instance methods and `this`/`super` references.

## Problem

Previously, the visitor would fail or produce incorrect transformations when converting method references without parameters to lambda expressions. This was particularly problematic for:
- Instance method references without arguments (e.g., `String::getBytes`)
- `this` and `super` method references (e.g., `super::toString`)

### Implementation Changes (`MemberReferenceToMethodInvocation.java`)

1. **Enhanced Lambda Generation Logic**: Added conditional logic to handle three distinct cases:
   - Method references with parameters (existing behavior)
   - Method references without parameters using `this` or `super`
   - Instance method references without parameters

2. **Empty Parameter Handling**: Modified `getLambdaArgNames()` to generate a synthetic parameter when the method has no parameters, ensuring the lambda has the correct receiver object.

3. **Template Generation**: The visitor now generates appropriate lambda expressions based on the method reference type:
   - `(params) -> object.method(params)` for methods with parameters
   - `() -> this.method()` or `() -> super.method()` for this/super references
   - `(object) -> object.method()` for instance methods without parameters

### Test Coverage (`MemberReferenceToMethodInvocationTest.java`)

Added two comprehensive test cases:

1. **`methodReferenceWithoutArguments`**: Tests conversion of instance method references without arguments (e.g., `String::getBytes` → `(String string) -> string.getBytes()`)

2. **`thisMethodReferenceWithoutArguments`**: Tests conversion of `super` method references (e.g., `super::toString` → `() -> super.toString()`)

## Testing

All new test cases pass, demonstrating that:
- Instance method references without arguments are correctly transformed
- `this` and `super` references maintain proper semantics
- Existing functionality for method references with parameters remains intact

### DIFF
--- a/src/test/java/org/openrewrite/java/spring/util/MemberReferenceToMethodInvocationTest.java
+++ b/src/test/java/org/openrewrite/java/spring/util/MemberReferenceToMethodInvocationTest.java
@@ -16,6 +16,8 @@
 package org.openrewrite.java.spring.util;
 
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
 import org.openrewrite.DocumentExample;
 import org.openrewrite.InMemoryExecutionContext;
 import org.openrewrite.java.JavaParser;
@@ -90,8 +92,9 @@ class MemberReferenceToMethodInvocationTest implements RewriteTest {
         );
     }
 
-    @Test
-    void superMethodReferenceWithoutArguments() {
+    @ParameterizedTest
+    @ValueSource(strings = {"this", "super"})
+    void superMethodReferenceWithoutArguments(String qualifier) {
         //language=java
         rewriteRun(
           java(
@@ -101,20 +104,20 @@ class MemberReferenceToMethodInvocationTest implements RewriteTest {
 
               class A {
                   public Supplier<String> getString() {
-                      return super::toString;
+                      return %s::toString;
                   }
               }
-              """,
+              """.formatted(qualifier),
             """
               import java.util.Optional;
               import java.util.function.Supplier;
 
               class A {
                   public Supplier<String> getString() {
-                      return () -> super.toString();
+                      return () -> %s.toString();
                   }
               }
-              """
+              """.formatted(qualifier)
           )
         );
     }


src/main/java/org/openrewrite/java/spring/util/MemberReferenceToMethodInvocation.java
```diff
@@ -37,11 +38,23 @@ public J visitMemberReference(J.MemberReference memberRef, ExecutionContext ctx)
...
+                    mr.getContaining(),
+                    mr.getReference().getSimpleName());
+        } else {
+            templateCode = String.format("(%s) -> %s.%s()",
+                    args.stream().map(Param::toString).collect(joining(", ")),
+                    args.get(0).getName(),
+                    mr.getReference().getSimpleName());
+        }
         return JavaTemplate.builder(templateCode)
                 .contextSensitive()
```
knutwannheden, line 59: Would be interesting to see if we could use `bindType()` here instead:
```suggestion
                .bindType(TypeUtils.toString(mr.getType()))
```

knutwannheden: I see that the template doesn't use `#{any()}` to reference the member reference's `containing` property. At least that would also be a requirement for this to work correctly. Something like `"(%s) -> #{any()}.%s(%s)"`.

src/test/java/org/openrewrite/java/spring/util/MemberReferenceToMethodInvocationTest.java
```diff
@@ -62,4 +62,60 @@ void test(ListenableFuture<String> future) {
...
+                      byte[] result = Optional.of("Test").map((String string) -> string.getBytes()).orElse(null);
+                  }
+              }
+              """
+          )
+        );
+    }
+
+    @Test
+    void thisMethodReferenceWithoutArguments() {
```
timtebeek:
```suggestion
    void superMethodReferenceWithoutArguments() {
```

src/main/java/org/openrewrite/java/spring/util/MemberReferenceToMethodInvocation.java
Jenson3210: This creates failing tests in `MigrateAuditorAwareToOptionalTest` due to missing type information. 

`afterTypeValidationOptions(TypeValidation.none())` fixes these tests, but I think we'd rather keep the type info?

knutwannheden: Indeed. These tests fail because the method reference includes `this` somewhere in its `containing`. So it doesn't look like this will be applicable here.

timtebeek (approved):

PR #772, opened 2025-08-26
Add ScanningRecipe for migrate `Hooks.enableAutomaticContextPropagation()` to Spring property

nurdy-kim:
## What's changed?

I added a new `ScanningRecipe` - `MigrateHooksToReactorContextProperty`. This recipe automatically removes the `Hooks.enableAutomaticContextPropagation()` method call from `main` methods within `@SpringBootApplication` classes. After that it adds the new `spring.reactor.context-propagation=true` property to `application.properties` or `application.yml` files, which provides the same functionality.

I'm particularly concerned about whether the addSpringProperty method is working as expected. I'd appreciate it if you could pay close attention to that part during the review.

@nmck257, @timtebeek

### Checklist


src/main/java/org/openrewrite/java/spring/boot3/MigrateHooksToReactorContextProperty.java
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    @Override
+    public String getDescription() {
+        return "Replace `Hooks.enableAutomaticContextPropagation()` with `spring.reactor.context-propagation=true`.";
+    }
+
+    @Override
+    public AtomicBoolean getInitialValue(ExecutionContext ctx) {
+        return new AtomicBoolean(false);
```
timtebeek: While perhaps not as common in practice, we'd need to correctly handle multi module projects too, which might each have this hook and a configuration properties file, or multiple. You can look at this example for how to handle that with an alternative Accumulator:
https://github.com/openrewrite/rewrite-spring/blob/e7a8b99bc3b956e42ae09a83edd19e5a4bb26ec5/src/main/java/org/openrewrite/java/spring/boot2/search/IntegrationSchedulerPoolRecipe.java#L245-L250

timtebeek (changes requested):
Great start here @nurdy-kim ! I've called out a potential source of issues with the `AtomicBoolean` you'd been using. That's an older pattern we're looking to phase out on new recipes. Could you make the necessary adjustments based on the example given?

nurdy-kim:
Thanks for the feedback! I'll make the necessary adjustments! :)

src/main/java/org/openrewrite/java/spring/boot3/MigrateHooksToReactorContextProperty.java
```diff
@@ -0,0 +1,123 @@
...
+import org.openrewrite.java.spring.AddSpringProperty;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.properties.tree.Properties;
+import org.openrewrite.yaml.tree.Yaml;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+
+import static org.openrewrite.Preconditions.and;
+
+public class MigrateHooksToReactorContextProperty extends ScanningRecipe<AtomicBoolean> {
```
timtebeek: Also might be good to hook this recipe into one of the larger composites, or Spring Boot best practices, such that it's run automatically.

nurdy-kim: This recipe can be applicable from Spring Boot 3.2. I believe it would be a good fit for the Migrate to Spring Boot 3.2 recipe. If not, could you please provide some guidance on this?

timtebeek: That then makes sense, yes thanks!

nurdy-kim:
Hi. Sorry for the delay.
I've applied the discussed changes and added my recipe to 'Migrate to Spring Boot 3.2'.

PR #774, opened 2025-08-27
Add Type information to the generated Lambda of MemberReferenceToMethodInvocation

Jenson3210:
This recipe was generating correct code, but the type attribution was wrong. This was causing recipes like ShortenFullyQualified -> imports to fail to act on these lines. 
By fixing the type attribution we won;t have this issue anymore. 

Other than this, the template of a MethodReference can be simplified to not contain the type in the parameters part.


src/main/java/org/openrewrite/java/spring/util/MemberReferenceToMethodInvocation.java
```diff
@@ -50,14 +54,16 @@ public J visitMemberReference(J.MemberReference memberRef, ExecutionContext ctx)
                     mr.getContaining(),
                     mr.getReference().getSimpleName());
         } else {
-            templateCode = String.format("(%s) -> %s.%s()",
-                    args.stream().map(Param::toString).collect(joining(", ")),
+            templateCode = String.format("%1$s -> %1$s.%2$s()",
```
timtebeek, line 57: Instead of building a string here I briefly tried with passing in a J.Identifier, but that leads to an invalid JavaTemplate string, which I don't propose we fix here and now.
```java
            Param param = args.get(0);
            J.Identifier identifier = new J.Identifier(
                    Tree.randomId(),
                    Space.EMPTY,
                    Markers.EMPTY,
                    emptyList(),
                    param.getName(),
                    param.getType(),
                    null);
            templateCode = String.format("#{any()} -> #{any()}.%s()", mr.getReference().getSimpleName());
            templateArgs.add(identifier);
            templateArgs.add(identifier);
```
```java
import java.util.Optional;
import java.util.List;
import org.springframework.boot.autoconfigure.kafka.KafkaConnectionDetails;
import org.openrewrite.java.internal.template.__M__;
import org.openrewrite.java.internal.template.__P__;
public class KafkaConfig{public List<String> getProducerServers(KafkaConnectionDetails connectionDetails){if(true) {return __M__.any(
/*__TEMPLATE__*/__P__.<org.springframework.boot.autoconfigure.kafka.KafkaConnectionDetails>/*__p0__*/p() -> __P__.<org.springframework.boot.autoconfigure.kafka.KafkaConnectionDetails>/*__p1__*/p().getProducerBootstrapServers()/*__TEMPLATE_STOP__*/
)./*__TEMPLATE_STOP__*/orElse(null);}
return null;
}}
```
Note how `p() ->` is not legal.

PR #778, opened 2025-09-02
Changing `ClientHttpResponse#getRawStatusCode()` to `getStatusCode().value()` in order to retain typing.

steve-aom-elliott:
## What's changed?
Instead of simply changing `getRawStatusCode()` (which returns `int`) to `getStatusCode()` (which returns `HttpStatusCode`), it now also chains `.value()` on the end (which returns `int`) in order no not break existing method invocations.

## What's your motivation?
The method name was being changed, but it was not accounting for the difference in method return types.

### Checklist


timtebeek (approved):
Perfect, thanks!

Issue #779, opened 2025-09-04
Migration recipe for SpringBatch 5.2


crankydillo:
## What problem are you trying to solve?

Need to upgrade to SpringBatch 5.2.+ and need to remove references to `StepBuilderFactory` and `JobBuilderFactory`.  It is fine for this to be tied to existing 'migrate away from these', which handles cases where these were autowired in.  We don't have to solve all cases.

## Describe the situation before applying the recipe

```java
class MyJobConfig {
    @Bean
    Step myStep(StepBuilderFactory stepBuilderFactory, Tasklet myTasklet, StepBuilderFactory anotherFactory) {
        return stepBuilderFactory.get("myStep")
                              .tasklet(myTasklet)
                              .build();
      }
}
```

## Describe the situation after applying the recipe

```java
class MyJobConfig {
    @Bean
    Step myStep(Tasklet myTasklet) {
        return stepBuilderFactory.get("myStep")
                              .tasklet(myTasklet)
                              .build();
      }
}
```

## Have you considered any alternatives or workarounds?

Using `DeleteMethodArgument`, but everyone would have to do that..

PR #780, opened 2025-09-04
[779] Remove `StepBuilderFactory` parameters from @Bean-annotated met…

crankydillo:
…hods

This is removed in SpringBoot 5.2 and the 5.0 recipe is already removing @Autowired fields.

I thought I would have to do something like this for `JobBuilderFactory`, but its recipe is already doing somethign like this.  It's a little unclear why there's more testing in this `StepBuilderFactory` case than there is for `JobBuilderFactory`.

Regardless, this class is removed, and I'm hoping this will solve many cases for us (and others!).

 #779

### Checklist


src/main/java/org/openrewrite/java/spring/batch/MigrateStepBuilderFactory.java
```diff
@@ -123,7 +123,7 @@ private boolean isJobRepositoryParameter(Statement statement) {
                            "org.springframework.batch.core.repository.JobRepository");
         }
 
-        private boolean isJobBuilderFactoryParameter(Statement statement) {
+        private boolean isStepBuilderFactoryParameter(Statement statement) {
```
crankydillo, line 123: I think this is a copy-paste issue, so addressed it.

src/testWithSpringBoot_3_0/java/org/openrewrite/java/spring/batch/MigrateStepBuilderFactoryTest.java
```diff
@@ -84,6 +84,48 @@ Step myStep(Tasklet myTasklet, JobRepository jobRepository) {
...
+              import org.springframework.batch.core.Step;
+              import org.springframework.batch.core.repository.JobRepository;
+              import org.springframework.batch.core.step.builder.StepBuilder;
+              import org.springframework.batch.core.step.tasklet.Tasklet;
+              import org.springframework.context.annotation.Bean;
+
+              class MyJobConfig {
+
+                  @Bean
+                  Step myStep( Tasklet myTasklet, JobRepository jobRepository) {
```
crankydillo: Note the unnecessary whitespace before the first argument.  If easy to fix, I will!

timtebeek: Ideally fixed indeed with a `ListUtils.mapFirst(args, arg -> arg.withPrefix(removedArg.getPrefix())`. Let me know if you can fit that in and otherwise I'll pick it up in review after my camping trip.

timtebeek:
Also you branch looks to be 95 commits behind; would you mind resolving the conflict?
https://github.com/crankydillo/rewrite-spring/compare/779_springbatch52...openrewrite%3Arewrite-spring%3Amain

src/testWithSpringBoot_3_0/java/org/openrewrite/java/spring/batch/MigrateStepBuilderFactoryTest.java
crankydillo: Since I don't capture `removedArg`, I used a slightly different approach.  Please modify it to a better approach (if you have time).

crankydillo: I did try `ListUtils#filter`, which I was hoping might handle the removed arg's whitespace, but it left the space as well.

timtebeek (approved):
Thanks a lot for the fix here @crankydillo !

PR #781, opened 2025-09-09
Stepbuilder chunk issue

crankydillo:
This is a quick reproducer of the problem described in [this discussion](https://github.com/openrewrite/rewrite/discussions/6016).  Hard to believe this is a rewrite-spring issue, but needed to make something quick.  I tried to recreate this in `rewrite` with inlined test classes, but couldn't quickly do that.

## What's your motivation?

I need to finish a recipe that converts `chunk(int)`, which is deprecated, to `chunk(int, PlatformTransactionManager)`.

### Checklist

### DIFF
--- a/.github/workflows/properties.yml
+++ b/.github/workflows/properties.yml
@@ -17,7 +17,7 @@ jobs:
         with:
           distribution: 'temurin'
           java-version: '21'
-      - uses: gradle/actions/setup-gradle@v4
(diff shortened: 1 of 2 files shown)


crankydillo:
I removed the ignore-type-check and our test now passes.  Thanks for doubling-back on this!

timtebeek:
Thanks for checking there as well! Glad to hear it's been resolved

PR #796, opened 2025-10-06
simpleclient_pushgateway to prometheus-metrics-exporter-pushgateway

sruffatti:
### Checklist


timtebeek (approved):
Thanks! Were there no further changes needed here beyond the dependency?

sruffatti:
@timtebeek Its not the *only* change. Looking for your guidance on this one. https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-3.5-Release-Notes#prometheus-pushgateway

timtebeek:
Hmm thanks that indeed would need an additional imperative recipe. Maybe best to make part of this PR such that folks don't see partial results with broken metrics. What kind of guidance would you need?

PR #803, opened 2025-10-15
`SpringBean` Trait for `@Bean` methods

MBoegers:
## What's changed?
enabled the `SpringBean` Trait to properly detect and forward Spring Bean defining methods. Prior it used to find `J.Annotation`s and tried to provide `J.MethodDeclaration`s leading to class cast exceptions.

## What's your motivation?
Simpler implement Spring Bean discovery

## Any additional context
Methods that are annotated with `@Bean` define the bean, not the annotation itself.

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/spring/trait/SpringBean.java
+++ b/src/main/java/org/openrewrite/java/spring/trait/SpringBean.java
@@ -22,6 +22,7 @@ import org.openrewrite.Tree;
 import org.openrewrite.java.AnnotationMatcher;
 import org.openrewrite.java.service.AnnotationService;
 import org.openrewrite.java.trait.Annotated;
+import org.openrewrite.java.trait.Literal;
 import org.openrewrite.java.tree.J;
 import org.openrewrite.trait.SimpleTraitMatcher;
 import org.openrewrite.trait.Trait;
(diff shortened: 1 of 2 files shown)


src/main/java/org/openrewrite/java/spring/trait/SpringBean.java
```diff
@@ -75,10 +77,12 @@ public static class Matcher extends SimpleTraitMatcher<SpringBean> {
                 if (springBean.matches(cursor)) {
                     return new SpringBean(cursor);
                 }
-            } else if (value instanceof J.Annotation) {
+            }
+            if (value instanceof J.MethodDeclaration) {
+                AnnotationService annotationService = new AnnotationService();
                 AnnotationMatcher beanAnnotation = new AnnotationMatcher("@org.springframework.context.annotation.Bean");
```
timtebeek, line 83: Was a bit surprised by the new instance here, as we usually go through `org.openrewrite.java.JavaVisitor#service` for a cached instance, but I guess that's not possible here.

MBoegers: It's not a visitor.. let me check once again but this seems to work as no state at all in there.

MBoegers: we could do:
```java
if(new Annotated.Matcher("org.springframework.context.annotation.Bean").lower(cursor).findFirst().isPresent()){
    return new SpringBean(cursor);
}
```
In general I was surprised to see the matchers created every time and being reused. Do you have an answer to this?

timtebeek: I think that's fine; we can tackle any performance issues should they arise, but I don't expect those given how we're using these for now.

MBoegers: Unclear to me, please choose.
A: merge as is
B: use second trait

timtebeek: Ah would have had a preference for the suggestion you made above for the annotated.matcher trait usage. But no rush.

PR #805, opened 2025-10-21
Convert Spring Fox's `ApiInfoBuilder` into Swagger's `Info`

dsgrieve:
## What's changed?

Adding recipes to handle transformation of programmatic use of Springfox API to equivalent swagger v3 API.

## What's your motivation?

It is a common pattern in Springfox to do programmatic configuration as opposed to using annotations, or some combination of the two. The existing recipes handle the annotations, e.g. @ApiInfo to @Info, but none for the direct use of the API.

The focus is on the most likely use case, which is that of doing programmatic configuration in lieu of using annotations.

## Any additional context

Springfox API for reference: https://github.com/springfox/springfox/blob/master/springfox-core/src/main/java/springfox/documentation
Swagger v3 API for reference: https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-models/src/main/java/io/swagger/v3/oas/models
### Checklist

### DIFF
--- a/src/test/java/org/openrewrite/java/spring/swagger/ApiInfoBuilderToInfoTest.java
+++ b/src/test/java/org/openrewrite/java/spring/swagger/ApiInfoBuilderToInfoTest.java
@@ -23,14 +23,12 @@ import org.openrewrite.test.RecipeSpec;
 import org.openrewrite.test.RewriteTest;
 
 import static org.openrewrite.java.Assertions.java;
-import static org.openrewrite.java.Assertions.mavenProject;
 
 class ApiInfoBuilderToInfoTest implements RewriteTest {
 
     @Override
     public void defaults(RecipeSpec spec) {
         spec
-//          .recipe(new TransformApiInfo())
           .recipeFromResources("org.openrewrite.java.springdoc.ApiInfoBuilderToInfo")
           .parser(JavaParser.fromJavaVersion()
             .classpathFromResources(new InMemoryExecutionContext(), "springfox-core-3.+")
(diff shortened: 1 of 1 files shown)


src/main/java/org/openrewrite/java/spring/swagger/TransformApiInfo.java
```diff
@@ -0,0 +1,141 @@
...
+                                        newContact.getArguments().get(2));
+                    }
+                }
+            }
+
+            // Handle license and licenseUrl together
+            // This is more complex - we need to look at the chain and combine license() and licenseUrl()
+            if (license.matches(m)) {
+                // Check if the next method call in the chain is licenseUrl
+                J.MethodInvocation parent = getCursor().getParentTreeCursor().getValue();
```
steve-aom-elliott: Wouldn't the parent be implying the method call prior to this? Either way, yes, more will have to be done for that considering they could be in any order in theory.

dsgrieve: This is Claude generated code, not vetted.

timtebeek:
Let me know when you're ready for a round of feedback!

src/main/java/org/openrewrite/java/spring/swagger/TransformApiInfo.java
```diff
@@ -0,0 +1,155 @@
...
+                        StringBuilder sb = new StringBuilder(".license(new License()");
+                        if (name != null) {
+                            sb.append(".name(\"").append(name.getArguments().get(0)).append("\")");
+                        }
+                        if (url != null) {
+                            sb.append(".url(\"").append(url.getArguments().get(0)).append("\")");
+                        }
+                        sb.append(')');
+                        return sb.toString();
+                    };
```
sambsnyd: Not much point in a final, named lambda, that's just a method declaration with an extra pointer lookup.
```java
private final String licenseTransformer(J.MethodInvocation name, J.MethodInvocation url) { } 
```
That can be used in all the same ways a lambda can.

src/main/java/org/openrewrite/java/spring/swagger/TransformApiInfo.java
```diff
@@ -0,0 +1,155 @@
...
+        private final MethodMatcher termsOfServiceUrl = new MethodMatcher("springfox.documentation.builders.ApiInfoBuilder termsOfServiceUrl(String)");
+        private final MethodMatcher licenseUrl = new MethodMatcher("springfox.documentation.builders.ApiInfoBuilder licenseUrl(String)");
+        private final MethodMatcher license = new MethodMatcher("springfox.documentation.builders.ApiInfoBuilder license(String)");
+        private final MethodMatcher contact = new MethodMatcher("springfox.documentation.builders.ApiInfoBuilder contact(springfox.documentation.service.Contact)");
+        private final MethodMatcher contactConstructor = new MethodMatcher("springfox.documentation.service.Contact <constructor>(String, String, String)");
+        private final MethodMatcher build = new MethodMatcher("springfox.documentation.builders.ApiInfoBuilder build()");
+
+        // accumulator for AppInfoBuilder methods
+        private static class Accumulator {
+            private final Map<String, Function<J.MethodInvocation, String>> transformations = new HashMap<>();
```
sambsnyd: A map of name -> lambda is a lot like a class with named methods, just less efficient.

src/test/java/org/openrewrite/java/spring/swagger/SpringfoxToSpringdocTransformationTest.java
```diff
@@ -0,0 +1,88 @@
...
+import org.openrewrite.DocumentExample;
+import org.openrewrite.InMemoryExecutionContext;
+import org.openrewrite.java.JavaParser;
+import org.openrewrite.test.RecipeSpec;
+import org.openrewrite.test.RewriteTest;
+
+import static org.openrewrite.java.Assertions.java;
+import static org.openrewrite.java.Assertions.mavenProject;
+
+class SpringfoxToSpringdocTransformationTest implements RewriteTest {
```
sambsnyd: Let's also have a test case where a method returns an incomplete builder which is finalized in another method.

build.gradle.kts
```diff
@@ -92,14 +92,16 @@ recipeDependencies {
...
+    parserClasspath("io.swagger.core.v3:swagger-models:2.+")
 
     testParserClasspath("com.nimbusds:nimbus-jose-jwt:9.13")
     testParserClasspath("io.projectreactor:reactor-core:3.6.3")
     testParserClasspath("io.springfox:springfox-core:3.+")
     testParserClasspath("io.springfox:springfox-spring-web:3.+")
     testParserClasspath("io.springfox:springfox-spi:3.+")
     testParserClasspath("io.springfox:springfox-bean-validators:3.+")
-    testParserClasspath("io.swagger.core.v3:swagger-models:2.+")
+    testParserClasspath("io.swagger.core.v3:swagger-core:2.+")
```
timtebeek: This one isn't needed; it was briefly used in the tests, but the tests only need the entries to compile the inputs, not the outputs.
```suggestion
```

src/main/java/org/openrewrite/java/spring/swagger/TransformApiInfo.java
timtebeek, line 1: This class is now no longer used; I suggest we compare approaches between this class and the new classes I've added. For my part I've tried to isolate the logic into smaller subtasks each of which can be achieved in isolation, and reuse for as much as possible existing recipes to change method names and change types.

Ideally my approach leads to less complexity in the recipes itself, less changes to unaffected lines, and less to maintain going forward.

PR #808, opened 2025-10-24
Replace `ApiKey` with `SecurityScheme` and `AuthorizationScope` with `Scopes`

dsgrieve:
## What's changed?

Beginning of recipe(s) to change from springfox security model to openapi security model. 
This changes use of springfox.documentation.service.ApiKey to io.swagger.v3.oas.models.security.SecurityScheme

## What's your motivation?

Continuation of transitioning programmatic use of springfox APIs to swagger v3. 

## Any additional context

- See also PR #805

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/spring/search/FindConfigurationProperties.java
+++ b/src/main/java/org/openrewrite/java/spring/search/FindConfigurationProperties.java
@@ -21,6 +21,7 @@ import org.openrewrite.ExecutionContext;
 import org.openrewrite.Recipe;
 import org.openrewrite.SourceFile;
 import org.openrewrite.TreeVisitor;
+import org.openrewrite.internal.ListUtils;
 import org.openrewrite.java.JavaIsoVisitor;
 import org.openrewrite.java.spring.table.ConfigurationPropertiesTable;
 import org.openrewrite.java.tree.Expression;
@@ -45,7 +46,7 @@ public class FindConfigurationProperties extends Recipe {
     public String getDescription() {
         //language=markdown
         return "Find all classes annotated with `@ConfigurationProperties` and extract their prefix values. " +
-               "This is useful for discovering all externalized configuration properties in Spring Boot applications.";
+                "This is useful for discovering all externalized configuration properties in Spring Boot applications.";
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


src/main/java/org/openrewrite/java/springdoc/SecurityContextToSecurityScheme.java
```diff
@@ -0,0 +1,125 @@
...
+import org.openrewrite.*;
+import org.openrewrite.java.*;
+import org.openrewrite.java.search.UsesMethod;
+import org.openrewrite.java.tree.Expression;
+import org.openrewrite.java.tree.J;
+
+import java.util.Arrays;
+import java.util.List;
+
+public class SecurityContextToSecurityScheme extends Recipe {
```
timtebeek, line 29: Be sure to reference this recipe from a declarative composite recipe as well; otherwise folks would have to search for and run this explicitly as a one off.

dsgrieve: I was going to ask you about this. After your refactor to pull all the ApiInfoBuilder stuff into one imperative recipe, there was no longer a reference to that recipe in springdoc.yml. So I wasn't sure if that was intentional or oversight.

timtebeek: Ah that must have been an oversight! I had intentionally reused the name of the declarative recipe for the imperative recipe, thinking it would be referenced by that name already. Best to add that reference in this PR as well then; thanks!

Let me know when you're ready for a round of review; from what I've seen it looks like we're close to a review & merge.

dsgrieve: It also feels to me that these should be in the package org.openrewrite.java.springdoc.

timtebeek: feel free to move! just mark the PR as [ready for review] when done such that I know when I can take over & merge. :)

src/main/java/org/openrewrite/java/springdoc/SecurityContextToSecurityScheme.java
```diff
@@ -0,0 +1,125 @@
...
+                                    .imports("io.swagger.v3.oas.models.security.SecurityScheme")
+                                    .javaParser(JavaParser.fromJavaVersion().classpathFromResources(ctx, "swagger-models"))
+                                    .build()
+                                    .apply(getCursor(), newClass.getCoordinates().replace(), newClass.getArguments().get(1));
+                        }
+                        return super.visitNewClass(newClass, ctx);
+                    }
+
+                    private String passAsToSecuritySchemeIn(Expression passAsExpr) {
+                        String passAs = ((J.Literal) passAsExpr).getValueSource();
```
steve-aom-elliott: May wish to handle `J.Identifier` and `J.FieldAccess` as possibilities as well

timtebeek: Here's a recipe that can find a few different types of values and lead them back to a String constant:
https://github.com/openrewrite/rewrite-feature-flags/blob/696ff32237e21b424444bc4c4ca39e654cd65f43/src/main/java/org/openrewrite/featureflags/RemoveBooleanFlag.java#L103-L107

Could be good to adopt the same here.

dsgrieve: @timtebeek Is this what you had in mind? https://github.com/openrewrite/rewrite-spring/blob/015f64ef2170043259ec2a7f743cdf16934f0a4f/src/main/java/org/openrewrite/java/springdoc/SecurityContextToSecurityScheme.java#L74-L87

Issue #692, opened 2025-03-28
Replace @MockBean and @SpyBean recipe incorrectly migrates @MockBeans({@MockBean(class)}


NickGotv:
## What version of OpenRewrite are you using?

I am using
- OpenRewrite v6.3.2
- Maven/Gradle plugin v6.3.2
- rewrite-spring v6.3.0

## How are you running OpenRewrite?

I am using the Maven plugin, and my project is a multiple module project.
```xml
<build>
        <plugins>
            <plugin>
                <groupId>org.openrewrite.maven</groupId>
                <artifactId>rewrite-maven-plugin</artifactId>
                <version>6.3.2</version>
                <configuration>
                    <exportDatatables>true</exportDatatables>
                    <activeRecipes>
                        <recipe>org.openrewrite.java.spring.boot3.ReplaceMockBeanAndSpyBean</recipe>
                    </activeRecipes>
                </configuration>
                <dependencies>
                    <dependency>
                        <groupId>org.openrewrite.recipe</groupId>
                        <artifactId>rewrite-spring</artifactId>
                        <version>6.3.0</version>
                    </dependency>
                </dependencies>
            </plugin>
        </plugins>
    </build>
```

## What is the smallest, simplest way to reproduce the problem?
have any JUNIT test with the following structure
```java
@MockBeans({@MockBean(ClassA.class), @MockBean(ClassB.class)})
class SomeTest {
}
```

## What did you expect to see?

```java
@MockitoBean(types = {ClassA.class, ClassB.class})
class SomeTest {
}
```
or

```java
@MockitoBeans(@MockitoBean(types = ClassA.class), @MockitoBean(types =ClassB.class)})
class SomeTest {
}
```
## What did you see instead?

```java
@MockBeans({@MockitoBean, @MockitoBean})
class SomeTest {
}
```

## What is the full stack trace of any errors you encountered?
none for the test with this issue

## Are you interested in [contributing a fix to OpenRewrite]
not at the moment

timtebeek:
Thanks for the report! Seems indeed like we should ensure we handle annotations nested as annotations arguments correctly.

PR #813, opened 2025-10-28
Add ReplaceMockBeanAndSpyBean to UpgradeSpringBoot_4_0 recipe

mcebanupgrade:
## What's changed?
Adding `org.openrewrite.java.spring.boot3.ReplaceMockBeanAndSpyBean` to `org.openrewrite.java.spring.boot3.UpgradeSpringBoot_3_4`

## What's your motivation?

MockBean and SpyBean annotations have been deprecated since Spring Boot 3.4, hence we should include this migration in the Spring Boot 3.4 upgrade.

Alternatively, this could be added to Spring Boot 4 recipe since these annotations are completely removed in Spring Boot 4.

### Checklist


timtebeek:
Hi thanks! We had had this recipe included in an earlier iteration but after feedback had that removed temporarily:
- https://github.com/openrewrite/rewrite-spring/issues/692

Perhaps it makes sense to add this to the Spring Boot 4 migration, as indeed if the annotations are removed there then there's no sense in keeping them at that point, and folks would have to explore the alternative anyway.

timtebeek:
I haven't yet been able to dive into the limitations, or .boot4 specifics, but as we get more usage and feedback that's likely to pop up again. Good to hear in your case it's already helped you through the migration at least (or wasn't a limiting factor).

PR #821, opened 2025-10-31
Migrate SecurityReference SecurityRequirement

dsgrieve:
## What's changed?

Adding more handling of migrating SecurityReference to SecurityRequirement

## What's your motivation?

Multiple imports are not replaced as part of SpringBoot upgrade. 

## Any additional context

- https://github.com/openrewrite/rewrite-spring/pull/805
- https://github.com/openrewrite/rewrite-spring/pull/808
- https://github.com/openrewrite/rewrite-spring/pull/811

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/springdoc/SecurityContextToSecurityScheme.java
+++ b/src/main/java/org/openrewrite/java/springdoc/SecurityContextToSecurityScheme.java
@@ -194,27 +194,18 @@ public class SecurityContextToSecurityScheme extends Recipe {
 
                     private boolean isPartOfCollection(Cursor cursor) {
                         // Check if this AuthorizationScope is part of a List.of, Arrays.asList, or array initializer
-                        Cursor parent = cursor.getParentTreeCursor();
-                        while (parent.getValue() instanceof J) {
-                            J parentTree = parent.getValue();
-                            if (parentTree instanceof J.MethodInvocation) {
-                                J.MethodInvocation mi = (J.MethodInvocation) parentTree;
-                                if (LIST_OF.matches(mi) || ARRAYS_AS_LIST.matches(mi)) {
-                                    return true;
-                                }
-                            } else if (parentTree instanceof J.NewArray) {
-                                return true;
-                            }
-                            parent = parent.getParentTreeCursor();
-                        }
-                        return false;
+                        J parentTree = cursor.getParentTreeCursor().getValue();
+                        return parentTree instanceof J.NewArray ||
+                                parentTree instanceof Expression &&
+                                        (LIST_OF.matches((Expression) parentTree) ||
+                                                ARRAYS_AS_LIST.matches((Expression) parentTree));
                     }
 
                     private boolean isPartOfSecurityReference(Cursor cursor) {
                         // Check if this array is an argument to a SecurityReference constructor
                         Cursor parent = cursor.getParentTreeCursor();
                         if (parent.getValue() instanceof J.NewClass) {
-                            J.NewClass newClass = (J.NewClass) parent.getValue();
+                            J.NewClass newClass = parent.getValue();
                             return SECURITY_REFERENCE_MATCHER.matches(newClass);
                         }
                         return false;
@@ -252,7 +243,7 @@ public class SecurityContextToSecurityScheme extends Recipe {
                                 .imports("io.swagger.v3.oas.models.security.Scopes")
                                 .javaParser(JavaParser.fromJavaVersion().classpathFromResources(ctx, "swagger-models"))
                                 .build()
-                                .apply(getCursor(), ((Expression) node).getCoordinates().replace(), templateArgs.toArray());
+                                .apply(getCursor(), node.getCoordinates().replace(), templateArgs.toArray());
                     }
 
                     class ScopeInfo {
@@ -318,7 +309,6 @@ public class SecurityContextToSecurityScheme extends Recipe {
                                 return JavaTemplate.builder("new SecurityRequirement().addList(#{any(String)}, #{any()}.keySet().stream().collect(Collectors.toList()))")
                                         .imports("io.swagger.v3.oas.models.security.SecurityRequirement", "java.util.stream.Collectors")
                                         .javaParser(JavaParser.fromJavaVersion().classpathFromResources(ctx, "swagger-models"))
-                                        .contextSensitive()
                                         .build()
                                         .apply(getCursor(), newClass.getCoordinates().replace(), referenceName, scopesArg);
                             }
@@ -361,7 +351,7 @@ public class SecurityContextToSecurityScheme extends Recipe {
                                 JavaType scopesType = arrayType.getElemType();
                                 maybeRemoveImport("springfox.documentation.service.AuthorizationScope");
                                 maybeAddImport("springfox.documentation.service.Scopes");
-                                decls = decls.withTypeExpression(
+                                return decls.withTypeExpression(
                                                 TypeTree.build("Scopes")
                                                         .withPrefix(decls.getTypeExpression().getPrefix()))
                                         .withType(scopesType);
@@ -379,7 +369,7 @@ public class SecurityContextToSecurityScheme extends Recipe {
                                 maybeRemoveImport("java.util.List");
                                 maybeRemoveImport("springfox.documentation.service.AuthorizationScope");
                                 maybeAddImport("springfox.documentation.service.Scopes");
-                                decls = decls.withTypeExpression(
+                                return decls.withTypeExpression(
                                                 TypeTree.build("Scopes")
                                                         .withPrefix(decls.getTypeExpression().getPrefix()))
                                         .withType(scopesType);
@@ -394,7 +384,7 @@ public class SecurityContextToSecurityScheme extends Recipe {
             private TreeVisitor<J, ExecutionContext> findArrayParametersOfType(String elementType) {
                 return new JavaIsoVisitor<ExecutionContext>() {
                     @Override
-                    public  J preVisit(J tree, ExecutionContext ctx) {
+                    public J preVisit(J tree, ExecutionContext ctx) {
                         stopAfterPreVisit();
                         if (tree instanceof JavaSourceFile) {
                             for (JavaType.Method method : ((JavaSourceFile) tree).getTypesInUse().getDeclaredMethods()) {


src/main/java/org/openrewrite/java/springdoc/SecurityContextToSecurityScheme.java
```diff
@@ -87,22 +93,294 @@ private String passAsToSecuritySchemeIn(Expression passAsExpr) {
 
             private Tree replaceAuthorizationScope(ExecutionContext ctx, Tree t) {
                 return Preconditions.check(new UsesMethod<>(AUTHORIZATION_SCOPE_MATCHER), new JavaVisitor<ExecutionContext>() {
+                    private final MethodMatcher LIST_OF = new MethodMatcher("java.util.List of(..)");
+                    private final MethodMatcher ARRAYS_AS_LIST = new MethodMatcher("java.util.Arrays asList(..)");
```
steve-aom-elliott: Any reason to not start with it restricted to `java.util.List of(springfox.documentation.service.AuthorizationScope...)` and `java.util.Arrays asList(springfox.documentation.service.AuthorizationScope...)`? The `...` indicating params-style arguments

src/main/java/org/openrewrite/java/springdoc/SecurityContextToSecurityScheme.java
```diff
@@ -87,22 +93,294 @@ private String passAsToSecuritySchemeIn(Expression passAsExpr) {
...
+                            templateBuilder.append("\n.addString(#{any(String)}, #{any(String)})");
+                            templateArgs.add(scope.name);
+                            templateArgs.add(scope.description);
+                        }
+
+                        return JavaTemplate.builder(templateBuilder.toString())
+                                .imports("io.swagger.v3.oas.models.security.Scopes")
+                                .javaParser(JavaParser.fromJavaVersion().classpathFromResources(ctx, "swagger-models"))
+                                .build()
+                                .apply(getCursor(), ((Expression) node).getCoordinates().replace(), templateArgs.toArray());
```
steve-aom-elliott: Would it make sense just to change the return type of `replaceWithChainedScopes` and the first argument's type to `Expression` rather than `J`? You'll always be producing an `Expression` based on your template, and you require an `Expression` anyway for obtaining the coordinates.

src/main/java/org/openrewrite/java/springdoc/SecurityContextToSecurityScheme.java
```diff
@@ -87,22 +93,294 @@ private String passAsToSecuritySchemeIn(Expression passAsExpr) {
...
+
+                    private boolean isPartOfSecurityReference(Cursor cursor) {
+                        // Check if this array is an argument to a SecurityReference constructor
+                        Cursor parent = cursor.getParentTreeCursor();
+                        if (parent.getValue() instanceof J.NewClass) {
+                            J.NewClass newClass = (J.NewClass) parent.getValue();
+                            return SECURITY_REFERENCE_MATCHER.matches(newClass);
+                        }
+                        return false;
+                    }
```
steve-aom-elliott, line 212: Might make sense to move this closer to the only usage / potentially inline it instead.

src/main/java/org/openrewrite/java/springdoc/ArrayTypeMatcher.java
```diff
@@ -0,0 +1,55 @@
...
+import org.jspecify.annotations.Nullable;
+import org.openrewrite.Tree;
+import org.openrewrite.java.JavaIsoVisitor;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.java.tree.JavaSourceFile;
+import org.openrewrite.java.tree.JavaType;
+import org.openrewrite.java.tree.TypeUtils;
+import org.openrewrite.marker.SearchResult;
+
+class ArrayTypeMatcher<P> extends JavaIsoVisitor<P> {
```
steve-aom-elliott: Would probably like to see the name be more specific, given it kind of hides the fact it's focused on method declarations containing an array of a type, not just arrays of a type in general

src/main/java/org/openrewrite/java/springdoc/SecurityContextToSecurityScheme.java
```diff
@@ -87,22 +93,294 @@ private String passAsToSecuritySchemeIn(Expression passAsExpr) {
...
+                                        .imports("io.swagger.v3.oas.models.security.SecurityRequirement", "java.util.stream.Collectors")
+                                        .javaParser(JavaParser.fromJavaVersion().classpathFromResources(ctx, "swagger-models"))
+                                        .build()
+                                        .apply(getCursor(), newClass.getCoordinates().replace(), referenceName, scopesArg);
+                            }
                         }
                         return super.visitNewClass(newClass, ctx);
                     }
 
+                    private List<Expression> extractScopeNames(Expression scopesArg) {
```
steve-aom-elliott: Could probably move this method closer to the usage and just have the parameter be `J.NewArray` since it's the only way you're using it.

src/main/java/org/openrewrite/java/springdoc/SecurityContextToSecurityScheme.java
```diff
@@ -87,22 +93,294 @@ private String passAsToSecuritySchemeIn(Expression passAsExpr) {
...
+                                // Check if type is Scopes[]
+                                if (varType instanceof JavaType.Array && decls.getTypeExpression() != null) {
+                                    JavaType.Array arrayType = (JavaType.Array) varType;
+                                    if (TypeUtils.isOfClassType(arrayType.getElemType(), "io.swagger.v3.oas.models.security.Scopes")) {
+                                        // Replace Scopes[] with Scopes (both syntax and type)
+                                        JavaType scopesType = arrayType.getElemType();
+                                        decls = decls.withTypeExpression(
+                                                        TypeTree.build("Scopes")
+                                                                .withPrefix(decls.getTypeExpression().getPrefix()))
+                                                .withType(scopesType);
```
steve-aom-elliott: Might be worth double checking whether you actually need to use both `.withTypeExpression(..)` and `.withType(..)`, as it appears that `.withType(..)` will already change the `typeExpression` if there's an existing value:

```
@Override
public VariableDeclarations withType(@Nullable JavaType type) {
    return typeExpression == null ? this :
            withTypeExpression(typeExpression.withType(type));
}
```

src/main/java/org/openrewrite/java/springdoc/SecurityContextToSecurityScheme.java
```diff
@@ -87,22 +93,294 @@ private String passAsToSecuritySchemeIn(Expression passAsExpr) {
...
+                                    if (TypeUtils.isOfClassType(paramType, "java.util.List") &&
+                                            paramType.getTypeParameters().size() == 1 &&
+                                            TypeUtils.isOfClassType(paramType.getTypeParameters().get(0), "io.swagger.v3.oas.models.security.Scopes")) {
+                                        // Replace List<Scopes> with Scopes (both syntax and type)
+                                        JavaType scopesType = paramType.getTypeParameters().get(0);
+                                        maybeRemoveImport("java.util.List");
+                                        decls = decls.withTypeExpression(
+                                                        TypeTree.build("Scopes")
+                                                                .withPrefix(decls.getTypeExpression().getPrefix()))
+                                                .withType(scopesType);
```
steve-aom-elliott: Same as above comment on lines 357-360

dsgrieve: We learned that this is necessary.

src/main/java/org/openrewrite/java/springdoc/SecurityContextToSecurityScheme.java
dsgrieve: We learned that this is necessary.

src/main/java/org/openrewrite/java/springdoc/SecurityContextToSecurityScheme.java
dsgrieve: Done.

src/main/java/org/openrewrite/java/springdoc/ArrayTypeMatcher.java
dsgrieve: Renamed to MethodDeclArrayTypeParamMatcher

src/main/java/org/openrewrite/java/springdoc/SecurityContextToSecurityScheme.java
dsgrieve: Done

src/main/java/org/openrewrite/java/springdoc/SecurityContextToSecurityScheme.java
dsgrieve: Done

src/main/java/org/openrewrite/java/springdoc/MethodDeclArrayTypeParamMatcher.java
```diff
@@ -0,0 +1,55 @@
...
+import org.jspecify.annotations.Nullable;
+import org.openrewrite.Tree;
+import org.openrewrite.java.JavaIsoVisitor;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.java.tree.JavaSourceFile;
+import org.openrewrite.java.tree.JavaType;
+import org.openrewrite.java.tree.TypeUtils;
+import org.openrewrite.marker.SearchResult;
+
+class MethodDeclArrayTypeParamMatcher<P> extends JavaIsoVisitor<P> {
```
timtebeek: Did you expect to reuse this more than once? Because otherwise I'd be inclined to inline this just to the one place we use it.

src/test/java/org/openrewrite/java/springdoc/SecurityContextToSecuritySchemeTest.java
```diff
@@ -93,4 +94,179 @@ Scopes authorizationScope() {
...
+                        new AuthorizationScope("write", "Write access"),
+                        new AuthorizationScope("admin", "Admin access")
+                    );
+                }
+            }
+            """,
+            """
+            import io.swagger.v3.oas.models.security.Scopes;
+
+            import java.util.List;
```
steve-aom-elliott: I think in this scenario we should be able to call `maybeRemoveImport(java.util.List)`

src/test/java/org/openrewrite/java/springdoc/SecurityContextToSecuritySchemeTest.java
```diff
@@ -93,4 +94,179 @@ Scopes authorizationScope() {
...
+                        new AuthorizationScope("admin", "Admin access")
+                    );
+                }
+            }
+            """,
+            """
+            import io.swagger.v3.oas.models.security.Scopes;
+
+            import java.util.Arrays;
+            import java.util.List;
```
steve-aom-elliott: Similar to above with removing imports no longer used.

src/test/java/org/openrewrite/java/springdoc/SecurityContextToSecuritySchemeTest.java
```diff
@@ -72,23 +72,198 @@ void AuthorizationScopeToScopes() {
...
+                          new AuthorizationScope("admin", "Admin access")
+                      );
+                  }
+              }
+              """,
+            """
+              import io.swagger.v3.oas.models.security.Scopes;
+
+              import java.util.Arrays;
+              import java.util.List;
```
timtebeek:
```suggestion
```

timtebeek: Same goes above and below.

src/main/java/org/openrewrite/java/springdoc/MethodDeclArrayTypeParamMatcher.java
dsgrieve: This probably won't be used elsewhere. I made it package access for this reason. I didn't inline it is just to reduce the amount of clutter in the SecurityContextToSecurityScheme recipe, which (IMO) is already cluttered enough. But I'm easily swayed on this point and will move it if @timtebeek has a preference.

timtebeek: Then let's inline it, as it avoids confusion about what's reused and what's not. Thanks!

sambsnyd: Note that in Java the rule isn't actually one top-level class per file, it's one _public_ top-level class per file. So you can chuck a package-private class right below the public one.

timtebeek (approved):

PR #828, opened 2025-11-12
Move SpringFox recipes out of SpringDoc package

timtebeek:
Follow up from 
- https://github.com/openrewrite/rewrite-spring/pull/819


src/main/java/org/openrewrite/java/spring/doc/ApiInfoBuilderToInfo.java
```diff
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.openrewrite.java.springdoc;
+package org.openrewrite.java.spring.doc;
```
timtebeek, line 16: We already had this package, and these recipes are for coming from SpringFox.

PR #856, opened 2025-12-02
Replace deprecated starters with their new names

wac84s:
## What's changed?

More updates from https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-4.0-Migration-Guide

# AOP Starter POM

spring-boot-starter-aop, it has been renamed to spring-boot-starter-aspectj

# Deprecated Starters

Several starter POMs have been renamed to provide better alignment with their corresponding module. The older starters remain, but should be considered **deprecated** and will be removed in a future release.

You should update your POM as follows:

| **Deprecated Starter**                              | **Replacement**                                             |
|------------------------------------------------------|-------------------------------------------------------------|
| `spring-boot-starter-oauth2-authorization-server`   | `spring-boot-starter-security-oauth2-authorization-server` |
| `spring-boot-starter-oauth2-client`                 | `spring-boot-starter-security-oauth2-client`               |
| `spring-boot-starter-oauth2-resource-server`        | `spring-boot-starter-security-oauth2-resource-server`      |
| `spring-boot-starter-web`                           | `spring-boot-starter-webmvc`                               |
| `spring-boot-starter-web-services`                  | `spring-boot-starter-webservices`                          |

## What's your motivation?

Follow migration guide

### Checklist

### DIFF
--- a/src/main/resources/META-INF/rewrite/spring-boot-40.yml
+++ b/src/main/resources/META-INF/rewrite/spring-boot-40.yml
@@ -53,6 +53,9 @@ recipeList:
   - org.openrewrite.gradle.plugins.UpgradePluginVersion:
       pluginIdPattern: org.springframework.boot
       newVersion: 4.0.x
+
+  # Replace deprecated starters with their new names
+  # https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-4.0-Migration-Guide#deprecated-starters
   - org.openrewrite.java.dependencies.ChangeDependency:
       oldGroupId: org.springframework.boot
       oldArtifactId: spring-boot-starter-oauth2-authorization-server


src/main/resources/META-INF/rewrite/spring-boot-40.yml
```diff
@@ -53,3 +53,27 @@ recipeList:
   - org.openrewrite.gradle.plugins.UpgradePluginVersion:
       pluginIdPattern: org.springframework.boot
       newVersion: 4.0.x
+  - org.openrewrite.java.dependencies.ChangeDependency:
+      oldGroupId: org.springframework.boot
+      oldArtifactId: spring-boot-starter-aop
+      newArtifactId: spring-boot-starter-aspectj
```
timtebeek: The migration guide recommends to reevaluate whether you need the replacement:
https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-4.0-Migration-Guide#aop-starter-pom

> To clarify the scope of spring-boot-starter-aop, it has been renamed to spring-boot-starter-aspectj. If you have added the starter explicitly in your application, please review if you actually need it before using the replacement.
> 
> To figure this out, review the following:
> 
> Your application does not use AspectJ, typically an annotation in the org.aspectj.lang.annotation package.
> 
> Search for usage of `org.aspectj.lang.annotation.Aspect` on the classpath of your application. They usually document the annotation that triggers the feature, such as Micrometer’s `@Timed` or `@Counted`.

Because of that nuance I think it's best not to include this step just yet, as to [do no harm](https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices#do-no-harm).

```suggestion
```

timtebeek (approved):

src/main/resources/META-INF/rewrite/spring-boot-40.yml
murdos: Then maybe we should do that instead. WDYT?
```yaml
  - org.openrewrite.java.dependencies.RemoveDependency:
      artifactId: spring-boot-starter-aop
  - org.openrewrite.java.dependencies.AddDependency:
      artifactId: spring-boot-starter-aspectj
      version: 4.0.x
      onlyIfUsing: org.aspectj.lang.annotation.*
```

murdos: Or (not tested yet):
```yaml
name: org.openrewrite.java.spring.boot4.MigrateStarterAOP
displayName: Rename spring-boot-starter-aop if necessary
description: >-
  Rename spring-boot-starter-aop to spring-boot-starter-aspectj if there's actually imports to org.aspectj.lang.annotation.*
preconditions:
  - org.openrewrite.java.search.FindImports:
      typePattern: org.aspectj.lang.annotation..*
recipeList:
  - org.openrewrite.java.dependencies.ChangeDependency:
      oldGroupId: org.springframework.boot
      oldArtifactId: spring-boot-starter-aop
      newArtifactId: spring-boot-starter-aspectj
      newVersion: 4.0.x
```

timtebeek: Thanks for the suggestions! That second might lead to the most idiomatic change, as it swaps out one dependency for another, as opposed to potentially placing the new dependency elsewhere. Thanks for iterating on this one!

murdos: I've tested the above recipe using preconditions, and it doesn't seem to work:
- precondition with org.openrewrite.java.search.FindImports doesn't work
- the recipe is incomplete, because we should remove the obsolete spring-boot-starter-aop dependency if there's no annotations used in code (otherwise project won't build)

timtebeek: Just saw the alternative you adopted, that'll work!

To explain why the second does not work it helps to know that preconditions typically apply _per file_. And in a pom.xml file or build.gradle file there is nothing that uses those imports. _Some_ preconditions use markers to mark all files, like `ModuleHasDependency`. In other cases recipes might contain build in scanning capabilities, like `unlessUsing` or `onlyIfUsing`. Those latter options are the best fit here, thanks!

Issue #831, opened 2025-11-18
Migration Recipe for Spring Batch 6.0


ielatif:
[Spring Boot 4.x](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-4.0-Release-Notes) introduces [Spring Batch 6.0](https://github.com/spring-projects/spring-batch/wiki/Spring-Batch-6.0-Migration-Guide) which includes significant changes and deprecations.

I’m interested in contributing to support these updates. Could you provide guidance on how to get started with this contribution?

Tasks:
---

#### Initialize recipe:
- [x] Initialize spring batch 6.0 migration recipe #832

#### Moved APIs:
- [x] Handle infrastructure package change #833 
- [x] Handle explore package change #834
- [x] Handle JDBC DAOs package change #835 
- [x] Handle MongoDB DAOs package change #836
- [x] Handle partition package change #838
- [x] Handle core listener package change #839
- [x] Handle core job package change #841
- [x] Handle core step package change #842 
- [x] Handle job parameters package change #843 

#### Renamed APIs:
- [x] Rename ChunkHandler to ChunkRequestHandler #847 
- [x] Rename JobRepositoryFactoryBean to JdbcJobRepositoryFactoryBean #848 
- [x] Rename JobExplorerFactoryBean to JdbcJobExplorerFactoryBean #849 
- [x] Migrate JobStep to use JobOperator instead of JobLauncher #858 
- [x] Migrate JobStepBuilder to use JobOperator instead of JobLauncher
- [x] Migrate JobLaunchingGateway constructor to use JobOperator
- [x] Migrate JobLaunchingMessageHandler constructor to use JobOperator
- [x] Migrate SystemCommandTasklet#setJobExplorer method
- [x] Migrate RemoteStepExecutionAggregator constructor to use JobRepository
- [x] Migrate RemoteStepExecutionAggregator#setJobExplorer method

timtebeek:
hi @ielatif ! Thanks for the offer to help. The best way to start is create a copy of this file for 6.0, and open a pull request. Then we can fill out the details either on your PR or in subsequent PRs.
https://github.com/openrewrite/rewrite-spring/blob/5e08d4f03e6ade6a59da71ef3d4da29cb3044824/src/main/resources/META-INF/rewrite/spring-batch-5.0.yml#L18-L31

timtebeek:
[...]

PR #858, opened 2025-12-02
Migrate JobStep to use JobOperator instead of JobLauncher

ielatif:
- Related to #831

## What's changed?
The setter `org.springframework.batch.core.step.job.JobStep#setJobLauncher(JobLauncher jobLauncher)` was renamed to `setJobOperator(JobOperator jobOperator)` and now requires a `JobOperator` instead of a `JobLauncher`
https://github.com/spring-projects/spring-batch/wiki/Spring-Batch-6.0-Migration-Guide#renamed-apis

## What's your motivation?
Support Spring batch latest version 6.0

## Any additional context
spring-batch 5.0.x version https://github.com/spring-projects/spring-batch/blob/5.0.x/spring-batch-core/src/main/java/org/springframework/batch/core/step/job/JobStep.java

spring-batch 6.0.x version https://github.com/spring-projects/spring-batch/blob/main/spring-batch-core/src/main/java/org/springframework/batch/core/step/job/JobStep.java

@timtebeek


src/main/java/org/openrewrite/java/spring/batch/MigrateJobStep.java
```diff
@@ -0,0 +1,237 @@
...
+			cd = cd.withBody(
+					cd.getBody().withStatements(
+							cd.getBody().getStatements().stream()
+									.map(this::cleanConstructor)
+									.filter(Objects::nonNull) // remove empty constructors
+									.collect(toList())
+					)
+			);
+
+			return cd;
```
timtebeek:
```suggestion
			);
```

src/main/java/org/openrewrite/java/spring/batch/MigrateJobStep.java
```diff
@@ -0,0 +1,237 @@
...
+		}
+
+		@Override
+		public J.MethodDeclaration visitMethodDeclaration(J.MethodDeclaration md, ExecutionContext ctx) {
```
timtebeek:
```suggestion
		public J.@Nullable MethodDeclaration visitMethodDeclaration(J.MethodDeclaration md, ExecutionContext ctx) {
```

timtebeek:
Glad to hear; I'm traveling right now and preparing a new release, so it might take me a little longer to dive into the case you've linked. What are the options you're considering?

src/main/java/org/openrewrite/java/spring/batch/MigrateJobStep.java
```diff
@@ -0,0 +1,198 @@
...
+                                .collect(toList());
+
+                        // Remove the assignment: this.jobLauncher = jobLauncher;
+                        J.Block newBody = md.getBody().withStatements(
+                                md.getBody().getStatements().stream()
+                                        .filter(s -> !isJobLauncherAssignment(s))
+                                        .collect(toList())
+                        );
+
+                        J.MethodDeclaration updated = md.withParameters(newParams).withBody(newBody);
```
timtebeek: Making a note for myself that we'll probably want to replace these stream usages with `ListUtils` methods `filter` and `map`.

timtebeek (commented):
Thanks again @ielatif ; Great to see your progress. I need to find some more time for the final polish, but have a workshop today that pushes this back a bit. I've added a quick note but need to check in more detail still.

timtebeek (approved):
Thanks for the changes here @ielatif ! Took me a little while longer to deduce what you were after, but I've distilled it down to as minimal a change as I could.

From past experiences we've learned it's best not to change variable names or method signatures, as those might be used in unexpected ways in lines outside what we test for here. Imagine for instance there might be conditionals, or tests, that expect a particular signature.

I'd also seen that more broadly [the JobLauncher is deprecated](https://github.com/spring-projects/spring-batch/blob/7eaaa029e24a570802c87dd03aee77588cd4d899/spring-batch-core/src/main/java/org/springframework/batch/core/launch/JobLauncher.java#L36-L42) with JobOperator the suggested replacement, so I figured a ChangeType would address more such cases. That then leaves us with changing the method name, which we already had a building block recipe for.

Hope you agree with the changes I've made; the tests you added were very helpful in doing so with confidence, thanks! 🙏🏻

PR #862, opened 2025-12-04
Extends ChangeSpringPropertyKey to handle properties in @SpringBootTest annotations

murdos:
## What's your motivation?
Properties should also be updated when used in tests configuration with annotations such as `@SpringBootTest` and other tests slices:
```java
@SpringBootTest(webEnvironment = RANDOM_PORT,
		properties = { "server.error.include-message=ALWAYS", "management.endpoints.enabled-by-default=false" })
```

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/java/spring/ChangeSpringPropertyKey.java
+++ b/src/main/java/org/openrewrite/java/spring/ChangeSpringPropertyKey.java
@@ -232,7 +232,9 @@ public class ChangeSpringPropertyKey extends Recipe {
             Pattern pattern = Pattern.compile("^" + quote(oldPropertyKey) + exceptRegex());
             Matcher matcher = pattern.matcher(value);
             if (matcher.find()) {
-                return literal.withValueSource(literal.getValueSource().replaceFirst(quote(oldPropertyKey), newPropertyKey)).withValue(value.replaceFirst(quote(oldPropertyKey), newPropertyKey));
+                return literal
+                        .withValue(value.replaceFirst(quote(oldPropertyKey), newPropertyKey))
+                        .withValueSource(literal.getValueSource().replaceFirst(quote(oldPropertyKey), newPropertyKey));
             }
             return literal;
         }


src/test/java/org/openrewrite/java/spring/ChangeSpringPropertyKeyTest.java
```diff
@@ -388,4 +389,101 @@ public String fooBazes() {
           )
         );
     }
+
+    @ParameterizedTest
```
murdos: I use parameterized tests because my intent was to test other test slices such as `DataJpaTest`, but I was not able to find the correct classpathResource to make it works :/

src/main/java/org/openrewrite/java/spring/ChangeSpringPropertyKey.java
```diff
@@ -117,6 +118,8 @@ private class JavaPropertyKeyVisitor extends JavaIsoVisitor<ExecutionContext> {
                 new AnnotationMatcher("@org.springframework.beans.factory.annotation.Value");
         private final AnnotationMatcher CONDITIONAL_ON_PROPERTY_MATCHER =
                 new AnnotationMatcher("@org.springframework.boot.autoconfigure.condition.ConditionalOnProperty");
+        private final AnnotationMatcher SPRINT_BOOT_TEST_MATCHER =
```
timtebeek:
```suggestion
        private final AnnotationMatcher SPRING_BOOT_TEST_MATCHER =
```

src/main/java/org/openrewrite/java/spring/ChangeSpringPropertyKey.java
```diff
@@ -172,24 +175,70 @@ public J.Annotation visitAnnotation(J.Annotation annotation, ExecutionContext ct
...
-                                );
+                            J.Literal newLiteral = changePropertyInLiteral(literal);
+                            if (newLiteral != literal) {
+                                arg = assignment.withAssignment(newLiteral);
                             }
                         }
                         return arg;
                     }));
                 }
+            } else if (SPRINT_BOOT_TEST_MATCHER.matches(annotation)) {
```
timtebeek:
```suggestion
            } else if (SPRING_BOOT_TEST_MATCHER.matches(annotation)) {
```

PR #873, opened 2025-12-15
spring-boot-40: handle rename of AOP Starter

murdos:
- See https://github.com/openrewrite/rewrite-spring/pull/856#discussion_r2588874522


timtebeek (approved):
Creative combination, I like it, thanks!

PR #874, opened 2025-12-16
Add kotlin tests

Laurens-W:
## What's changed?

Added Kotlin based tests to validate these recipes work

### Checklist


src/test/java/org/openrewrite/java/spring/ImplicitWebAnnotationNamesTest.java
```diff
@@ -182,4 +185,123 @@ public ResponseEntity<String> getUser(@PathVariable("uid") Long id,
           )
         );
     }
+
+    @Test
+    void removeUnnecessaryAnnotationArgumentKotlin() {
```
timtebeek, line 190: Might it make sense to add the Kotlin tests in a `@Nested class Kotlin { ... }` ? That way it's clear they might mimic coverage for a specific purpose.

PR #875, opened 2025-12-18
spring-boot-40: handle rename of spring-boot-data-jpa packages

murdos:
And handle metrics packages of spring-boot-hibernate

### Checklist


timtebeek (approved):
Thanks a lot for the continued refining here.

PR #877, opened 2025-12-18
Add JUnit6BestPractices migration to spring-boot-40.yml

murdos:
Since SpringBoot 4 is using JUnit6, we should have a recipe that upgrade to JUnit6 API.
 I'm not sure what's best between `org.openrewrite.java.testing.junit.JUnit6BestPractices` and `org.openrewrite.java.testing.junit6.JUnit5to6Migration`, since junit recipes are a bit of a mess.


murdos:
Alright. I changed the recipe.

PR #885, opened 2025-12-22
Fix JdbcTemplateObjectArrayArgToVarArgs for Object[] argument reordering

Gautam-aman:
### What
Fixes JdbcTemplateObjectArrayArgToVarArgs to correctly handle deprecated JdbcTemplate overloads that accept Object[] arguments.

### Why
Some Object[]-based overloads were not consistently handled, especially when used with RowMapper and queryForList variants.

### Changes
- Safely reorder Object[] arguments without variable inlining
- Cover query, queryForObject, and queryForList variants
- Add tests asserting both change and no-change behavior

### Verification
All tests pass (`./gradlew test`).

### DIFF
--- a/src/main/resources/META-INF/rewrite/examples.yml
+++ b/src/main/resources/META-INF/rewrite/examples.yml
@@ -1722,7 +1722,7 @@ examples:
   - before: project
     language: mavenProject
   - before: |
-      <project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+      <project>
         <modelVersion>4.0.0</modelVersion>
         <groupId>org.springframework.samples</groupId>
         <artifactId>spring-petclinic</artifactId>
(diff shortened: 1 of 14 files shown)


timtebeek:
Thanks for the help here @Gautam-aman ; I'd noticed you'd reduced all the tests down to using a single text block of the target state: That's not accurate unfortunately.

Tests use two text blocks: one showing the before situation, and one showing the after situation, and then assert that the recipe turns one into the other with some correctness checks added on top. In rare cases tests can use a single text block as input, mostly to verify that no changes are made in a certain condition, for instance because a precondition isn't met, or because the target end state is already achieved.

In your case all tests were of that latter form, which meant the actual code of the recipe was no longer exercised. I've instead restored (and trimmed) the original tests to verify that we still make the changes we did before. I've also added a new test for the case you were likely after: https://github.com/openrewrite/rewrite-spring/pull/885/commits/ee51cf830ad655ce3cae2adc2cba9f38927157e6

Hope that helps! I look forward to what else you'll find as you explore and improve our recipes.

Gautam-aman:
Thanks for the detailed explanation , that clarifies the testing pattern well.
I appreciate you restoring and trimming the tests and adding coverage for the case I was aiming for.
This was very helpful. Looking forward to contributing more.

PR #891, opened 2026-01-02
Add Spring doc 3 migration recipe

wyfrel:
## What's changed?

Add spring doc 3  migration recipe and add it to Spring Boot 4 migration recipe list

## What's your motivation?

Align with compatibility matrix : https://springdoc.org/#what-is-the-compatibility-matrix-of-springdoc-openapi-with-spring-boot

Should i update the recipe.csv ?

### Checklist


timtebeek:
You're welcome; that validation is rather new; I'm adding hints here if it were to fail again in the future:
- https://github.com/openrewrite/rewrite/pull/6475

src/main/resources/META-INF/rewrite/spring-boot-40.yml
murdos, line 103: @wyfrel, @timtebeek: This was already handled here, so this one is now redundant.

timtebeek: Thanks for noticing! Now removed in https://github.com/openrewrite/rewrite-spring/commit/d79e08564a726fb9f225a28e28917c8b669d66c7

PR #893, opened 2026-01-06
feat: Add OpenTelemetry migration recipes for Spring Boot applications

sandy2008:
## What's changed?

Add comprehensive OpenTelemetry migration recipes to help Spring Boot applications migrate from legacy tracing solutions to OpenTelemetry.

**New YAML Recipes (`opentelemetry.yml`):**
- `AddOpenTelemetryAutoInstrumentation` - Add OpenTelemetry auto-instrumentation with `opentelemetry-spring-boot-starter`
- `MigrateToOpenTelemetryFromSleuth` - Migrate from Spring Cloud Sleuth to OpenTelemetry
- `MigrateBraveToOpenTelemetry`, `MigrateFromZipkinToOpenTelemetry` - Migrate from Brave/Zipkin
- `ConfigureOpenTelemetryForSpringBoot` - Configure OTLP exporter and service name
- `OpenTelemetryCompleteMigration` - Complete end-to-end migration recipe
- `MigrateSleuthToMicrometerTracing` - Migrate Spring Cloud Sleuth API types to Micrometer Tracing
- `MigrateSleuthAnnotationsToOpenTelemetry` - Migrate `@NewSpan`, `@SpanTag`, `@ContinueSpan` to `@WithSpan`, `@SpanAttribute`
- `MigrateOpenTracingToOpenTelemetry` - Migrate OpenTracing API to OpenTelemetry API
- `MigrateDatadogToOpenTelemetry` - Migrate Datadog tracing API and `@Trace` annotation to OpenTelemetry

## What's your motivation?

OpenTelemetry has become the industry standard for observability. Spring Cloud Sleuth is deprecated in favor of Micrometer Tracing, and many organizations are migrating from vendor-specific solutions (Datadog, Zipkin, Brave) to OpenTelemetry for vendor-neutral observability.

These recipes automate the migration process, handling:
- Dependency management (BOM, starter dependencies)
- API type changes (Tracer, Span, SpanContext, Scope)
- Annotation migrations (@NewSpan → @WithSpan, @Trace → @WithSpan)
- Property configuration for OTLP exporters

- The OpenTelemetry version used is **2.23.0** (latest as of January 2026) with SDK **1.57.0**
- The recipes follow the official [OpenTelemetry Spring Boot documentation](https://opentelemetry.io/docs/zero-code/java/spring-boot-starter/)
- Using `opentelemetry-spring-boot-starter` for zero-code auto-instrumentation approach

- Considered using `classpathFromResources` for tests but used `dependsOn` with stub types for better test isolation
- The YAML recipes delegate to existing OpenRewrite recipes (`AddDependency`, `ChangeType`, `ChangeSpringPropertyKey`) for maintainability

## Any additional context

Uses OpenTelemetry Java Instrumentation **2.23.0** which includes:
- Auto-instrumentation for Spring Web, WebFlux, JDBC, Kafka, etc.
- OTLP exporter for traces, metrics, and logs
- Support for `@WithSpan` and `@SpanAttribute` annotations

### Checklist

### DIFF
--- a/src/main/resources/META-INF/rewrite/opentelemetry.yml
+++ b/src/main/resources/META-INF/rewrite/opentelemetry.yml
@@ -36,6 +36,12 @@ recipeList:
   - org.openrewrite.java.spring.opentelemetry.MigrateNewRelicToOpenTelemetry
   - org.openrewrite.java.spring.opentelemetry.MigrateDatadogToOpenTelemetry
 
+  - org.openrewrite.java.dependencies.AddDependency:
+      groupId: io.opentelemetry.instrumentation
+      artifactId: opentelemetry-logback-appender-1.0
+      onlyIfUsing: ch.qos.logback.*
+      acceptTransitive: true
+
 ---
 type: specs.openrewrite.org/v1beta/recipe
 name: org.openrewrite.java.spring.opentelemetry.MigrateSleuthToOpenTelemetry
@@ -64,12 +70,14 @@ recipeList:
   - org.openrewrite.java.dependencies.AddDependency:
       groupId: io.micrometer
       artifactId: micrometer-tracing-bridge-otel
+      version: 1.x
       onlyIfUsing: org.springframework.cloud.sleuth.*
       acceptTransitive: true
   # Add OpenTelemetry exporters for common backends
   - org.openrewrite.java.dependencies.AddDependency:
       groupId: io.opentelemetry
       artifactId: opentelemetry-exporter-otlp
+      version: 1.x
       onlyIfUsing: org.springframework.cloud.sleuth.*
       acceptTransitive: true
 
@@ -93,12 +101,14 @@ recipeList:
   - org.openrewrite.java.dependencies.AddDependency:
       groupId: io.opentelemetry.instrumentation
       artifactId: opentelemetry-instrumentation-annotations
+      version: 2.x
       onlyIfUsing: org.springframework.cloud.sleuth.*
       acceptTransitive: true
   # Add Spring Boot AOP starter for annotation processing
   - org.openrewrite.java.dependencies.AddDependency:
       groupId: org.springframework.boot
       artifactId: spring-boot-starter-aop
+      version: 3.x
       onlyIfUsing: io.opentelemetry.instrumentation.annotations.*
       acceptTransitive: true
 
@@ -181,11 +191,6 @@ recipeList:
       acceptTransitive: true
   # Note: The user will need to manually update their collector configuration
   # to accept OTLP instead of Zipkin format
-  - org.openrewrite.java.dependencies.AddDependency:
-      groupId: io.opentelemetry.instrumentation
-      artifactId: opentelemetry-logback-appender-1.0
-      onlyIfUsing: ch.qos.logback.*
-      acceptTransitive: true
 
 ---
 type: specs.openrewrite.org/v1beta/recipe
@@ -209,6 +214,7 @@ recipeList:
   - org.openrewrite.java.dependencies.AddDependency:
       groupId: io.opentelemetry
       artifactId: opentelemetry-api
+      version: 1.x
       onlyIfUsing: brave.*
       acceptTransitive: true
 
@@ -260,6 +266,7 @@ recipeList:
   - org.openrewrite.java.dependencies.AddDependency:
       groupId: io.opentelemetry
       artifactId: opentelemetry-api
+      version: 1.x
       onlyIfUsing: io.opentelemetry.api.*
       acceptTransitive: true
 
@@ -313,6 +320,7 @@ recipeList:
   - org.openrewrite.java.dependencies.AddDependency:
       groupId: io.opentelemetry.instrumentation
       artifactId: opentelemetry-instrumentation-annotations
+      version: 2.x
       onlyIfUsing: com.newrelic.api.agent.*
       acceptTransitive: true
 
@@ -346,6 +354,7 @@ recipeList:
   - org.openrewrite.java.dependencies.AddDependency:
       groupId: io.opentelemetry.instrumentation
       artifactId: opentelemetry-instrumentation-annotations
+      version: 2.x
       onlyIfUsing: datadog.trace.api.*
       acceptTransitive: true
 


src/main/resources/META-INF/rewrite/opentelemetry.yml
```diff
@@ -0,0 +1,611 @@
...
+  # Add OpenTelemetry annotations dependency
+  - org.openrewrite.java.dependencies.AddDependency:
+      groupId: io.opentelemetry.instrumentation
+      artifactId: opentelemetry-instrumentation-annotations
+      onlyIfUsing: io.opentelemetry.instrumentation.annotations.*
+      acceptTransitive: true
+
+---
+type: specs.openrewrite.org/v1beta/recipe
+name: org.openrewrite.java.spring.boot3.MigrateDatadogToOpenTelemetry
```
timtebeek: This recipe doesn't seem to be referenced anywhere; meanwhile there is also a recipe with the same name but different package in `org.openrewrite.java.spring.opentelemetry.MigrateDatadogToOpenTelemetry`. Can that latter one be merged into this declarative recipe?

The same goes for `MigrateOpenTracingToOpenTelemetry` and `MigrateSleuthAnnotationsToOpenTelemetry`: those also each have an apparent duplicate.

src/main/java/org/openrewrite/java/spring/opentelemetry/MigrateSleuthToMicrometerTracing.java
```diff
@@ -0,0 +1,93 @@
...
+/**
+ * Migrates Spring Cloud Sleuth API to Micrometer Tracing API.
+ * <p>
+ * Spring Cloud Sleuth has been deprecated and is replaced by Micrometer Tracing
+ * with OpenTelemetry as the recommended backend.
+ *
+ * @see <a href="https://github.com/spring-cloud/spring-cloud-sleuth#spring-cloud-sleuth">Spring Cloud Sleuth README</a>
+ * @see <a href="https://micrometer.io/docs/tracing">Micrometer Tracing Documentation</a>
+ */
+public class MigrateSleuthToMicrometerTracing extends Recipe {
```
timtebeek: This recipe isn't referenced anywhere; does it need to be included in `opentelemetry.yml`? Or can it be removed perhaps?

src/main/resources/META-INF/rewrite/opentelemetry.yml
```diff
@@ -0,0 +1,611 @@
...
+  # Add OpenTelemetry annotations dependency
+  - org.openrewrite.java.dependencies.AddDependency:
+      groupId: io.opentelemetry.instrumentation
+      artifactId: opentelemetry-instrumentation-annotations
+      onlyIfUsing: io.opentelemetry.instrumentation.annotations.*
+      acceptTransitive: true
+
+---
+type: specs.openrewrite.org/v1beta/recipe
+name: org.openrewrite.java.spring.boot3.OpenTelemetryCompleteMigration
```
timtebeek: Should this recipe be included in any larger composite such as a particular Spring Boot version migration? I'm not sure at which point it might have become a required migration.

timtebeek (changes requested):
I had a couple quick questions pop up during the initial reading; would be good to get clarity on those before we continue.

src/main/resources/META-INF/rewrite/opentelemetry.yml
sandy2008: Thank you! I added `This is an optional migration for users who want to
  adopt OpenTelemetry as their observability solution. It is not included in standard Spring Boot
  version migrations because those use Micrometer Tracing (which can work with either Brave or
  OpenTelemetry as a backend). Use this recipe if you want to fully embrace OpenTelemetry's
  native instrumentation and exporters.`

Since Sleuth → Micrometer Tracing + Brave is already in `UpgradeSpringBoot_3_0`, is it better to make it optional..?

src/main/java/org/openrewrite/java/spring/opentelemetry/MigrateDatadogToOpenTelemetry.java
```diff
@@ -0,0 +1,83 @@
...
+ * This recipe handles the migration of:
+ * <ul>
+ *   <li>Datadog Tracer → OpenTelemetry Tracer</li>
+ *   <li>Datadog Span → OpenTelemetry Span</li>
+ *   <li>Datadog SpanContext → OpenTelemetry SpanContext</li>
+ *   <li>Datadog Scope → OpenTelemetry Scope</li>
+ *   <li>Datadog @Trace annotation → OpenTelemetry @WithSpan</li>
+ * </ul>
+ */
+public class MigrateDatadogToOpenTelemetry extends Recipe {
```
timtebeek: This still looks to have a parallel in
https://github.com/openrewrite/rewrite-spring/blob/91c93ce37b9a2d89a665de29086e12a4902d358d/src/main/resources/META-INF/rewrite/opentelemetry.yml#L460-L484

I think ideally we move all of these steps into `opentelemetry.yml`, as I don't quite see a reason to have this as an imperative recipe.

src/main/java/org/openrewrite/java/spring/opentelemetry/MigrateOpenTracingToOpenTelemetry.java
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ * Migrates OpenTracing API to OpenTelemetry API.
+ * <p>
+ * OpenTracing has been superseded by OpenTelemetry and is no longer actively maintained.
+ * This recipe handles the migration of core OpenTracing types to their OpenTelemetry equivalents.
+ *
+ * @see <a href="https://opentelemetry.io/docs/migration/opentracing/">OpenTracing Migration Guide</a>
+ */
+public class MigrateOpenTracingToOpenTelemetry extends Recipe {
```
timtebeek: Same here as above; I think we can move the steps used here into the similarly named
https://github.com/openrewrite/rewrite-spring/blob/91c93ce37b9a2d89a665de29086e12a4902d358d/src/main/resources/META-INF/rewrite/opentelemetry.yml#L365-L403

If not we need to find a better way to distinguish those recipes, as it would be confusing as is right now to have the same names in different packages.

PR #894, opened 2026-01-06
Add `newVersion` to `ChangeDependency` steps for Spring Boot 4.0

timtebeek:
As we got feedback about these not-yet being version managed previously leading to issues.

### DIFF
--- a/src/main/resources/META-INF/rewrite/spring-cloud-2025.yml
+++ b/src/main/resources/META-INF/rewrite/spring-cloud-2025.yml
@@ -102,32 +102,32 @@ recipeList:
       oldGroupId: org.springframework.cloud
       oldArtifactId: spring-cloud-gateway-server
       newArtifactId: spring-cloud-gateway-server-webflux
-      newVersion: 5.0.x
+      newVersion: 4.3.x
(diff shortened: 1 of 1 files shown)


src/main/resources/META-INF/rewrite/spring-boot-40.yml
```diff
@@ -63,22 +63,27 @@ recipeList:
       oldGroupId: org.springframework.boot
       oldArtifactId: spring-boot-starter-oauth2-authorization-server
       newArtifactId: spring-boot-starter-security-oauth2-authorization-server
+      newVersion: 4.0.x
```
murdos, line 66: This is strange: this should be covered by this recipe that is executed in the same recipe, no?:
```
  - org.openrewrite.java.dependencies.UpgradeDependencyVersion:
      artifactId: "*"
      newVersion: 4.0.x
      overrideManagedVersion: false
```

timtebeek: It should, yes, but we're also getting some edge cases reported by friends also involved with
- https://github.com/openrewrite/rewrite/issues/6463

This ought to help them there, and shouldn't be harmful elsewhere ideally.

PR #895, opened 2026-01-07
Add recipe Spring Boot 4.0 Modular Starters Flyway Dependency Precondition

wleese:
## What's changed?

Adds a recipe that compliments `org.openrewrite.java.spring.boot4.MigrateToModularStarters`, specifically the handling of the Flyway starter migration.

## What's your motivation?

The existing recipe does not take into account that Flyway autoconfiguration in Spring Boot may be used without any actual code imports from `flyway-core`.

I didn't aim to cover more than MySQL and Postgres cases. 
I also wasn't sure on where to put this recipe.
Finally, recipe naming might be a bit awkward.

## Any additional context

- Note that I had to use `onlyIfUsing` to workaround [this issue](https://github.com/openrewrite/rewrite/issues/6468) which we encountered due to a Maven project setup we commonly use.

### Checklist

I often don't see negative cases for recipes so I didn't add them. Please correct me if I'm wrong and I'll fix.

- [ x ] I've read and applied the [recipe conventions and best practices](https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices)
- [ x ] I've used the IntelliJ IDEA auto-formatter on affected files


src/main/resources/META-INF/rewrite/spring-boot-40.yml
```diff
@@ -35,6 +35,7 @@ recipeList:
   - org.openrewrite.java.testing.testcontainers.Testcontainers2Migration
   - org.openrewrite.java.springdoc.UpgradeSpringDoc_3_0
   - org.openrewrite.java.spring.boot4.MigrateToModularStarters
+  - org.openrewrite.java.spring.boot4.MigrateToModularStartersFlywayDependencyPrecondition
```
murdos: IMO this should be added to `org.openrewrite.java.spring.boot4.MigrateToModularStarters` instead, since it completes https://github.com/openrewrite/rewrite-spring/blob/96e1007da5756f453f9a257eb3af7ad839808fdb/src/main/resources/META-INF/rewrite/spring-boot-40-modular-starters.yml#L58-L63

src/main/resources/META-INF/rewrite/spring-boot-40-modular-starters-flyway-dependency-precondition.yml
```diff
@@ -0,0 +1,44 @@
...
+  Spring Boot 4.0 has a modular design requiring explicit starters for each feature.
+  This recipe detects feature usage via dependencies and add the Flyway starter.
+preconditions:
+  - org.openrewrite.java.spring.boot4.MigrateToModularStartersFlywayDependencyPreconditions
+recipeList:
+  - org.openrewrite.java.dependencies.AddDependency:
+      groupId: org.springframework.boot
+      artifactId: spring-boot-starter-flyway
+      version: 4.0.x
+      onlyIfUsing: "*..*" # https://github.com/openrewrite/rewrite/issues/6468
```
murdos: How about just using and remove the preconditions especially as you're only handling mysql and postgres databases? 
```suggestion
      onlyIfUsing: "org.flywaydb.*"
```

src/main/resources/META-INF/rewrite/spring-boot-40.yml
wleese: I've moved the recipes to `MigrateToModularStarters`

src/main/resources/META-INF/rewrite/spring-boot-40-modular-starters-flyway-dependency-precondition.yml
wleese: The functional requirement here is to add the Spring Boot Flyway starter based on only the presence of a dependency. This mimics applications which do not directly call flyway via their application code, but rather autoconfiguration being triggered by the presence of migration sql files.
I'm using `onlyIfUsing` here just to work around the problem reported here
- https://github.com/openrewrite/rewrite/issues/6468

PR #896, opened 2026-01-08
Bump spring ws when moving to spring boot 3

Jenson3210:
This is the version of Spring WS you’ll need if you are building SOAP-based applications with Spring Boot 3.0.

more information can be found [here](https://spring.io/blog/2022/11/21/spring-web-services-4-0-0-is-now-ga)

We will immediately bump to [4.1.0](https://spring.io/blog/2025/05/20/spring-ws-4-1-0-available-now)

### DIFF
--- a/src/main/resources/META-INF/rewrite/spring-boot-30.yml
+++ b/src/main/resources/META-INF/rewrite/spring-boot-30.yml
@@ -47,6 +47,10 @@ recipeList:
       groupId: org.springframework
       artifactId: "*"
       newVersion: 6.0.x
+  - org.openrewrite.java.dependencies.UpgradeDependencyVersion:
+      groupId: org.springframework.ws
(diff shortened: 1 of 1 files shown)


timtebeek:
> I see https://docs.spring.io/spring-boot/3.5/appendix/dependency-versions/coordinates.html that SB 3.5 has some ws dependencies managed on version 4.1.x, do we also want to add that bump?

I indeed see 3.5 using `4.1.2`, but that should still be covered by the `4.x` we're using in the recipe right?
image

https://github.com/openrewrite/rewrite-spring/blob/756416b7cdc694cf7deed00e063d05e16d22bb5a/src/main/resources/META-INF/rewrite/spring-boot-30.yml#L50-L53

timtebeek (approved):

Issue #754, opened 2025-07-04
Spring Boot 4 migration recipe


Laurens-W:
## What problem are you trying to solve?
Prepare for the Spring Boot 4 release later this year

## Describe the solution you'd like
A Yaml recipe that can be used to compose the migration recipes
Early draft PRs with test cases so the community can provide feedback

## Additional context

- https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-4.0-Release-Notes
- https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-4.0-Migration-Guide
- The official release will probably be somewhere towards November [see source, final paragraph](https://spring.io/blog/2024/10/01/from-spring-framework-6-2-to-7-0)
- https://github.com/spring-projects/spring-framework/wiki/Spring-Framework-7.0-Release-Notes

sruffatti:
Here is a start! Please review: https://github.com/openrewrite/rewrite-spring/pull/766

wyfrel:
@sruffatti : I've noticed that the LocalServerPort annotation is moved from `org.springframework.boot.test.web.server.LocalServerPort` to `org.springframework.boot.web.server.test.LocalServerPort`.

I don't want to mess up with your MR may be you can add a recipe for this ?

DidierLoiseau:
FYI, Testcontainers 2.0 made it for SB 4! This involves a renaming of the packages – I didn’t check if relocation was published, maybe a migration recipe will be required?

Moreover this includes the long awaited fix for testcontainers/testcontainers-java#970 – removing the JUnit 4 dependency! This means the recipe I updated in openrewrite/rewrite-testing-frameworks#746 should be refined to only apply for Testcontainers 1.

wyfrel:
> FYI, Testcontainers 2.0 made it for SB 4! This involves a renaming of the packages – I didn’t check if relocation was published, maybe a migration recipe will be required?

done here : https://github.com/openrewrite/rewrite-spring/pull/823

PR #901, opened 2026-01-15
Add recipes for Jackson 3 migration in Spring Kafka 4

jjank:
- Contributes to #754

## What's changed?
* Initial support for Spring Kafka 4 migration
* In particular: [Jackson 3 related migrations](https://docs.spring.io/spring-kafka/reference/whats-new.html#x40-jackson3-support)

I "plugged" this into the Boot 4 migrations. But it could - in theory - also be part of the Spring Framework 7 migration since Spring Kafka has no direct dependency to Spring Boot.

I'm not terribly familiar with the "new" typetables stuff - I hope I got the parser classpath right

## Any additional context
There's more Jackson-related deprecations that are similar drop-in-replacements in Spring-Kafka. But I wanted to focus on the specific classes pointed out in the documentation first.

### Checklist

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -17,6 +17,7 @@ recipeDependencies {
     parserClasspath("javax.persistence:javax.persistence-api:2.+")
     parserClasspath("javax.validation:validation-api:2.0.1.Final")
     parserClasspath("org.junit.jupiter:junit-jupiter-api:5.+")
+
     parserClasspath("org.springframework.boot:spring-boot:1.+")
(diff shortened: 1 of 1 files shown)


build.gradle.kts
```diff
@@ -18,6 +18,7 @@ recipeDependencies {
     parserClasspath("javax.validation:validation-api:2.0.1.Final")
     parserClasspath("org.junit.jupiter:junit-jupiter-api:5.+")
 
+
```
timtebeek:
```suggestion

```

PR #902, opened 2026-01-15
Update package names for spring-boot-graphql

tschut:
Noticed a missing package update in this recipe


timtebeek (approved):
Thanks a lot!

PR #931, opened 2026-02-16
Explain JSpecify annotation migration in Spring Framework 7 description

Jenson3210:
## Summary
- Updated `UpgradeSpringFramework_7_0` recipe description to mention the JSpecify annotation migration

Customers running Spring Boot 4 migrations were confused about why `org.springframework.lang.@Nullable` and `@NonNull` were being changed. The description now explains that Spring Framework 7 has adopted JSpecify as its nullability annotation standard.

- Closes moderneinc/customer-requests#1869


Jenson3210:
We should not list every possible change from a recipe in its description

Issue #745, opened 2025-06-20
Migrate RepositoryRestConfigurerAdapter to RepositoryRestConfigurer


FieteO:
## What problem are you trying to solve?
As part of the spring-boot (2.4) update, `spring-data-rest-webmvc` is updated to `3.4.X`. With this release, the deprecated `import org.springframework.data.rest.webmvc.config.RepositoryRestConfigurerAdapter` is removed and `import org.springframework.data.rest.webmvc.config.RepositoryRestConfigurer` should be implemented instead.

## Before

```java
import org.springframework.context.annotation.Configuration;
import org.springframework.data.rest.core.config.RepositoryRestConfiguration;
import org.springframework.data.rest.webmvc.config.RepositoryRestConfigurerAdapter;

@Configuration
public class FooRepositoryConfiguration extends RepositoryRestConfigurerAdapter {

    @Override
    public void configureRepositoryRestConfiguration(RepositoryRestConfiguration config) {}
}
```

## After

```java
import org.springframework.context.annotation.Configuration;
import org.springframework.data.rest.core.config.RepositoryRestConfiguration;
import org.springframework.data.rest.webmvc.config.RepositoryRestConfigurer;

@Configuration
public class CatalogDbRepositoryConfiguration implements RepositoryRestConfigurer {

    @Override
    public void configureRepositoryRestConfiguration(RepositoryRestConfiguration config) {}
}
```

## Are you interested in [contributing this feature to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#new-features)?
No

PR #934, opened 2026-02-17
Replace `RepositoryRestConfigurerAdapter` with `RepositoryRestConfigurer`

MichalTHEDUDE:
## What's changed?
Migrating RepositoryRestConfigurerAdapter into RepositoryRestConfigurer

## What's your motivation?
- [](https://github.com/openrewrite/rewrite-spring/issues/745)

I'm not sure about the solution as it is copy of `MigrateWebMvcConfigurerAdapter` recipe :).
Could you also give me some tips about enhancing tests? Is there any other way to test this?
Would appreciate all the feedback and potential hints.

PS. Don't know why but not all tests are passing (not mine) on my env. There is some error with upgarding to Gradle version 10.0.
I'm using Ubuntu 24.04

N/A.

Nope.

## Any additional context
N/A.

### Checklist


src/main/java/org/openrewrite/java/spring/data/MigrateRepositoryRestConfigurerAdapter.java
```diff
@@ -0,0 +1,122 @@
...
+
+            @Override
+            public J.ClassDeclaration visitClassDeclaration(J.ClassDeclaration classDecl, ExecutionContext ctx) {
+                J.ClassDeclaration cd = super.visitClassDeclaration(classDecl, ctx);
+                if (cd.getExtends() != null && TypeUtils.isOfClassType(cd.getExtends().getType(), REPOSITORY_REST_CONFIGURER_ADAPTER)) {
+                    cd = cd.withExtends(null);
+                    updateCursor(cd);
+                    // This is an interesting one... RepositoryRestConfigurerAdapter implements RepositoryRestConfigurer
+                    // remove the super type from the class type to prevent a stack-overflow exception when the JavaTemplate visits class type.
+                    JavaType.Class type = (JavaType.Class) cd.getType();
```
MaciejWaj, line 56: What is a cd? is hard to understand what that mean and it's anti-pattern. Use full name

src/main/java/org/openrewrite/java/spring/data/MigrateRepositoryRestConfigurerAdapter.java
```diff
@@ -0,0 +1,122 @@
...
+                        );
+                    }
+
+                    maybeRemoveImport(REPOSITORY_REST_CONFIGURER_ADAPTER);
+                    maybeAddImport(REPOSITORY_REST_CONFIGURER);
+                }
+                return super.visitMethodDeclaration(md, ctx);
+            }
+
+            class RemoveSuperStatementVisitor extends JavaIsoVisitor<ExecutionContext> {
```
MaciejWaj, line 108: it's a helper class ? This class should be private if you only use it here.

src/main/java/org/openrewrite/java/spring/data/MigrateRepositoryRestConfigurerAdapter.java
```diff
@@ -0,0 +1,122 @@
...
+                final MethodMatcher wm = new MethodMatcher("org.springframework.data.rest.webmvc.config.RepositoryRestConfigurerAdapter *(..)");
+
+                @Override
+                public J.@Nullable MethodInvocation visitMethodInvocation(J.MethodInvocation method, ExecutionContext ctx) {
+                    J.MethodInvocation mi = super.visitMethodInvocation(method, ctx);
+                    if (wm.matches(method.getMethodType())) {
+                        return null;
+                    }
+                    return mi;
```
MaciejWaj, line 117: you can write it more easily 
if (wm.matches(method.getMethodType())) ? null : mi;

timtebeek (approved):
Great to see this covered too; thanks for linking the reference you used for the implementation.

PR #936, opened 2026-02-18
Ensure properties with on-cloud-platform remain unchanged

cachescrubber:
## What's changed?
Added a test (`onCloudPlatformLeftAsIs`) to takle the issue.

## What's your motivation?
Profiles activate by `spring.config.activate.on-cloud-platform` should not be split in separate files because the automatic activation would be lost.

See discussion in Slack #questions.

@timtebeek 

- Review and revert the changes after recipe run.

### Checklist


Jenson3210:
Should we not add this for yaml also?

cachescrubber:
> Should we not add this for yaml also?

Good catch. I'll supply a yaml example and open another pr.

PR #956, opened 2026-03-05
Handle RestClientResponseException subclasses in getRawStatusCode migration

Jenson3210:
## Summary

- Extend `MigrateResponseStatusExceptionGetRawStatusCodeMethod` to also migrate `getRawStatusCode()` calls on `RestClientResponseException` and its subclasses (`HttpServerErrorException`, `HttpClientErrorException`, etc.)
- Add tests for `RestClientResponseException`, `HttpServerErrorException`, and `HttpClientErrorException`

## Problem

The recipe only matched `ResponseStatusException.getRawStatusCode()` but `RestClientResponseException` and its subclasses (like `HttpServerErrorException`) also have `getRawStatusCode()` deprecated in Spring 6, requiring the same migration to `getStatusCode().value()`.

## Solution

Switched from a single `MethodAccess.Matcher` to two `MethodMatcher` instances — one for `ResponseStatusException` and one for `RestClientResponseException` — so that subclasses of both hierarchies are now handled.

## Test plan

- [x] Existing tests pass (2 original tests)
- [x] New tests added for `RestClientResponseException`, `HttpServerErrorException`, `HttpClientErrorException`

- Fixes moderneinc/customer-requests#1955

### DIFF
--- a/src/main/java/org/openrewrite/java/spring/framework/MigrateResponseStatusExceptionGetRawStatusCodeMethod.java
+++ b/src/main/java/org/openrewrite/java/spring/framework/MigrateResponseStatusExceptionGetRawStatusCodeMethod.java
@@ -21,28 +21,37 @@ import org.openrewrite.Recipe;
 import org.openrewrite.TreeVisitor;
 import org.openrewrite.java.JavaParser;
 import org.openrewrite.java.JavaTemplate;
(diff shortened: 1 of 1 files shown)


Jenson3210:
@timtebeek I doubt if we should name the class different when adding this (as this might break usages elsewhere)

PR #995, opened 2026-04-10
Add explicit `scope: test` to test starter dependencies in MigrateToModularStarters

timtebeek:
## Summary
The `MigrateToModularStarters` recipe was adding test starters (e.g. `spring-boot-starter-cache-test`) via `AddDependency` without explicitly setting `scope: test`. This adds `scope: test` to all 7 test starter `AddDependency` entries to ensure the dependencies are always added with test scope.


timtebeek:
We could rely on `onlyIfUisng` to detect the scope that these should go into, but I think it'd be very rare for it not to be `test`, and this makes it consistent with recipes we already use internally.

PR #997, opened 2026-04-14
Add recipe to rename logback.xml to logback-spring.xml

Jenson3210:
## Summary
- Adds `RenameLogbackToLogbackSpring` recipe that renames `logback.xml` → `logback-spring.xml` when the file contains Spring-specific tags (`<springProfile>` or `<springProperty>`)
- Registers the recipe in the `UpgradeSpringBoot_3_0` migration
- Spring Boot only processes these extensions when the config file uses the `-spring.xml` naming convention; a plain `logback.xml` is loaded too early by logback itself

- Closes moderneinc/customer-requests#2210

## Test plan
- [x] Rename when `<springProfile>` is present
- [x] Rename when `<springProperty>` is present
- [x] No change when no Spring tags are present
- [x] No change for `logback-test.xml` (Spring Boot doesn't support `logback-spring-test.xml`)
- [x] Deeply nested Spring tags still trigger rename


timtebeek (approved):

src/main/java/org/openrewrite/java/spring/boot3/RenameLogbackToLogbackSpring.java
```diff
@@ -0,0 +1,54 @@
+/*
+ * Copyright 2024 the original author or authors.
```
steve-aom-elliott: Copyright year should probably be 2026

src/test/java/org/openrewrite/java/spring/boot3/RenameLogbackToLogbackSpringTest.java
```diff
@@ -0,0 +1,162 @@
+/*
+ * Copyright 2024 the original author or authors.
```
steve-aom-elliott: Same deal for copyright year.

src/main/java/org/openrewrite/java/spring/boot3/RenameLogbackToLogbackSpring.java
```diff
@@ -0,0 +1,54 @@
...
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return Preconditions.check(
+                Preconditions.and(
+                        new FindSourceFiles("**/logback.xml").getVisitor(),
+                        Preconditions.or(
+                                new FindTags("//springProfile").getVisitor(),
+                                new FindTags("//springProperty").getVisitor()
+                        )
+                ),
```
steve-aom-elliott, line 50: To confirm, `FindSourceFiles` ANDed with the find tags do actually scope the tag detection to the same file as one named `logback.xml` because it checks per file, right?

Jenson3210: Preconditions run per file, so that should be the case, yes! 
Only a file matching both of these will be renamed.

src/test/java/org/openrewrite/java/spring/boot3/RenameLogbackToLogbackSpringTest.java
```diff
@@ -0,0 +1,162 @@
...
+                </springProfile>
+                <root level="INFO">
+                  <appender-ref ref="STDOUT"/>
+                </root>
+              </configuration>
+              """,
+            spec -> spec.path("src/main/resources/logback.xml")
+          )
+        );
+    }
```
steve-aom-elliott, line 65: Can we verify the after name? Currently the tests just read like the file becomes the same contents but starts as `.../logback.xml`

PR #1001, opened 2026-04-18
Quote YAML values containing colons in `AddSpringProperty`

### LINKED ISSUE #999: AddSpringProperty fails to quote YAML values containing colons
## Problem

`AddSpringProperty` fails when adding a property whose value contains colons (`:`) to a YAML configuration file (`application.yml`). The value is not quoted in the generated YAML snippet, causing `MergeYaml` to fail with a parse error.

### Example

Recipe:
```yaml
- org.openrewrite.java.spring.AddSpringProperty:
    property: my.property
    value: "TODO: Follow this link https://example.com/page"
```

When applied to an `application.yml` file, `AddSpringProperty.createMergeYamlVisitor()` constructs:
```yaml
my:
  property: TODO: Follow this link https://example.com/page
```

`MergeYaml` then fails to parse this because `TODO: Follow...` looks like a YAML key-value pair.

## Root Cause

The `quoteValue` method uses a regex that only matches strings consisting **entirely** of non-alphanumeric, non-whitespace characters:

```java
private static final Pattern scalarNeedsAQuote = Pattern.compile("[^a-zA-Z\\d\\s]*");
private boolean quoteValue(String value) {
    return scalarNeedsAQuote.matcher(value).matches();
}
```

A mixed value like `TODO: Follow link https://example.com` contains alphanumeric characters and spaces, so `.matches()` returns `false` and the value is left unquoted — even though it contains `:` which is a special YAML character.

## Expected Behavior

Values containing YAML-special characters (`:`, `#`, `{`, `}`, `[`, `]`, `,`, `&`, `*`, `?`, `|`, `-`, `<`, `>`, `=`, `!`, `%`, `@`, `` ` ``) should be quoted when constructing the YAML merge snippet.

## Workaround

Use `.properties` files instead of `.yml` — colons are not special in `.properties` format and the recipe works correctly there.

timtebeek:
## Summary
- Replaces the ad-hoc `quoteValue` regex in `AddSpringProperty.createMergeYamlVisitor()` with a call to `StringUtils.quoteIfNeeded()` from rewrite-yaml (added in openrewrite/rewrite#7416).
- The previous regex only quoted values made up entirely of special characters, so mixed values like `TODO: Follow this link https://example.com/page` slipped through unquoted and broke `MergeYaml` parsing.
- Adds a regression test covering that exact scenario for both `.yml` and `.properties`.

- Fixes #999

## Test plan
- [x] `./gradlew test --tests "org.openrewrite.java.spring.AddSpringPropertyTest"`

### DIFF
--- a/src/main/java/org/openrewrite/java/spring/AddSpringProperty.java
+++ b/src/main/java/org/openrewrite/java/spring/AddSpringProperty.java
@@ -27,7 +27,6 @@ import org.openrewrite.yaml.tree.Yaml;
 
 import java.nio.file.Path;
 import java.util.List;
-import java.util.regex.Pattern;
 
 /**
  * A recipe to uniformly add a property to Spring configuration file. This recipe supports adding properties to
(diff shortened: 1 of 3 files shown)


Jenson3210 (approved):
🙏 thanks! This was on my todo list still.

PR #1003, opened 2026-04-20
Remove AddConfigurationAnnotationIfBeansPresent  from SpringBoot2BestPractices (fixes #1002)

### LINKED ISSUE #1002: "Spring Boot 2.x best practices" recipe  breaks Feign clients
“Add missing @Configuration annotation” migration (org.openrewrite.java.spring.boot2.AddConfigurationAnnotationIfBeansPresent ) added to “Spring Boot 2.x best practices” (org.openrewrite.java.spring.boot2.SpringBoot2BestPractices)  breaks Feign client configuration. 
As documented in https://docs.spring.io/spring-cloud-openfeign/reference/spring-cloud-openfeign.html#spring-cloud-feign-overriding-defaults, configuration classes containing beans specific to Feign clients doesn’t need to be annotated with @Configuration. This way the scope of these beans will be limited to specific Feign client only. On the other hand, if @Configuration is added, it will cause the beans to be registered in global application scope, which might not be correct setup.

Please remove “Add missing @Configuration annotation” from “Spring Boot 2.x best practices” to avoid issues caused by adding this annotation where it should not be present. 

I am using

- OpenRewrite v1.2.3
- Maven plugin v6.36.0
- rewrite-spring v6.28.0

I am using the Maven plugin, and my project is a single module project. I'm running it with following command:

```
mvn org.openrewrite.maven:rewrite-maven-plugin:run "-Drewrite.recipeArtifactCoordinates=org.openrewrite.recipe:rewrite-spring:6.28.0" -Drewrite.activeRecipies=org.openrewrite.java.spring.boot3.UpgradeSpringBoot_3_5
```

```java
class A {
    @Bean
    String foo() {
        return "foo";
    }
}
```

No changes to this class

## What did you see instead?

```java
@Configuration
class A {
    @Bean
    String foo() {
        return "foo";
    }
}
```

## Are you interested in [contributing a fix to OpenRewrite]

Yes, I can make the change.

kosmaty:
- Fix for #1002

## What's changed?
`AddConfigurationAnnotationIfBeansPresent` has been removed from `AddConfigurationAnnotationIfBeansPresent`.  

## What's your motivation?
“Add missing @Configuration annotation” migration (`org.openrewrite.java.spring.boot2.AddConfigurationAnnotationIfBeansPresent` ) added to “Spring Boot 2.x best practices” (`org.openrewrite.java.spring.boot2.SpringBoot2BestPractices`)  breaks Feign client configuration. 
As documented in https://docs.spring.io/spring-cloud-openfeign/reference/spring-cloud-openfeign.html#spring-cloud-feign-overriding-defaults, configuration classes containing beans specific to Feign clients doesn’t need to be annotated with `@Configuration`. This way the scope of these beans will be limitted to specific Feign client only. On the other hand, if `@Configuration` is added, it will cause the beans to be registered in global application scope, which might not be correct setup.

@timtebeek this partially reverts your change done in https://github.com/openrewrite/rewrite-spring/commit/f449874acf0a2454fe08fd13735f3403897ac40b - please make your opinion on that.

### DIFF
--- a/src/main/resources/META-INF/rewrite/best-practices.yml
+++ b/src/main/resources/META-INF/rewrite/best-practices.yml
@@ -31,7 +31,6 @@ recipeList:
 #  - org.openrewrite.java.spring.ImplicitWebAnnotationNames
   - org.openrewrite.java.spring.boot2.UnnecessarySpringExtension
   - org.openrewrite.java.spring.NoAutowiredOnConstructor
-  - org.openrewrite.java.spring.boot2.AddConfigurationAnnotationIfBeansPresent
   - org.openrewrite.java.spring.boot2.RestTemplateBuilderRequestFactory
   - org.openrewrite.java.spring.boot2.ReplaceDeprecatedEnvironmentTestUtils
   - org.openrewrite.java.spring.SeparateApplicationPropertiesByProfile


timtebeek:
Hi! Rather than take out the whole recipe, is there a way we can recognize the cases where you don't want to see changes and avoid them in the first place? I'd need a bit more clarity on when (not) to make a change.

kosmaty:
When using spring-cloud-openfeign, a configuration for specific feign client can be defined like this:

```java
import feign.RequestInterceptor;
import feign.auth.BasicAuthRequestInterceptor;
import org.springframework.context.annotation.Bean;

public class MyFeignConfiguration {
    @Bean
    public RequestInterceptor authInterceptor() {
        return new BasicAuthRequestInterceptor("user", "pass");
    }
    // auth request interceptor is common use case, but other Feign-related beans can be defined as well
}
```
and it's referenced in Feign client:
```java
import org.springframework.cloud.openfeign.FeignClient;

@FeignClient(
        name = "my-service",
        configuration = MyFeignConfiguration.class)
public class MyFeignClient {
    // feign client methods omitted
}
```
If we can recognize that configuration class is referenced in `@FeignClient` annotation and avoid adding `@Configuration` annotation to it, it would be great.

timtebeek:
Thanks! Based on your feedback I'm considering this alternative:
 - https://github.com/openrewrite/rewrite-spring/pull/1007

Happy to hear your thoughts.

kosmaty:
Thanks, it looks good to me, however I'm not an expert on recipes.

kosmaty:
Closing this in favor of #1007

PR #1006, opened 2026-04-23
Fix generic type loss in AutowiredFieldIntoConstructorParameterVisitor (fixes #1005)

### LINKED ISSUE #1005: AutowiredFieldIntoConstructorParameterVisitor loses generic type parameters

- rewrite-spring v6.30.0-SNAPSHOT (latest `main` branch)

I am using `AutowiredFieldIntoConstructorParameterVisitor` programmatically as part of a custom recipe. The issue is in the visitor itself, not in plugin configuration.

Input class with a parameterized `@Autowired` field and an existing constructor:

```java
package demo;

import java.util.List;
import org.springframework.beans.factory.annotation.Autowired;

public class A {

    @Autowired
    private List<String> a;

    A() {
    }
}
```

Invoked via:
```java
new AutowiredFieldIntoConstructorParameterVisitor("demo.A", "a")
```

```java
package demo;

import java.util.List;

public class A {

    private final List<String> a;

    A(List<String> a) {
        this.a = a;
    }
}
```

## What did you see instead?

```java
package demo;

import java.util.List;

public class A {

    private final List<String> a;

    A(List a) {
        this.a = a;
    }
}
```

The constructor parameter uses the raw type `List` instead of `List<String>`. This happens because both `AddConstructorVisitor` and `AddConstructorParameterAndAssignment` use `FullyQualified.getClassName()` to render the type string for the `JavaTemplate`, which only returns the raw class name without type parameters.

The same issue affects nested generics (`Map<String, List<Integer>>`), wildcards (`List<?>`), and bounded wildcards (`List<? extends Number>`, `List<? super Integer>`).

## What is the full stack trace of any errors you encountered?

No errors — the recipe completes successfully but produces incorrect output with raw types.

## Are you interested in [contributing a fix to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#contributing-fixes)?

Yes. I have an implementation idea ready that recursively renders the simple type name (including type parameters, wildcards, and bounded wildcards) from `JavaType`, along with test coverage for all the above cases. Happy to open a PR if this approach is welcome.

dim-kod:
- Fixes issue #1005

## What's changed?

`AutowiredFieldIntoConstructorParameterVisitor` now preserves generic type parameters when converting `@Autowired` fields to constructor parameters.

Both `AddConstructorVisitor` and `AddConstructorParameterAndAssignment` were using `TypeUtils.asFullyQualified(type).getClassName()` to render the field type in the generated constructor, which strips generics — so `List<String>` ends up as raw `List`.

The fix switches to `TypeTree.toString()`, which goes through `JavaPrinter` and gives you the full source-level type string. I also added `addImportsForType`/`collectFullyQualifiedNames` helpers to register type parameter FQNs with the `JavaTemplate` builder.

A second commit broadens the type guard from `TypeUtils.asFullyQualified() == null` to `null || Primitive` — this fixes a separate issue where array-typed fields like `String[] a` were silently skipped during constructor generation.

## What's your motivation?

Any `@Autowired` field with a parameterized type ends up with a raw-type constructor parameter after the recipe runs, which produces compiler warnings and loses type safety.

- The `TypeTree.toString()` approach — it uses `JavaPrinter.printTrimmed()` under the hood and already handles all type forms (parameterized, wildcards, bounded wildcards, nested generics, arrays).
- The broadened type guard (`Primitive` check instead of `asFullyQualified == null`) — this intentionally lets `JavaType.Array` through.

- A custom `renderTypeSimple(JavaType)` that recursively builds the simple-name type string from `JavaType` — works, but duplicates what `TypeTree.toString()` already does.
- `TypeUtils.toString(JavaType)` — produces fully qualified names (e.g. `java.util.List<java.lang.String>`), not source-level simple names, so it doesn't work for template strings.

## Any additional context

Test coverage went from 11 to 22 tests. New tests cover:
- Generic types into new/existing/parameterized constructors (`List<String>`)
- Nested generics (`Map<String, List<Integer>>`)
- Wildcards: unbounded (`List<?>`), upper-bounded (`List<? extends Number>`), lower-bounded (`List<? super Integer>`)
- User-defined generic types across packages (`MyService<MyConfig>`)
- Array types (`String[]`)
- `@Autowired(required = false)`
- Multiple annotations (`@Autowired @Qualifier("myBean")`)

### Checklist


dim-kod:
Hello @timtebeek, any updates?

timtebeek:
Hi no just need some more time for a proper review and merge; but with Devoxx UK this week, and Code Remix Summit next week it can take a bit longer to run through. Thanks for the work and patience in seeing this through!

dim-kod:
Understood, thanks.

MBoegers:
Hey @dim-kod, thanks for your analysis on this one. I took your branch and evolved it by using more built-in utilities. Do you mind reviewing the result? https://github.com/openrewrite/rewrite-spring/pull/1037
If you are okay with the proposal you can close this PR.

PR #1012, opened 2026-04-30
Migrate SpringFox `SecurityConfiguration` bean to Springdoc properties

steve-aom-elliott:
## Summary

Adds a new recipe `MigrateSpringFoxSecurityConfiguration` that translates `@Bean` methods returning `springfox.documentation.swagger.web.SecurityConfiguration` into the equivalent `springdoc.swagger-ui.*` configuration properties on the existing `application.properties` or `application.yml` file. The bean and its now-unused imports are removed once the properties have been written.

The recipe is wired into the `SpringFoxToSpringDoc` chain after `SecurityContextToSecurityScheme`, closing a gap where SpringFox security UI configuration beans were left behind by the migration.

### Builder method to property mapping

| Builder method | Property |
| --- | --- |
| `clientId` | `springdoc.swagger-ui.oauth.client-id` |
| `clientSecret` | `springdoc.swagger-ui.oauth.client-secret` |
| `realm` | `springdoc.swagger-ui.oauth.realm` |
| `appName` | `springdoc.swagger-ui.oauth.app-name` |
| `scopeSeparator` | `springdoc.swagger-ui.oauth.scope-separator` |
| `useBasicAuthenticationWithAccessCodeGrant` | `springdoc.swagger-ui.oauth.use-basic-authentication-with-access-code-grant` |
| `enableCsrfSupport` | `springdoc.swagger-ui.csrf.enabled` |

### Conservative semantics

- Only literal builder arguments are migrated; chains with non-literal arguments are left untouched for manual review.
- Unsupported builder methods (`apiKey`, `apiKeyName`, `apiKeyVehicle`, `additionalQueryStringParams`) cause the chain to be skipped — Springdoc has no direct property equivalents.
- If no Spring application configuration file exists in the project, the bean is left in place rather than silently dropping configuration.

## Test plan

- [x] New tests in `MigrateSpringFoxSecurityConfigurationTest` cover the `enableCsrfSupport` literal case, multiple builder calls, YAML output, and the three opt-out scenarios (non-literal arg, unsupported method, non-`@Bean` method).
- [x] Existing `spring.doc.*` and `springdoc.*` tests continue to pass.
- [x] `recipes.csv` regenerated via `./gradlew recipeCsvGenerate`.


steve-aom-elliott:
Seems green now following the fix in `openrewrite/rewrite`

timtebeek (approved):
I've made some changes, mostly to support multi module projects and to use `service(AnnotationService.class)`. Ok to merge if you agree.

PR #1015, opened 2026-05-06
Wrap `List` arguments to `ItemWriter.write` call sites in `new Chunk<>(...)`

Jenson3210:
## Summary

Closes the call-site gap in `MigrateItemWriterWrite` for the Spring Batch 4→5 migration.

`MigrateItemWriterWrite` migrates `ItemWriter.write(List<T>)` to `write(Chunk<? extends T>)` and rewrites usages of the **outer parameter** inside the method body, but it does not touch invocations on **other** `ItemWriter` instances. After the recipe runs, calls like `this.someWriter.write(list)` still pass a `List` to a method that now expects a `Chunk`, leaving the migrated code uncompilable.

This PR adds a sibling recipe `org.openrewrite.java.spring.batch.MigrateItemWriterWriteCalls`, chained immediately after `MigrateItemWriterWrite` in `SpringBatch4To5Migration`, that wraps the `List` argument of any `ItemWriter.write` invocation in `new Chunk<>(...)`. Arguments already typed `Chunk` are skipped, so the parameter-passthrough case (e.g. `delegate.write(items)` inside the migrated `write(Chunk<T> items)`) is not double-wrapped.

## Implementation notes

Matching uses `simpleName` + `select` type rather than `MethodMatcher`. `MigrateItemWriterWrite` reparses the migrated method body with the Spring Batch 5.1 classpath via `JavaTemplate`; when an inner `someWriter.write(list)` argument doesn't fit `write(Chunk)`, the call's resolved `methodType` is left null, so a `MethodMatcher` keyed on either signature would miss it. A name + receiver-type check stays robust regardless of which recipe runs first or what the resolved `methodType` looks like.

## Customer request

- Closes [moderneinc/customer-requests#2334](https://github.com/moderneinc/customer-requests/issues/2334)

## Test plan

- [x] New `MigrateItemWriterWriteCallsTest` covers six scenarios:
  1. `wrapsListPassedToWriter` (`@DocumentExample`) — basic field-typed `ItemWriter` + `List` argument.
  2. `wrapsListLocalVariableFromForLoop` — the customer's exact pattern.
  3. `wrapsArrayListSubtype` — `ArrayList` argument (subtype of `List`).
  4. `doesNotWrapWhenReceiverIsNotItemWriter` — non-`ItemWriter` receiver is skipped.
  5. `combinedRecipesProduceFullyMigratedCustomerCode` — runs `MigrateItemWriterWrite` + `MigrateItemWriterWriteCalls` together against the customer pattern and verifies the fully-migrated, compiling result.
  6. `combinedRecipesDoNotDoubleWrapPassthrough` — verifies the parameter-passthrough case isn't double-wrapped.
- [x] Existing `MigrateItemWriterWriteTest` still passes (no regression).
- [x] `recipeCsvGenerate` ran clean and `recipeCsvValidateContent` passed.
- [x] Three pre-existing `MigrateJobBuilderFactoryTest` failures reproduce on unmodified `main` and are unrelated to this change.


src/main/java/org/openrewrite/java/spring/batch/MigrateItemWriterWriteCalls.java
```diff
@@ -0,0 +1,77 @@
+/*
+ * Copyright 2024 the original author or authors.
```
steve-aom-elliott: Copyright year.

Jenson3210:
```suggestion
 * Copyright 2026 the original author or authors.
```

src/main/java/org/openrewrite/java/spring/batch/MigrateItemWriterWriteCalls.java
```diff
@@ -0,0 +1,77 @@
...
+                Expression select = mi.getSelect();
+                if (select == null || !TypeUtils.isAssignableTo(ITEM_WRITER_FQN, select.getType())) {
+                    return mi;
+                }
+                if (TypeUtils.isAssignableTo(CHUNK_FQN, arg.getType())) {
+                    return mi;
+                }
+                if (!TypeUtils.isAssignableTo(LIST_FQN, arg.getType())) {
+                    return mi;
+                }
```
steve-aom-elliott: Any particular reason to not just find `org.springframework.batch.item.ItemWriter write(List)` using a method matcher rather than checking each of the pieces?

src/test/java/org/openrewrite/java/spring/batch/MigrateItemWriterWriteCallsTest.java
```diff
@@ -0,0 +1,304 @@
+/*
+ * Copyright 2024 the original author or authors.
```
steve-aom-elliott: Copyright year here as well

src/main/java/org/openrewrite/java/spring/batch/MigrateItemWriterWriteCalls.java
Jenson3210: After `MigrateItemWriterWrite` runs first in the chain, it reparses the migrated method body via `JavaTemplate` with the 5.1 classpath. In that reparse, `ItemWriter.write(List)` no longer exists (5.1 has `write(Chunk)`) and `list` is still typed `List` (not `Chunk`), so the inner call's resolved `methodType` ends up `null`. 

Pushed an inline comment in the recipe so the rationale is visible at the call site.

steve-aom-elliott: A bit disappointing in that case. Almost would have wished they would be done in the same recipe to avoid that scenario, but alright.

Jenson3210: I am working on that change now indeed. But he's pushing not to do that due to the preconditions there. Think that could be bypassed.

Jenson3210:
Took a second pass on @steve-aom-elliott's `MethodMatcher` question — folded the call-site wrapping into `MigrateItemWriterWrite` itself instead of keeping it as a sibling recipe (commit 05bdf14d).

Because `super.visitMethodDeclaration` walks children *before* the body reparse, the inner call's `methodType` is still intact at the moment my new `visitMethodInvocation` runs, so a plain `MethodMatcher` does the job there — no `simpleName`/select-type workaround. Passthrough is handled by skipping when the argument resolves to the parameter of the enclosing `ItemWriter.write(List)` declaration.

Net change: dropped `MigrateItemWriterWriteCalls.java` (and its YAML chain entry), widened `MigrateItemWriterWrite`'s precondition to `Preconditions.or(DeclaresMethod, UsesMethod)`, moved the six tests into `MigrateItemWriterWriteTest` (renamed the previously-chained ones), and regenerated `recipes.csv`. PTAL.

steve-aom-elliott (approved):

PR #1018, opened 2026-05-15
Allow opting additional YAML/properties files into Spring property recipes

steve-aom-elliott:
## What's changed?

Adds an opt-in mechanism for treating YAML/properties files as Spring configuration when they live outside a recognized source set such as `src/main/resources`.

- New `SpringConfigFile` presence marker.
- New `MarkAdditionalSpringConfigFiles` recipe accepting a list of glob patterns (using `PathUtils.matchesGlob`, so `!` negation and `{a,b}` alternation work). It stamps `SpringConfigFile` onto matching `Properties.File` / `Yaml.Documents` source files that are not already covered by a `SourceSet` marker.
- `IsPossibleSpringConfigFile` now passes when either a `SourceSet` marker (current behavior) or a `SpringConfigFile` marker is present. The widening is purely additive — every file that passes today still passes.

- This makes the existing guard ([#976](https://github.com/openrewrite/rewrite-spring/pull/976)) reachable for repositories whose Spring config files live in non-standard locations (e.g. `<service>/properties/base-prod.properties`), without re-opening the door to clobbering GitHub Actions workflows, Docker Compose, Helm charts, and other YAML/properties files that share keys with Spring config.

## What's your motivation?

Today `ChangeSpringPropertyKey`, `ChangeSpringPropertyValue`, `DeleteSpringProperty`, and `CommentOutSpringPropertyKey` are all gated by `IsPossibleSpringConfigFile`, which requires a `SourceSet` marker. Files outside recognized Maven/Gradle source sets never get the marker and are silently skipped. For teams that don't follow the standard Spring resource layout, this can mean a Spring Boot migration runs cleanly without touching files that genuinely contain Spring properties.

- Broadening the default precondition would regress [#976](https://github.com/openrewrite/rewrite-spring/pull/976), so an opt-in marker pathway is preferable: existing repos behave exactly as before, and non-standard layouts can be enabled with a single recipe prepended to the migration chain.

## Usage

```yaml
name: com.example.MigrateSpringBoot4
recipeList:
  - org.openrewrite.java.spring.MarkAdditionalSpringConfigFiles:
      pathPatterns:
        - "**/properties/*.properties"
        - "**/config/*.yml"
  - org.openrewrite.java.spring.boot4.UpgradeSpringBoot_4_0
```

One recipe entry propagates through every composed `Change/Delete/CommentOutSpringProperty*` invocation downstream — no per-recipe option threading needed.

- **Same-cycle handoff.** `MarkAdditionalSpringConfigFiles` is a plain `Recipe`, not a `ScanningRecipe`, so its marker edits land in `editSources` and are visible to subsequent recipes' `Preconditions.check(...)` evaluations in the same cycle. There is an integration test (`enablesChangeSpringPropertyKeyOnFileOutsideSourceSet`) asserting end-to-end that `MarkAdditionalSpringConfigFiles` composed with `ChangeSpringPropertyKey` performs the rename on a file outside `src/main/resources` in a single run.
- **Negative regression guard.** `changeSpringPropertyKeyAloneSkipsFileOutsideSourceSet` confirms `ChangeSpringPropertyKey` on its own still skips such files — preserving the existing default.
- **Source-set short-circuit.** The recipe explicitly skips files that already carry a `SourceSet` marker, so it does not add redundant markers to files the existing check would already let through.
- **Marker scope.** `SpringConfigFile` lives in `rewrite-spring`, not promoted to `rewrite-core`, so consumers that don't already pull in `rewrite-spring` won't need to deserialize it.

### Checklist


src/main/java/org/openrewrite/java/spring/MarkAdditionalSpringConfigFiles.java
```diff
@@ -0,0 +1,76 @@
...
+        return "Attach a `SpringConfigFile` marker to YAML/properties files matching the provided glob patterns " +
+                "so that Spring property recipes such as `ChangeSpringPropertyKey`, `DeleteSpringProperty`, " +
+                "`ChangeSpringPropertyValue`, and `CommentOutSpringPropertyKey` will visit files that live " +
+                "outside standard resource source sets. Files that already pass the `SourceSet`-based check are " +
+                "skipped to avoid redundant markers.";
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return new TreeVisitor<Tree, ExecutionContext>() {
```
timtebeek: Did you consider using a precondition here to match the path patterns? Figured that might reduce the need for separate matching below.

timtebeek (approved):
Seems reasonable to have them opt-in certain files to avoid false positives and unintended changes resulting from those. Thanks!

src/main/java/org/openrewrite/java/spring/MarkAdditionalSpringConfigFiles.java
steve-aom-elliott: Fair point. I guess I couldn't remember an existing visitor precondition for something like that, but I'll see what we have / make a small one if we don't have one.

steve-aom-elliott: Ah, `FindSourceFiles`. Working on switching over now.

PR #1022, opened 2026-05-18
Pass cursor to `RemoveAnnotationVisitor` in `AutowiredFieldIntoConstructorParameterVisitor`

Jenson3210:
## Summary
- The upstream `RemoveAnnotationVisitor` in `rewrite-java` now relies on its cursor context (e.g. for `maybeRemoveBlankLines` to locate the enclosing `JavaSourceFile`). The previous two-argument `visitNonNull(tree, ctx)` call left it without a cursor, breaking `AutowiredFieldIntoConstructorParameterVisitorTest` (7 of 11 tests failing with `Expected to find enclosing JavaSourceFile`).
- Switch to the three-argument `visit(tree, ctx, cursor)` form. Because we're re-visiting `multiVariable` — the tree the current cursor already points to — we pass `getCursor().getParentOrThrow()` to avoid the `cursorAcyclic` assertion. This mirrors the pattern already used in `NoAutowiredOnConstructor`.

## Test plan
- [x] `./gradlew test --tests org.openrewrite.java.spring.AutowiredFieldIntoConstructorParameterVisitorTest` → all 11 tests pass

### DIFF
--- a/src/main/java/org/openrewrite/java/spring/AutowiredFieldIntoConstructorParameterVisitor.java
+++ b/src/main/java/org/openrewrite/java/spring/AutowiredFieldIntoConstructorParameterVisitor.java
@@ -118,7 +118,7 @@ public class AutowiredFieldIntoConstructorParameterVisitor extends JavaVisitor<E
                 multiVariable.getVariables().size() == 1 &&
                 fieldName.equals(multiVariable.getVariables().get(0).getName().getSimpleName())) {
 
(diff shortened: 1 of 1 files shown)


Jenson3210:
Also, after https://github.com/openrewrite/rewrite/pull/7701/changes#diff-960b8aebf2f2a16a80aa08fefa07a42a1f81c74e04c2ce2762897be4cb2e6781R115 change, we need to have a valid (up to SourceFile) cursor sent.

timtebeek (approved):

PR #1026, opened 2026-05-26
Emit Kotlin array literal for single-element `@MockBean(types = ...)`

MBoegers:
## Summary

`UnwrapMockAndSpyBeanContainers` previously emitted the Java single-element annotation-array shorthand (`types = Foo::class`) when a `@MockBeans` / `@SpyBeans` container collapsed to one inner annotation, regardless of source language. Kotlin rejects that shape:

> Argument type mismatch: actual type is `KClass<Foo>`, but `Array<KClass<*>>` was expected.

The fix restricts the shorthand path to Java and lets the existing Kotlin `K.ListLiteral` branch handle `size == 1`. Java output is unchanged.

Before / after (Kotlin):

```diff
-@MockBean(types = Foo::class)
+@MockBean(types = [Foo::class])
```

- PR #1023 made the multi-element Kotlin case correct; this PR completes the single-element case.

- Resolves moderneinc/customer-requests#2435.

## Test plan

- [x] Updated `kotlinSingleInnerAnnotation` to assert the array-literal form — fails without the fix.
- [x] Added `javaSingleInnerAnnotationKeepsShorthand` confirming Java shorthand is preserved.
- [x] `./gradlew test --tests UnwrapMockAndSpyBeanContainersTest` — 8/8 pass (3 Java + 5 Kotlin).

## Scope note

This PR is intentionally narrow to the recipe that produced the reported failure; a broader sweep would be a separate piece of work.

### DIFF
--- a/src/main/java/org/openrewrite/java/spring/boot4/UnwrapMockAndSpyBeanContainers.java
+++ b/src/main/java/org/openrewrite/java/spring/boot4/UnwrapMockAndSpyBeanContainers.java
@@ -188,10 +188,7 @@ public class UnwrapMockAndSpyBeanContainers extends Recipe {
         );
 
         Expression rhs;
-        if (typeExpressions.size() == 1 && !kotlin) {
-            // Java single-element shorthand: types = Foo.class
-            rhs = typeExpressions.get(0).withPrefix(Space.SINGLE_SPACE);
-        } else if (kotlin) {
+        if (kotlin) {
             // Format as [A::class, B::class]
             List<JRightPadded<Expression>> padded = new ArrayList<>();
             for (int i = 0; i < typeExpressions.size(); i++) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/openrewrite/java/spring/boot4/UnwrapMockAndSpyBeanContainers.java
```diff
@@ -188,7 +188,8 @@ private static J.Assignment createTypesAssignment(List<Expression> typeExpressio
         );
 
         Expression rhs;
-        if (typeExpressions.size() == 1) {
+        if (typeExpressions.size() == 1 && !kotlin) {
+            // Java single-element shorthand: types = Foo.class
             rhs = typeExpressions.get(0).withPrefix(Space.SINGLE_SPACE);
         } else if (kotlin) {
```
timtebeek: Nit: if we place `if (isKotlin)` first then we wouldn't need to check that condition twice, and it's clearer that the other cases are both for Java with 1 or multiple args.

timtebeek (approved):

src/main/java/org/openrewrite/java/spring/boot4/UnwrapMockAndSpyBeanContainers.java
MBoegers: agree.

PR #1032, opened 2026-05-29
Add recipe to migrate Tomcat LegacyCookieProcessor to Rfc6265CookieProcessor

timtebeek:
Adds `org.openrewrite.java.spring.boot3.UseRfc6265CookieProcessor`, a declarative recipe that rewrites the `className` attribute of the Tomcat `<CookieProcessor>` element from `org.apache.tomcat.util.http.LegacyCookieProcessor` to `org.apache.tomcat.util.http.Rfc6265CookieProcessor` in Tomcat configuration files (e.g. `context.xml`, `server.xml`).

```xml
<Context>
    <CookieProcessor className="org.apache.tomcat.util.http.LegacyCookieProcessor"/>
</Context>
```
becomes
```xml
<Context>
    <CookieProcessor className="org.apache.tomcat.util.http.Rfc6265CookieProcessor"/>
</Context>
```

`Rfc6265CookieProcessor` has been the Tomcat default since 8.5; `LegacyCookieProcessor` is retained only for backwards compatibility. The recipe is intentionally **not** wired into any version upgrade because RFC 6265 parsing is stricter than the legacy behavior, so it is offered as an opt-in change. The `oldValue` guard ensures only the legacy class is rewritten, leaving custom or already-migrated cookie processors untouched (covered by tests).

- Reported via customer feedback (moderneinc/customer-requests#2462).

### DIFF
--- a/src/main/resources/META-INF/rewrite/tomcat.yml
+++ b/src/main/resources/META-INF/rewrite/tomcat.yml
@@ -1,0 +1,34 @@
+#
+# Copyright 2026 the original author or authors.
+# <p>
+# Licensed under the Moderne Source Available License (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+# <p>
+# https://docs.moderne.io/licensing/moderne-source-available-license
+# <p>
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


Jenson3210 (commented):
Don't you think ChangeType should be used?

https://github.com/moderneinc/customer-requests/issues/2462#issuecomment-4574499523

Jenson3210 (commented):
Should this one not be created in rewrite-java-application-server or moderne-rewrite-spring?

PR #1061, opened 2026-07-06
Render moved `@Autowired` field types as written instead of fully qualified

MBoegers:
### What's changed?

When `AutowiredFieldIntoConstructorParameterVisitor` promotes an `@Autowired` field into a constructor parameter, it rendered the parameter type from the resolved `JavaType` (fully qualified) and relied on a follow-up `ImportService#shortenFullyQualifiedTypeReferencesIn` step to collapse it and add imports. That shortening needs type attribution the template parser can't produce for third-party types, so it silently no-op'd — leaving fully-qualified names, including every generic type argument, in the generated constructor (noisy diffs that deviate from the surrounding style).

This renders the parameter type from the field's own `TypeTree` ("print as written") instead: simple names stay valid through the file's existing imports, deliberately fully-qualified references stay fully qualified, and generic type arguments are preserved verbatim. It also uses the post-`RemoveAnnotationVisitor` type expression, so an `@Autowired` written between the modifier and the type (which the parser folds into the field's `J.AnnotatedType`) is not re-emitted onto the generated parameter — which would otherwise reference the now-removed import.

### Testing

- `promotesAutowiredFieldIntoExistingConstructor` — a generic field (`Converter<Jwt, AbstractAuthenticationToken>`) is promoted with simple names + imports intact.
- `keepFQNsInUse` — a field declared with a fully-qualified type keeps the FQN on the parameter.
- `annotationBetweenModifierAndType` — an inline `@Autowired` is not carried onto the parameter.

### Depends on

- openrewrite/rewrite#8183 — `replaceParameters` templates need the source file's imports to attribute simple type names. Until that is merged, released, and `rewriteVersion` is bumped here, the tests exercising generated parameters fail in CI, so this shouldn't merge before then.


src/test/java/org/openrewrite/java/spring/AutowiredFieldIntoConstructorParameterVisitorTest.java
```diff
@@ -794,4 +785,133 @@ public class A {
...
+                  }
+
+              }
+              """,
+            """
+              package demo;
+
+              public class A {
+
+                  private final  String a;
```
timtebeek: I see we still have two spaces between `final. String`. Can we fix that? As it's likely to trip up formatter checks and pedantic reviewers like myself. 🙃 

```suggestion
                  private final String a;
```

timtebeek (approved):

PR #1065, opened 2026-07-10
Add parameterized `RemoveAutoConfigurationExclude` recipe

### LINKED ISSUE #1064: Add a parameterized `RemoveAutoConfigurationExclude` recipe
### Problem

`org.openrewrite.java.spring.boot3.RemoveSolrAutoConfigurationExclude` already implements the exact mechanical work of removing a class from `@SpringBootApplication` / `@EnableAutoConfiguration`'s `exclude` attribute — dropping the attribute entirely when the array empties, and cleaning up the now-unused import. But it is hardcoded to `SolrAutoConfiguration` and only handles the `exclude` (class-literal) attribute, not `excludeName` (string-FQN) form.

Other migrations (and one-off cleanups) need the same behavior for arbitrary auto-configuration classes and don't have a general-purpose recipe to call.

### Proposal

Add a parameterized recipe, e.g. `org.openrewrite.java.spring.RemoveAutoConfigurationExclude`, with:

- **`fullyQualifiedName`** (required, `String`) — the FQN of the auto-configuration class to remove from the exclusion.
- Behavior:
  - Handles both `@SpringBootApplication` and `@EnableAutoConfiguration`.
  - Removes the entry from `exclude` (class literal) **and** `excludeName` (string form).
  - Removes the attribute when the resulting array is empty.
  - Removes the now-unused import.
  - No-op when the annotation doesn't reference the given class.

### Migration of the existing recipe

Suggest keeping `RemoveSolrAutoConfigurationExclude` as a thin delegate (declarative YAML that pins `fullyQualifiedName` to `org.springframework.boot.autoconfigure.solr.SolrAutoConfiguration`) so existing consumers keep working. Happy to instead deprecate it in favor of the parameterized form if that's preferred.

### Notes

- `excludeName` support is a real addition beyond what the Solr recipe does today, and worth having so this recipe covers 100% of exclusion syntaxes.
- Naming open to bikeshedding — `RemoveAutoConfigurationExclude` mirrors the existing recipe's naming.

steve-aom-elliott:
- Closes #1064.

Generalizes the mechanical work of `RemoveSolrAutoConfigurationExclude` into a new, parameterized recipe at `org.openrewrite.java.spring.RemoveAutoConfigurationExclude` with a required `fullyQualifiedName` option. Adds support for the `excludeName` (string-FQN) attribute — both single value and array form — in addition to the existing `exclude` (class-literal) handling.

`RemoveSolrAutoConfigurationExclude` becomes a thin delegate that pins `fullyQualifiedName` to the Solr auto-config FQN via `getRecipeList()`, so existing consumers of the Solr-specific recipe are unaffected. Its tests are unchanged and pass against the delegate.

### What the new recipe covers
- `@SpringBootApplication` and `@EnableAutoConfiguration`.
- `exclude = FooAutoConfiguration.class`, `exclude = { A.class, B.class }`, and fully-qualified `exclude = com.foo.FooAutoConfiguration.class` forms.
- `excludeName = "com.foo.FooAutoConfiguration"` and the array form.
- Mixed `exclude` + `excludeName` on the same annotation.
- Removes the attribute entirely when the resulting array is empty; removes the now-unused import.

### Tests
- New: `org.openrewrite.java.spring.RemoveAutoConfigurationExcludeTest` covers `exclude` (array/single/FQN), `excludeName` (single/array), both annotations, mixed usage, and no-op cases for unrelated auto-configuration classes.
- Existing: `RemoveSolrAutoConfigurationExcludeTest` continues to pass against the delegate.

### Notes
- `recipes.csv` regenerated via `./gradlew recipeCsvGenerate`.
- `excludeName` support is a real behavioral addition beyond what the Solr recipe covered — worth calling out for reviewers.


timtebeek (approved):
Great to see, thanks!

PR #1071, opened 2026-07-16
Preserve Spring Boot managed Jackson versions during Boot 4 migration

### LINKED ISSUE #1070: Spring Boot 4 migration introduces explicit Jackson dependency versions instead of preserving Spring Boot managed dependency properties

I am using:

- rewrite-maven-plugin 6.44.0
- rewrite-spring 6.35.0
- Maven
- Java 25

I am using the Maven plugin on a single-module Spring Boot project.

Recipe:

```yaml
---
name: example.SpringBoot4Migration

recipeList:
  - org.openrewrite.java.spring.boot4.UpgradeSpringBoot_4_0
```

Plugin:

```xml
<plugin>
    <groupId>org.openrewrite.maven</groupId>
    <artifactId>rewrite-maven-plugin</artifactId>
    <version>6.44.0</version>

    <configuration>
        <activeRecipes>
            <recipe>example.SpringBoot4Migration</recipe>
        </activeRecipes>
    </configuration>

    <dependencies>
        <dependency>
            <groupId>org.openrewrite.recipe</groupId>
            <artifactId>rewrite-spring</artifactId>
            <version>6.35.0</version>
        </dependency>
    </dependencies>
</plugin>
```

The recipe is executed using:

```bash
mvn rewrite:dryRun
```

Start with a Spring Boot 3.5.x project using the Spring Boot parent and overriding the managed Jackson BOM version:

```xml
<parent>
    <artifactId>spring-boot-starter-parent</artifactId>
    <version>3.5.14</version>
</parent>

<properties>
    <jackson-bom.version>2.21.1</jackson-bom.version>
</properties>

<dependencies>
    <dependency>
        <groupId>com.fasterxml.jackson.core</groupId>
        <artifactId>jackson-databind</artifactId>
    </dependency>

    <dependency>
        <groupId>com.fasterxml.jackson.dataformat</groupId>
        <artifactId>jackson-dataformat-xml</artifactId>
    </dependency>
</dependencies>
```

Run:

```yaml
recipeList:
  - org.openrewrite.java.spring.boot4.UpgradeSpringBoot_4_0
```

The Spring Boot 4 migration already includes the Jackson 2 → Jackson 3 migration through the Spring Framework 7 upgrade.

I expected the migration to preserve Spring Boot's managed dependency model.

If the recipe upgrades Jackson to a newer version (for example 3.1.5), I expected one of the following:

- the managed Spring Boot property (`jackson-bom.version`) to be updated, while leaving Jackson dependencies versionless, or
- the existing override to be migrated appropriately (for example to `jackson-2-bom.version` if it represents a Jackson 2 compatibility override).

For example:

```xml
<properties>
    <jackson-bom.version>3.1.5</jackson-bom.version>
</properties>

<dependency>
    <groupId>tools.jackson.core</groupId>
    <artifactId>jackson-databind</artifactId>
</dependency>

<dependency>
    <groupId>tools.jackson.dataformat</groupId>
    <artifactId>jackson-dataformat-xml</artifactId>
</dependency>
```

In other words, Spring Boot dependency management should continue to control Jackson versions rather than introducing explicit versions onto every Jackson dependency.

## What did you see instead?

The dependency coordinates were migrated correctly to Jackson 3, however explicit versions were introduced onto each dependency.

For example:

```xml
<dependency>
    <groupId>tools.jackson.core</groupId>
    <artifactId>jackson-databind</artifactId>
    <version>3.1.5</version>
</dependency>
```

rather than updating the managed Spring Boot version property.

Additionally, during the migration Maven attempts to resolve the following invalid intermediate coordinate:

```
tools.jackson:jackson-bom:2.21.1
```

resulting in warnings similar to:

```
Failed to download tools.jackson:jackson-bom:pom:2.21.1
```

This appears to occur because the existing Spring Boot 3 `jackson-bom.version` property still contains the Jackson 2 version while the recipe has already migrated the BOM coordinates to Jackson 3.

## Why I believe this is a problem

Spring Boot manages Jackson through a BOM.

Once the migration has completed, the project no longer benefits from that dependency management because explicit versions have been written onto each Jackson dependency.

This means future Spring Boot upgrades will no longer automatically align the Jackson dependency versions.

It also appears inconsistent with how Spring Boot managed dependency overrides are normally represented (via version properties rather than dependency-level versions).

## Additional information

I appreciate there are multiple possible correct migration outcomes depending on whether the existing property is intended to override Jackson 2 compatibility or the primary Jackson BOM.

However, I would expect the migration to preserve the Spring Boot managed dependency model rather than introducing explicit dependency versions.

## Are you interested in contributing a fix to OpenRewrite?

Yes. I would be happy to help verify a fix or provide a minimal reproducer.

timtebeek:
- Fixes #1070

## Problem

Migrating a Spring Boot 3 project that overrides the Jackson BOM version to Spring Boot 4 produced two problems:

1. Explicit versions were written onto every migrated Jackson 3 dependency (e.g. `tools.jackson.core:jackson-databind:3.1.5`) instead of leaving them managed by Spring Boot's BOM.
2. The Spring Boot 3 `jackson-bom.version` override was left untouched, so Maven tried to resolve the invalid coordinate `tools.jackson:jackson-bom:2.21.1`.

## Root cause

In Spring Boot 3, the `jackson-bom.version` property overrides the **Jackson 2** BOM (`com.fasterxml.jackson:jackson-bom`). In Spring Boot 4, that same property controls the **Jackson 3** BOM (`tools.jackson:jackson-bom`, default `3.1.x`), while the Jackson 2 compatibility BOM is controlled by a new `jackson-2-bom.version` property:

```xml
<jackson-2-bom.version>2.21.2</jackson-2-bom.version>  <!-- com.fasterxml.jackson:jackson-bom -->
<jackson-bom.version>3.1.2</jackson-bom.version>       <!-- tools.jackson:jackson-bom -->
```

A carried-over `jackson-bom.version=2.21.1` therefore points the Jackson 3 BOM import at a non-existent version, breaking dependency management. With the BOM import failing, the `tools.jackson` dependencies are not seen as managed while the Jackson migration runs, so explicit versions get added.

## Fix

In `UpgradeSpringBoot_4_0` (`spring-boot-40.yml`), after the parent/BOM upgrade:

- Rename the Maven property `jackson-bom.version` → `jackson-2-bom.version`, preserving the user's Jackson 2 override intent and restoring a valid Jackson 3 BOM import.
- Remove now-redundant explicit `tools.jackson` versions so Spring Boot dependency management controls them again.

After the fix, the reproducer keeps Jackson dependencies versionless and migrates the override:

```xml
<properties>
    <jackson-2-bom.version>2.21.1</jackson-2-bom.version>
</properties>
<dependencies>
    <dependency>
        <groupId>tools.jackson.core</groupId>
        <artifactId>jackson-databind</artifactId>
    </dependency>
    <dependency>
        <groupId>tools.jackson.dataformat</groupId>
        <artifactId>jackson-dataformat-xml</artifactId>
    </dependency>
</dependencies>
```

## Tests

Added two tests to `UpgradeSpringBoot_4_0Test`: one covering the override-property reproducer from the issue, and one covering the common case (no override property) where dependencies should stay BOM-managed.


src/main/resources/META-INF/rewrite/spring-boot-40.yml
```diff
@@ -55,6 +55,14 @@ recipeList:
       groupId: org.springframework.boot
       artifactId: spring-boot-starter-parent
       newVersion: 4.0.x
+  # In Spring Boot 4 `jackson-bom.version` controls the Jackson 3 BOM; an existing override
+  # targets Jackson 2, so migrate it to `jackson-2-bom.version` to keep pointing at Jackson 2.
+  - org.openrewrite.maven.RenamePropertyKey:
+      oldKey: jackson-bom.version
```
Kyran121: This rename appears unconditional. On an already-Spring-Boot-4 POM with a legitimate `<jackson-bom.version>3.1.x</jackson-bom.version>` override, rerunning `UpgradeSpringBoot_4_0` would rename it to `jackson-2-bom.version`, causing the Jackson 2 BOM to point at a Jackson 3 version and dropping the intended Jackson 3 override.

Could this be gated to Boot 3.x input, or otherwise made conditional on the old property value/source model? A regression test for an already-Boot-4 POM with `jackson-bom.version=3.1.x` would protect idempotency. It may also be worth defining behavior when both `jackson-bom.version` and `jackson-2-bom.version` are already present.

Kyran121 (commented):
One idempotency concern before this merges.

PR #1078, opened 2026-07-17
Migrate jsonschema2pojo config during Spring Boot 4 upgrade

### LINKED ISSUE #1073: Migrate jsonschema2pojo configuration when upgrading to Spring Boot 4

When `UpgradeSpringBoot_4_0` is applied to a Spring Boot project using `jsonschema2pojo-maven-plugin`, OpenRewrite does not update the plugin version or its Jackson and Jakarta Validation generation options.

For example, an existing Spring Boot 3 project may have:

```xml
<properties>
    <jsonschema2pojo.version>1.2.2</jsonschema2pojo.version>
</properties>

<plugin>
    <groupId>org.jsonschema2pojo</groupId>
    <artifactId>jsonschema2pojo-maven-plugin</artifactId>
    <version>${jsonschema2pojo.version}</version>
    <configuration>
        <annotationStyle>jackson2</annotationStyle>
        <includeJsr303Annotations>true</includeJsr303Annotations>
    </configuration>
</plugin>
```

After the Spring Boot 4 migration:

- the plugin version may remain too old to support Jackson 3 annotation generation;
- `annotationStyle` may remain set to `jackson2`, or may be absent and therefore continue using the `jackson2` default;
- generated validation annotations may continue using `javax.validation` rather than `jakarta.validation`.

This can cause regenerated model classes to remain incompatible with the containing Spring Boot 4 application.

jsonschema2pojo `1.3.0` added the `jackson3` annotation style. However, the plugin continues to default `annotationStyle` to `jackson2`.

The plugin also provides `useJakartaValidation`, which switches generated JSR-303 validation annotations from `javax.validation` to `jakarta.validation`. This option only has an effect when `includeJsr303Annotations=true`.

The Spring Boot 4 migration should therefore ensure that the plugin supports Jackson 3 generation and update the applicable generator configuration.

The transformation should apply when:

- the Spring Boot 4 upgrade recipe is being applied;
- `org.jsonschema2pojo:jsonschema2pojo-maven-plugin` is present.

The recipe should resolve the configured jsonschema2pojo version and:

- upgrade versions below `1.3.0` to at least `1.3.0`;
- preserve versions already equal to or newer than `1.3.0`;
- avoid applying `annotationStyle=jackson3` until a compatible version has been established.

The Jackson configuration should be migrated when:

- `annotationStyle` is `jackson`;
- `annotationStyle` is `jackson2`; or
- `annotationStyle` is absent, because the plugin default is `jackson2`.

The recipe should not replace deliberately selected non-Jackson styles such as:

- `gson`;
- `jsonb`;
- `jsonb2`;
- `moshi1`;
- `none`;
- a custom annotation strategy.

The Jakarta Validation configuration should only be added or changed when JSR-303 annotation generation is active, for example when:

```xml
<includeJsr303Annotations>true</includeJsr303Annotations>
```

This issue is intentionally limited to the Maven plugin within a Spring Boot 4 migration.

Gradle plugin support and generic jsonschema2pojo Jackson or Jakarta migrations outside Spring Boot could be handled separately.

The plugin version may be configured through a Maven property:

```xml
<properties>
    <jsonschema2pojo.version>1.2.2</jsonschema2pojo.version>
</properties>

<plugin>
    <groupId>org.jsonschema2pojo</groupId>
    <artifactId>jsonschema2pojo-maven-plugin</artifactId>
    <version>${jsonschema2pojo.version}</version>
    <configuration>
        <annotationStyle>jackson2</annotationStyle>
        <includeJsr303Annotations>true</includeJsr303Annotations>
    </configuration>
</plugin>
```

Alternatively, the version may be declared directly:

```xml
<plugin>
    <groupId>org.jsonschema2pojo</groupId>
    <artifactId>jsonschema2pojo-maven-plugin</artifactId>
    <version>1.2.2</version>
    <configuration>
        <includeJsr303Annotations>true</includeJsr303Annotations>
    </configuration>
</plugin>
```

In the second example, `annotationStyle` is absent, so jsonschema2pojo implicitly uses its `jackson2` default.

The version may also be supplied through `pluginManagement` or through a property declared in a parent POM.

When the version is configured through a property, the property declaration should be updated while preserving the existing reference:

```xml
<properties>
    <jsonschema2pojo.version>1.3.0</jsonschema2pojo.version>
</properties>

<plugin>
    <groupId>org.jsonschema2pojo</groupId>
    <artifactId>jsonschema2pojo-maven-plugin</artifactId>
    <version>${jsonschema2pojo.version}</version>
    <configuration>
        <annotationStyle>jackson3</annotationStyle>
        <includeJsr303Annotations>true</includeJsr303Annotations>
        <useJakartaValidation>true</useJakartaValidation>
    </configuration>
</plugin>
```

When the version is declared directly, it should be updated directly:

```xml
<plugin>
    <groupId>org.jsonschema2pojo</groupId>
    <artifactId>jsonschema2pojo-maven-plugin</artifactId>
    <version>1.3.0</version>
    <configuration>
        <annotationStyle>jackson3</annotationStyle>
        <includeJsr303Annotations>true</includeJsr303Annotations>
        <useJakartaValidation>true</useJakartaValidation>
    </configuration>
</plugin>
```

Conceptually:

```diff
<properties>
-    <jsonschema2pojo.version>1.2.2</jsonschema2pojo.version>
+    <jsonschema2pojo.version>1.3.0</jsonschema2pojo.version>
</properties>
```

and:

```diff
<configuration>
-    <annotationStyle>jackson2</annotationStyle>
+    <annotationStyle>jackson3</annotationStyle>
     <includeJsr303Annotations>true</includeJsr303Annotations>
+    <useJakartaValidation>true</useJakartaValidation>
</configuration>
```

If `annotationStyle` is absent, the recipe should add it explicitly:

```diff
<configuration>
+    <annotationStyle>jackson3</annotationStyle>
</configuration>
```

This is necessary because leaving the option absent continues to select the plugin's `jackson2` default.

If Jakarta Validation is explicitly disabled while validation generation is active:

```diff
-    <useJakartaValidation>false</useJakartaValidation>
+    <useJakartaValidation>true</useJakartaValidation>
```

If `includeJsr303Annotations` is absent or false, the recipe should not add `useJakartaValidation`, because it would have no effect.

The version migration should follow the project's existing version-management style:

- if the plugin version references a Maven property, update that property rather than replacing the reference with a literal version;
- if the version is declared directly, update it directly;
- if the version is managed through `pluginManagement`, update the managed declaration;
- if the property or managed version is declared in a parent POM included in the rewrite run, update the declaration there;
- preserve any existing version equal to or newer than `1.3.0`;
- do not downgrade a newer developer-selected version.

If the plugin has no explicit version, the recipe should add or manage a compatible version unless a compatible resolved version can already be established through plugin management.

If the version cannot be resolved or its declaration is outside the rewrite run, the recipe should avoid silently adding unsupported Jackson 3 configuration and should report that the plugin version could not be verified or upgraded.

The plugin version and configuration can be updated manually after running the Spring Boot 4 recipe.

However, this is easy to overlook because the affected classes are generated sources. A project may appear to have migrated its handwritten source code successfully but continue regenerating models containing Jackson 2 databind or `javax.validation` annotations.

A generic Jackson 3 recipe could potentially update `annotationStyle`, and a generic Jakarta migration could potentially update `useJakartaValidation`.

However, for the narrower Spring Boot 4 use case:

- the application is known to target Jackson 3;
- the application is known to use Jakarta namespaces;
- the compatible jsonschema2pojo version floor is known;
- the required transformation is deterministic when Jackson and JSR-303 generation are enabled.

## Any additional context

jsonschema2pojo `1.3.0` added Jackson 3 annotation support. (double check this)

The relevant plugin options are:

- `annotationStyle=jackson3`, for Jackson 3 databind annotations;
- `includeJsr303Annotations=true`, to enable generated validation annotations;
- `useJakartaValidation=true`, to generate those validation annotations from `jakarta.validation` rather than `javax.validation`.

Jackson 3 continues to use the existing `com.fasterxml.jackson.annotation` package for core annotations, while Jackson databind annotations such as `JsonDeserialize` move to the Jackson 3 `tools.jackson.databind.annotation` package.

Support for the jsonschema2pojo Gradle plugin, non-Spring projects, non-Jackson annotation styles, and independent Jakarta migrations could be addressed separately.

Potentially, with guidance on where this transformation should be composed into the existing Spring Boot 4 recipes.

Kyran121:
- Fixes #1073.

Adds `MigrateJsonschema2PojoToSpringBoot4`, run as part of `UpgradeSpringBoot_4_0`, which updates `jsonschema2pojo-maven-plugin` configuration for Spring Boot 4 and Jackson 3. It sets `annotationStyle` to `jackson3` only when the existing value is `jackson`, `jackson2`, or absent, enables `useJakartaValidation` when `includeJsr303Annotations=true`. The recipe first upgrades the plugin to at least `1.3.0`, the release that introduced these options.


timtebeek (approved):
Perfect, thanks a lot! Let us know how that works out in practice for you there.

Issue #1081, opened 2026-07-18
[Tracking] Add Spring Data MongoDB 5.0 migration support


Kyran121:
## Related work

- #1080 introduces the datastore-neutral `spring-data-4.yml` and `UpgradeSpringData_4_0` foundation.
- #1063 tracks the broader Spring Data 2025.1 / JPA-oriented migration request.
- #1079 covers Spring Boot-owned MongoDB modular package moves and dedicated MongoDB test starters.

## What problem are you trying to solve?

Spring Boot 4 uses the Spring Data 2025.1 release train, which includes Spring Data MongoDB 5.0. `rewrite-spring` does not yet provide focused migration support for the Spring Data MongoDB 4.x to 5.x changes.

Spring Data support in `rewrite-spring` is incremental and does not need to cover every datastore before `UpgradeSpringData_4_0` becomes useful. This issue tracks the MongoDB workstream because MongoDB has confirmed migration requirements and concrete user demand.

The common Spring Data 4 YAML and public composite belong to #1080. This issue owns the MongoDB roadmap, sequencing, and integration status; implementation details and acceptance criteria are owned by the linked child issues below.

## Proposed structure

Once #1080 provides the common extension point, introduce a MongoDB-specific composite such as:

```text
org.openrewrite.java.spring.data.MigrateSpringDataMongoDb5
```

and compose it into:

```text
org.openrewrite.java.spring.data.UpgradeSpringData_4_0
```

Each confirmed MongoDB migration concern should remain independently implementable and reviewable. Completed child recipes should be appended to the MongoDB sub-composite so that `UpgradeSpringData_4_0` improves incrementally.

## Child implementation issues

The proposed work in this tracking issue will be delivered through these child issues:

- [x] #1082 — Add the MongoDB 5 migration composite and dependency alignment.
- [ ] #1083 — Detect missing explicit UUID and large-number representation configuration.
- [ ] #1084 — Preserve or report changed `DefaultMessageListenerContainer` startup behaviour.
- [ ] #1085 — Detect removed Spring Data MongoDB JMX support.
- [ ] #1092 — Replace removed MongoDB Java Driver uber JAR dependencies safely.
- [ ] #1093 — Migrate direct MongoDB legacy API usage to the current synchronous API where safe.

Each child issue owns its detailed scope, implementation approach, tests, and acceptance criteria. This parent issue should be updated with the corresponding PR links and completion status rather than duplicating those details here.

## Order of implementation

[...]

PR #1098, opened 2026-07-24
Detect missing MongoDB value representations

### LINKED ISSUE #1083: [Spring Data MongoDB 5.0] Detect missing explicit value representation configuration
## Parent tracking issue

- #1081

## Prerequisite

## Problem

Spring Data MongoDB 5 no longer provides default representations for persisted UUID, BigInteger, and BigDecimal values. Applications persisting these types must explicitly configure the corresponding representation rather than relying on earlier defaults.

A migration recipe must not silently choose a representation because selecting the wrong value can change how existing MongoDB data is read or written.

## Implemented solution

PR #1098 adds `FindMissingMongoValueRepresentation`, a project-scoped diagnostic recipe that:

- Detects explicitly MongoDB-mapped `UUID`, `BigDecimal`, and non-ID `BigInteger` fields requiring an explicit representation, including nested collection and map-value types.
- Excludes irrelevant fields such as non-Mongo domain fields, transient/static fields, explicitly targeted numeric fields, map keys, and `BigInteger` IDs.
- Reports all affected fields through the `MongoValueRepresentationFields` data table.
- Recognises valid Java configuration and Spring Boot configuration in `.properties` and YAML, including relaxed spellings and property placeholders.
- Ignores test-source configuration when evaluating main-source persistence and continues to diagnose invalid profile overrides even when another configuration file contains a valid default.
- Adds the relevant property with the bindable value `UNSPECIFIED` and an explanatory comment to the preferred main `application*.properties`/`.yml`/`.yaml` file when configuration is missing.
- Generates `src/main/resources/application.properties` when no suitable application configuration file exists.
- Adds persisted comments directly to blank, malformed, unsupported, `null`, or explicitly `UNSPECIFIED` Java, properties, and YAML configuration.
- Avoids duplicate suggestions, comments, and data-table rows across recipe cycles and separate invocations.
- Is composed into `UpgradeSpringDataMongoDb_5_0`.

`UNSPECIFIED` is intentionally not treated as a selected storage format. It keeps the generated configuration bindable while requiring the user to choose the representation matching the existing BSON data.

## Out of scope

- Choosing a concrete representation automatically from the Java field type.
- Rewriting existing MongoDB data.
- MongoDB dependency alignment; see #1082.
- Listener-container behaviour or removed JMX support.
- General Spring Boot property migration already handled by generated Boot property recipes.

## Acceptance criteria

- [ ] UUID persistence usage is detected when no explicit UUID representation is configured.
- [ ] `BigInteger` / `BigDecimal` persistence usage is detected when no explicit numeric representation is configured.
- [ ] Existing supported Java, properties, and YAML configuration suppresses the corresponding diagnostic.
- [ ] Invalid, blank, malformed, `null`, and `UNSPECIFIED` configuration is diagnosed at the offending source.
- [ ] No concrete representation is silently selected by the migration recipe.
- [ ] Missing configuration receives a bindable `UNSPECIFIED` property with an explanatory comment, including generation of a baseline configuration file when required.
- [ ] Tests cover configured, unconfigured, irrelevant, profile-specific, nested-value, generated-file, composite, and idempotent cases.
- [ ] The recipe is composed into the Spring Data MongoDB 5 migration composite.

## References

- Spring Boot 4.0 migration guide: https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-4.0-Migration-Guide

Kyran121:
## What's been done

- Split into two recipes per ADR 0002: `FindMissingMongoValueRepresentation` (pure search, `SearchResult` + data table only) and `AddMongoValueRepresentationProperty` (file mutation, moved out of `.search`).
- Detect explicitly MongoDB-mapped `UUID`, `BigDecimal`, and non-ID `BigInteger` fields needing a value representation for Spring Data MongoDB 5, including array/collection and map-value types.
- Recognise valid configuration in Java, `.properties`, and YAML — relaxed value spellings, property placeholders, and the legacy `spring.data.mongodb.uuid-representation` key.
- Preserve Spring Data MongoDB 4 behaviour when configuration is missing by explicitly adding `java-legacy` for UUID values and `string` for BigDecimal/BigInteger values.
- Add those defaults to the preferred base `application*.properties`/`.yml`/`.yaml` file; generate `application.properties` when no base file exists.
- Flag blank, malformed, unsupported, `null`, or explicitly `UNSPECIFIED` existing configuration with a persisted comment rather than overwriting it.
- Fix the profile-specific-file bottleneck: a value set only in `application-<profile>.*` or behind a Java `@Profile` no longer suppresses the base default needed for every other profile.
- Collapse a redundant 3-visitor split into one `JavaIsoVisitor` per format; scope every Java visitor with `Preconditions.check` per house convention.
- Fix an `NPE` on non-Java compilation units from an over-broad `instanceof` check.
- Fix duplicate data-table rows caused by `getVisitor()` running once per file instead of once per run; no test needs more than the framework's default cycle count anymore.
- Simplify preferred-configuration-source scoring now that profile-specific files are never merged into it.
- Compose both recipes into `UpgradeSpringDataMongoDb_5_0` and register recipe metadata.

## Validation

- Updated focused tests for Spring Data MongoDB 4 default preservation in generated configuration, existing properties/YAML, profile-specific configuration, cross-project isolation, composite execution, invalid explicit configuration, and idempotency.
- Explicit user-written `UNSPECIFIED`, `null`, malformed, or unsupported values remain diagnostic and are not silently replaced.
- Latest upstream CI attempts fail before compilation/tests while resolving the OpenRewrite build-plugin classpath: `artifacts.codegenomeproject.org` returns `401 Unauthorized` for Rewrite `8.91.6`. An earlier CI run from the same fork passed on September 6/7, so the failure is not established as being caused by this PR. The build resolves `org.openrewrite.build.recipe-library` via `latest.release`, so an external build-plugin change is a possible cause, but the root cause is not yet confirmed. No project tests ran in the failing attempts.

Closes #1083.
Part of #1081.


Kyran121:
@timtebeek - ready to review

Kyran121:
Closing to retrigger CI.

src/main/java/org/openrewrite/java/spring/data/search/FindMissingMongoValueRepresentation.java
```diff
@@ -0,0 +1,285 @@
...
+
+    /**
+     * A baseline file from {@link #generate} is created empty: the scan feeding that cycle's
+     * {@link Accumulator} predates the file, so populating it needs a further cycle to first scan
+     * the now-persisted file. Without opting in, a real run would stop after the generating cycle.
+     */
+    @Override
+    public boolean causesAnotherCycle() {
+        return true;
+    }
```
greg-at-moderne: Is the file always generated?
I am trying to understand whether this always-true is not too aggressive.
It's a costly setting after all.

greg-at-moderne (commented):
Thanks for the contribution!
I had an initial run at the code here and trying to understand what exactly it does.

Two unrelated questions:

1. At a high level, I am a bit confused about the recipe developed here. It's part of the `search` package, it's called `Find...`, yet it makes some changes.
Isn't it doing too much? Or maybe the name/placement of the recipe is not ideal?
2. How does this recipe interact with `UpgradeSpringBoot_4_0`? For instance - should this new recipe treat both `spring.data.mongodb.uuid-representation` and `spring.mongodb.representation.uuid` as configured?

src/main/java/org/openrewrite/java/spring/data/search/FindMissingMongoValueRepresentation.java
```diff
@@ -0,0 +1,285 @@
...
+    public TreeVisitor<?, ExecutionContext> getVisitor(Accumulator acc) {
+        return new TreeVisitor<Tree, ExecutionContext>() {
+            @Override
+            public @Nullable Tree visit(@Nullable Tree tree, ExecutionContext ctx) {
+                if (!(tree instanceof SourceFile)) {
+                    return tree;
+                }
+                SourceFile source = (SourceFile) tree;
+                JavaProject project = SpringConfigFileSupport.javaProject(source);
+                return project == null ? source : MongoValueRepresentationDiagnostics.apply(
```
MBoegers: The recipe is placed in the `search` package and is named `Find*` but produces code changes, this contradicts with our ADR about nameing and placing recipes https://github.com/openrewrite/rewrite/blob/main/doc/adr/0002-recipe-naming.md#recipe-package-naming.

Users expect a `Find*`-recipe and those under `search` subpackages to emit `SearchResult` markers and a data table. 
I ask you to follow the ADR and either rename/move or split the recipe.

src/main/java/org/openrewrite/java/spring/data/search/FindMissingMongoValueRepresentation.java
```diff
@@ -0,0 +1,285 @@
...
+            @Override
+            public @Nullable Tree visit(@Nullable Tree tree, ExecutionContext ctx) {
+                if (!(tree instanceof SourceFile)) {
+                    return tree;
+                }
+                SourceFile source = (SourceFile) tree;
+                JavaProject project = SpringConfigFileSupport.javaProject(source);
+                if (project != null) {
+                    MongoValueRepresentationScanner.scan(source, project, acc, ctx);
+                }
```
MBoegers: Is there a reason why you choose that the `TreeVisitor` forward into static utility that is applying a `JavaVisitor`?
Usually, we tend to implement the visitors directly and together.
in `MongoValueRepresentationScanner` you define 2 java visitors to visit disjunctive LST elements. Plus the TreeVisitor you now have 3 visitors where one would be sufficient.

Also using a `JavaVisitor` directly you can omit the ` if (!(tree instanceof SourceFile))` check.

In this case roughly:
```java
public TreeVisitor<?, ExecutionContext> getScanner(Accumulator acc) {
  return new JavaIsoVisitor<ExecutionContext>() {
    public J.CompilationUnit visitCompilationUnit(J.CompilationUnit cu, ExecutionContext ctx) {
       if (source.getMarkers().findFirst(ProjectDiagnostic.class).isPresent()) {
            acc.finalizedProjects.add(project);
        }
        if (source.getMarkers().findFirst(RowsRecorded.class).isPresent()) {
            acc.rowsInsertedProjects.add(project);
        }
        return SpringConfigFileSupport.isMainSource(source) ? super.visitCompilationUniy(cu, ctx) : cu:
    }

    public J.MethodInvocation visitMethodInvocation(J.MethodInvocation method, ExecutionContext p) {
      J.MethodInvocation m = super.visitMethodInvocation(method, p);
      if (UUID_REPRESENTATION.matches(m)) {
         recordJavaConfigurationAttempt(ValueKind.UUID, m, source, project, acc);
      }
      if (BIG_NUMBER_REPRESENTATION.matches(m)) {
          recordJavaConfigurationAttempt(ValueKind.BIG_NUMBER, m, source, project, acc);
      }
      return m;
    }

    @Override
    public J.VariableDeclarations visitVariableDeclarations(J.VariableDeclarations declarations, ExecutionContext p) {
      //...
    ]
  }
}
```

src/test/java/org/openrewrite/java/spring/data/search/FindMissingMongoValueRepresentationTest.java
```diff
@@ -0,0 +1,400 @@
...
+                    private UUID externalId;
+                    private BigDecimal balance;
+                    private BigInteger sequence;
+                }
+                """,
+              spec -> spec.path("src/main/java/com/example/Account.java")
+            ),
+            properties(
+              null,
+              spec -> spec
```
MBoegers: Why don't give the real output? this would be easier to understand.

src/test/java/org/openrewrite/java/spring/data/search/MongoValueRepresentationTestSupport.java
```diff
@@ -0,0 +1,162 @@
...
+            public Yaml.Mapping.Entry visitMappingEntry(Yaml.Mapping.Entry entry, AtomicBoolean marked) {
+                Yaml.Mapping.Entry e = super.visitMappingEntry(entry, marked);
+                if (e.getKey() instanceof Yaml.Scalar && key.equals(e.getKey().getValue()) &&
+                        Comments.of(getCursor()).getComments().stream().anyMatch(comment -> comment.contains(key))) {
+                    marked.set(true);
+                }
+                return e;
+            }
+        }.visit(file, found);
+        assertThat(found.get()).as("Expected YAML entry '%s' to carry a diagnostic comment", key).isTrue();
```
MBoegers: You know you can use this setup to have a decalrative easy to follow test with OpenRewrite tests?
```java
rewriteRun(
  yaml("""
    some:
      key: value
      other: also
    """,
    """
    some:
      # ~~ some.key found  ~~>
      key: value
      other: also
    """
)
```

src/test/java/org/openrewrite/java/spring/data/search/FindMissingMongoValueRepresentationTest.java
```diff
@@ -0,0 +1,400 @@
...
+import static org.openrewrite.properties.Assertions.properties;
+import static org.openrewrite.yaml.Assertions.yaml;
+
+class FindMissingMongoValueRepresentationTest extends MongoValueRepresentationTestSupport {
+
+    @DocumentExample
+    @Test
+    void reportsOncePerProjectAndListsAllAffectedFields() {
+        rewriteRun(
+          spec -> spec.cycles(4).expectedCyclesThatMakeChanges(3)
```
MBoegers: `spec.cycles(4).expectedCyclesThatMakeChanges(3)` is usually a sign for a recipe not behaving as expected. Why do we need more than one cyle to populate the data tables and add the SearchMarkers?

MBoegers (changes requested):
Thanks again for the contribution to OpenRewrite.
I added a few central comments to the submission that I ask you to address.
Also I ask you to revisit and respect the following guides, as they explain in more depth what I called out here and why.

- Testing an OpenRewrite recipe: https://docs.openrewrite.org/authoring-recipes/recipe-testing
- OpenRewrite recipes best practices https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices
- `SearcheResult` marker documentation: https://docs.openrewrite.org/concepts-and-explanations/markers#searchresult
- `ScanningRecipe` documentation: https://docs.openrewrite.org/concepts-and-explanations/recipes#scanning-recipes

CI problem should be gone after a merge of main.

src/main/java/org/openrewrite/java/spring/data/search/FindMissingMongoValueRepresentation.java
Kyran121: Good catch — that always-true was left over and unjustified. It's gone: there's no forced-regeneration flag anywhere in the recipe now. The recipe settles within the framework's default 2 cycles (verified by removing every `cycles(...)` override from the tests), so it isn't doing unnecessary repeat work per project.

src/main/java/org/openrewrite/java/spring/data/search/FindMissingMongoValueRepresentation.java
Kyran121: Done — split per the ADR. `FindMissingMongoValueRepresentation` stays in `.search` and is now a pure search recipe (only `SearchResult` markers + the data table, no file mutation). The property-suggestion/comment-insertion logic moved to a new `AddMongoValueRepresentationProperty`, placed outside the `search` package since it edits files. `UpgradeSpringDataMongoDb_5_0` composes both.

src/main/java/org/openrewrite/java/spring/data/search/FindMissingMongoValueRepresentation.java
Kyran121: Fixed as you suggested. `FindMissingMongoValueRepresentation.getVisitor()` is now a single `JavaIsoVisitor` implemented directly (visitCompilationUnit + visitVariableDeclarations together, no forwarding, no extra wrapper) — matches the shape you sketched. The shared scanning logic used by *both* recipes now lives in `MongoValueRepresentationScanner`, a single class with one visitor per source kind (Java/Properties/YAML); the outer `TreeVisitor`+`instanceof` there is dispatching between three genuinely different LST types (no common base below `TreeVisitor`), which is the same pattern already used in this codebase by `IntegrationSchedulerPoolRecipe` for the same reason.

src/test/java/org/openrewrite/java/spring/data/search/FindMissingMongoValueRepresentationTest.java
Kyran121: Switched to real output throughout — tests now use two-arg `java(before, after)`/`yaml(before, after)` showing the actual `SearchResult` marker text (e.g. `/*~~(missing MongoDB value representation configuration)~~>*/`) instead of custom assertions.

src/test/java/org/openrewrite/java/spring/data/search/MongoValueRepresentationTestSupport.java
Kyran121: Adopted this pattern across the suite — both properties and YAML tests now show the real before/after marker output declaratively, as you outlined.

src/test/java/org/openrewrite/java/spring/data/search/FindMissingMongoValueRepresentationTest.java
Kyran121: You were right — that was masking a bug. `getVisitor(acc)` was being invoked once per source file (not once per run), which was inserting duplicate data-table rows and needing extra cycles to settle. Fixed by guarding row-insertion in the accumulator (`markRowsInserted`) so it only happens once per project. No test in the suite overrides `cycles(...)` anymore — everything now passes on the framework's default (2 cycles, 1 that makes changes).

src/main/java/org/openrewrite/java/spring/data/AddMongoValueRepresentationProperty.java
```diff
@@ -0,0 +1,351 @@
...
+public class AddMongoValueRepresentationProperty extends ScanningRecipe<MongoValueRepresentationAccumulator> {
+
+    @Getter
+    final String displayName = "Add a placeholder MongoDB value representation property";
+
+    @Getter
+    final String description = "Add a placeholder value-representation property for MongoDB-mapped UUID, BigInteger, " +
+            "and BigDecimal fields that require one when migrating to Spring Data MongoDB 5, and flag any existing " +
+            "configuration whose value isn't a concrete representation, without choosing a representation on the " +
+            "user's behalf.";
```
MBoegers: > without choosing a representation on the user's behalf.
This is not true. This PR falls back to default, which will break for applications that already have a database underneth.

src/test/java/org/openrewrite/java/spring/data/AddMongoValueRepresentationPropertyTest.java
```diff
@@ -0,0 +1,363 @@
...
+              accountWithUuidAndBigDecimal(),
+              spec -> spec.path("src/main/java/com/example/Account.java")
+            ),
+            properties(
+              null,
+              """
+                # `spring.data.mongodb.representation.big-decimal` needs a concrete big-number representation matching the existing BSON data.
+                spring.data.mongodb.representation.big-decimal=UNSPECIFIED
+                # `spring.mongodb.representation.uuid` needs a concrete UUID representation matching the existing BSON data.
+                spring.mongodb.representation.uuid=UNSPECIFIED
```
MBoegers: we should use the value that Spring Data 4 uses. `UNSPECIFIED` will break at next application start due to incompatibilities

MBoegers (changes requested):
Sorry to push back here again. 
From the technical perspective we break applications at next startup if they relay on the frameworks defaults. 
I ask you to comply to [do-no-harm](https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices#do-no-harm) rule

src/main/java/org/openrewrite/java/spring/data/AddMongoValueRepresentationProperty.java
Kyran121: Agreed — the automatic path no longer emits `UNSPECIFIED`. For previously unconfigured applications it now makes the Spring Data MongoDB 4 effective defaults explicit: UUID -> `java-legacy` (`JAVA_LEGACY`) and BigDecimal/BigInteger -> `string` (`STRING`). That preserves the pre-upgrade storage behaviour instead of falling through to the Spring Data MongoDB 5 behaviour. Existing explicit invalid/`UNSPECIFIED` configuration is still only diagnosed and is not overwritten. The implementation, tests, PR description, and #1083 contract have been updated accordingly.

src/test/java/org/openrewrite/java/spring/data/AddMongoValueRepresentationPropertyTest.java
Kyran121: Fixed. Missing UUID configuration now gets `java-legacy`, and missing BigDecimal/BigInteger configuration gets `string`, matching the Spring Data MongoDB 4 defaults. The properties, YAML, generated-baseline, profile/cross-project boundary, and Spring Data MongoDB 5 composite tests have been updated to assert those values. Explicit user-written `UNSPECIFIED` remains diagnostic rather than being silently replaced.

Kyran121:
CI note: the current head fails before compilation/tests while resolving the build plugin classpath. The failing runs resolve `org.openrewrite.build.recipe-library` `latest.release` to `2.24.0` and receive Code Genome `401 Unauthorized` responses for Rewrite `8.91.6` artifacts. The last passing run (#5833) was before `2.24.0` was published, when `latest.release` was `2.23.4`. No project tests or recipe metadata checks are reached.

PR #1104, opened 2026-07-27
Include Jackson PropertyNamingStrategy migration in Spring Boot 2.5 upgrade

steve-aom-elliott:
## Summary

Spring Boot 2.5 ships with Jackson 2.12, which deprecated `com.fasterxml.jackson.databind.PropertyNamingStrategy`'s inner classes (`SnakeCaseStrategy`, `KebabCaseStrategy`, `UpperCamelCaseStrategy`, `LowerCamelCaseStrategy`, `LowerCaseStrategy`, `LowerDotCaseStrategy`) and companion constants (`SNAKE_CASE`, `KEBAB_CASE`, ...) in favor of the new `PropertyNamingStrategies` class. Later 2.x releases have removed the deprecated inner classes outright, so code using the old form compiles against 2.12 but fails against current 2.x.

`org.openrewrite.java.jackson.ReplacePropertyNamingStrategyConstants` already handles this rewrite (via `ChangeType` for the inner classes and `ReplaceConstantWithAnotherConstant` for the constants), but it wasn't reachable from any Spring Boot composite. This PR wires it into `UpgradeSpringBoot_2_5` so applications getting bumped to SB 2.5 receive the Jackson deprecation fix as part of the same upgrade path.

## Changes

- `spring-boot-25.yml` — add the recipe to `UpgradeSpringBoot_2_5`'s `recipeList` alongside the existing `ChangeType`-based deprecated-API replacements.
- `MigrateJacksonPropertyNamingStrategyTest` — new end-to-end test that activates `UpgradeSpringBoot_2_5` and asserts the rename fires for both the `@JsonNaming(PropertyNamingStrategy.SnakeCaseStrategy.class)` annotation form and the `PropertyNamingStrategy.SNAKE_CASE` constant form.
- `build.gradle.kts` — pin `jackson-databind` to `2.19.+` on the test parser classpath (the newest 2.x that still contains the deprecated inner classes, so the parser can resolve the pre-migration form the test asserts on). Matches what `rewrite-jackson`'s own tests use.
- `classpath.tsv.gz` — regenerated by `./gradlew createTestTypeTable`.

## Test plan

- [x] `./gradlew test --tests "org.openrewrite.java.spring.boot2.MigrateJacksonPropertyNamingStrategyTest"` — both tests pass end-to-end through `UpgradeSpringBoot_2_5`.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -105,6 +105,7 @@ recipeDependencies {
     parserClasspath("jakarta.ws.rs:jakarta.ws.rs-api:4.0.0")
 
     testParserClasspath("com.fasterxml.jackson.core:jackson-core:2.20.+")
+    testParserClasspath("com.fasterxml.jackson.core:jackson-databind:2.19.+")
     testParserClasspath("com.nimbusds:nimbus-jose-jwt:9.13")
     testParserClasspath("io.projectreactor:reactor-core:3.6.3")
     testParserClasspath("io.springfox:springfox-core:3.+")
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -105,6 +105,7 @@ recipeDependencies {
     parserClasspath("jakarta.ws.rs:jakarta.ws.rs-api:4.0.0")
 
     testParserClasspath("com.fasterxml.jackson.core:jackson-core:2.20.+")
+    // 2.19 is the last release with the deprecated `PropertyNamingStrategy` inner classes we migrate away from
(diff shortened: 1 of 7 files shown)


steve-aom-elliott:
`jackson-databind` 2.20 was the threshold where the old class names were actually removed, while 2.12 was where the new ones were introduced, so the `testParserClasspath` could have been for a slightly lower version if we wanted, but it just needed to be below 2.20

PR #1109, opened 2026-07-29
Migrate to the new Spring Batch 6.0 chunk-oriented step model

timtebeek:
### What's changed?

`SpringBatch5To6Migration` left `StepBuilder.chunk(int, PlatformTransactionManager)` calls untouched, so every chunk-oriented step still tripped SonarQube's `java:S5738` ("Deprecated code should not be used") after running the migration.

New recipe `MigrateToChunkOrientedStepBuilder`, wired into `SpringBatch5To6Migration`:

```java
// before
return new StepBuilder("myStep", jobRepository)
        .<String, String>chunk(10, transactionManager)
        .reader(reader)
        .writer(writer)
        .build();

// after
return new StepBuilder("myStep", jobRepository)
        .<String, String>chunk(10)
        .transactionManager(transactionManager)
        .reader(reader)
        .writer(writer)
        .build();
```

That moves the step from the deprecated `SimpleStepBuilder`/`FaultTolerantStepBuilder` onto `ChunkOrientedStepBuilder`, which is a change of `Step` implementation, so the recipe only fires when the whole chain can be expressed on the new builder:

- the chain has to end in `build()`, so the widened return type stays contained (and a `SimpleStepBuilder`-typed intermediate variable is left alone);
- the result must not be assigned to or returned as `TaskletStep`, since the new builder returns `ChunkOrientedStep`;
- every method in the chain has to have an equivalent on `ChunkOrientedStepBuilder`;
- `listener(..)` has to take a `StepListener` — the new model silently drops `listener(Object)` annotation-based listeners and Spring Retry `RetryListener`s;
- `taskExecutor(..)` has to take an `AsyncTaskExecutor`, which is what the new model narrowed the parameter to.

Chains calling `backOffPolicy`, `retryPolicy`, `retryContextCache`, `keyGenerator`, `noRetry`, `noRollback`, `noSkip`, `processorNonTransactional`, `readerIsTransactionalQueue`, `chunkOperations`, `stepOperations`, `exceptionHandler` or `chunk(CompletionPolicy, PlatformTransactionManager)` have no counterpart in the new model, so they are left untouched and remain a manual migration step. The recipe description enumerates them so it shows up in the generated docs.

`AddTransactionManagerToTaskletAndChunk` also no longer adds a transaction manager argument when the chain already chains a `transactionManager(..)` call. Spring Batch 6.0 reintroduced the single argument `tasklet` and `chunk` methods, so without this the two recipes fought each other inside `SpringBatch5To6Migration` and produced a duplicated `transactionManager(..)` call on the second cycle. It also means running the 4→5 migration against an already-migrated Spring Batch 6 codebase is now a no-op there.

### Anything in particular you'd like reviewers to focus on?

The list of methods considered safe on `ChunkOrientedStepBuilder`, in `SUPPORTED_BUILDER_METHODS`. It is a whitelist, so anything unrecognised makes the recipe bail, but a method wrongly included would produce code that no longer compiles or silently changes behavior.

### Have you considered any alternatives or workarounds?

Rewriting the full fault-tolerant retry configuration as well. The new model dropped Spring Retry (`org.springframework.retry.RetryPolicy` → `org.springframework.core.retry.RetryPolicy`, no `BackOffPolicy` equivalent), and the Spring Batch team keeps the legacy implementation available for the whole v6 generation, so there is no mechanical mapping for those. Documenting them as manual steps in the recipe description seemed better than a partial rewrite that does not compile.

### Any additional context

- Reported via https://github.com/moderneinc/customer-requests/issues/2915. Behavior verified against the [Spring Batch 6.0 migration guide](https://github.com/spring-projects/spring-batch/wiki/Spring-Batch-6.0-Migration-Guide#new-chunk-oriented-model-implementation) and the `ChunkOrientedStepBuilder` source.

---

**Checklist**

- [x] I've used the [IntelliJ IDEA auto-formatter](https://docs.openrewrite.org/reference/code-style-guide) on affected files


steve-aom-elliott:
Working on getting some adjustments together for this one

steve-aom-elliott (approved):
I believe I've push all the changes that seemed meaningful now.

timtebeek:
Thanks for the thorough fixes!

Issue #1084, opened 2026-07-18
[Spring Data MongoDB 5.0] Preserve DefaultMessageListenerContainer startup behaviour


Kyran121:
## Parent tracking issue

- #1081

## Prerequisite

- #1082 provides the Spring Data MongoDB 5 migration composite.

## What problem are you trying to solve?

In Spring Data MongoDB 4.x, `DefaultMessageListenerContainer#isAutoStartup()` always returns `false`. Spring Data MongoDB 5 changes the default to `true` and adds `setAutoStartup(boolean)` for explicit configuration.

As a result, a `DefaultMessageListenerContainer` that is registered as a Spring-managed bean and relies on the previous default will be started by Spring's `SmartLifecycle` processing during application-context startup after migration. If subscriptions have been registered, their listener tasks will start at that point rather than waiting for an explicit call to `start()`.

This can change application sequencing without any corresponding application source change. For example, an application may intentionally delay listener startup until another component is ready, a feature flag is enabled, or leadership has been acquired.

The migration objective is to preserve the Spring Data MongoDB 4.x manual-start behaviour for affected Spring-managed containers by configuring `autoStartup` to `false` explicitly where this can be done safely. It is not to guess whether an application would prefer the new automatic-start behaviour.

Instances that are not managed by the Spring application context are not started automatically merely because this default changed and should not be modified solely on that basis.

## Proposed solution

Add a focused recipe or diagnostic that:

- Identifies Spring-managed `DefaultMessageListenerContainer` instances, including common Java `@Bean` methods that directly construct or return the container.
- Detects containers that do not already configure startup behaviour explicitly.
- Preserves the previous default where unambiguous by adding an equivalent of `container.setAutoStartup(false)` before the container is exposed as a Spring bean.
- Leaves containers with explicit startup configuration unchanged.
- Does not modify non-Spring-managed instances solely because the `SmartLifecycle` default changed.
- Produces a clear diagnostic where bean ownership or lifecycle configuration is too indirect or dynamic to preserve mechanically with confidence.
- Composes the completed recipe into the MongoDB 5 migration composite introduced by #1082.

The implementation should prioritise common Java `@Bean` configuration patterns. XML support should only be included where the relevant rewrite infrastructure and test coverage make the transformation precise.

## Out of scope

- Changing unrelated Spring lifecycle configuration.
- Opting applications into the new automatic-start behaviour when they previously relied on manual startup.
[...]

PR #1118, opened 2026-08-06
Preserve MongoDB listener container startup behavior

### LINKED ISSUE #1085: [Spring Data MongoDB 5.0] Migrate removed MongoDB JMX support without breaking compilation
## Parent tracking issue

- #1081

## Prerequisite

Spring Data MongoDB 5 removes its dedicated JMX monitoring support. Applications upgrading from Spring Data MongoDB 4.x may still contain Java or XML configuration that depends on APIs which no longer exist in 5.x.

Simply diagnosing those usages and then continuing with the Spring Data MongoDB 5 dependency upgrade can leave the migrated project with unresolved types and compilation failures. The migration should not knowingly introduce that broken state.

There is no universal one-to-one replacement for the removed monitoring functionality. The appropriate replacement depends on the application and deployment environment and may involve Spring Boot Actuator, Micrometer, MongoDB driver observability, or organisation-specific monitoring infrastructure.

The migration therefore needs to distinguish source that can be removed or rewritten safely from application-specific usage that requires manual migration, while preserving compilability.

## Proposed solution

Add focused Spring Data MongoDB JMX migration support that:

- Identifies Java references to Spring Data MongoDB JMX APIs that were actually removed in 5.0, using attributed types rather than broad name matching.
- Detects removed Spring Data MongoDB XML JMX configuration where it can be matched precisely and reliably.
- Removes obsolete configuration whose purpose is exclusively the removed Spring Data MongoDB JMX facility when that removal is locally safe.
- Cleans up imports and configuration made unused by a successful removal.
- Detects when removed JMX types escape into otherwise surviving application code, such as custom subclasses, fields, parameters, return types, cross-bean dependencies, or application-specific logic.
- Reports unsupported migration sites clearly rather than deleting or rewriting arbitrary application behaviour.
- Exposes a reusable module/project migration-blocker signal so the Spring Data MongoDB 5 dependency transition is not applied when unsupported removed-JMX usage remains.
- Leaves unrelated Spring Framework, JDK, third-party, and application-defined JMX usage unchanged.
- Does not automatically select or configure a replacement monitoring strategy.

### Compilability invariant

`UpgradeSpringDataMongoDb_5_0` must not knowingly introduce unresolved references to Spring Data MongoDB APIs removed in 5.0.

For a module whose removed JMX usage is fully and safely migratable:

1. remove the obsolete JMX configuration;
2. remove now-unused imports/configuration;
3. continue with the Spring Data MongoDB 5 migration.

For a module containing unsupported removed-JMX usage:

1. preserve the source requiring manual migration;
2. report the exact blocking usages;
3. do not apply the Spring Data MongoDB 5 dependency transition for that affected module through this composite.

This issue owns the safety contract of the MongoDB-specific composite. Higher-level dependency management, including Spring Boot 4 parent/BOM upgrades that may independently bring Spring Data MongoDB 5 into a project, must honour the same blocker through separate integration work.

### Java detection boundary

Detection should be based on resolved Spring Data MongoDB types and on an explicit 4.x-present / 5.x-absent API set.

#### Types and configuration to detect

Spring Data MongoDB's 4.x JMX documentation exposes the monitoring MBeans backed by `org.springframework.data.mongodb.monitor`, and the 4.5 API marks that package for removal. Spring Data MongoDB 5 removes the package as part of discontinuing dedicated JMX support.

The removed Java type set to detect is therefore:

- `org.springframework.data.mongodb.monitor.AbstractMonitor`
- `org.springframework.data.mongodb.monitor.AssertMetrics`
- `org.springframework.data.mongodb.monitor.BackgroundFlushingMetrics`
- `org.springframework.data.mongodb.monitor.BtreeIndexCounters`
- `org.springframework.data.mongodb.monitor.ConnectionMetrics`
- `org.springframework.data.mongodb.monitor.GlobalLockMetrics`
- `org.springframework.data.mongodb.monitor.MemoryMetrics`
- `org.springframework.data.mongodb.monitor.OperationCounters`
- `org.springframework.data.mongodb.monitor.ServerInfo`

The XML/configuration surface should also account for:

- the Mongo namespace JMX element: `<mongo:jmx/>`;
- `org.springframework.data.mongodb.config.MongoJmxParser`, which backs that namespace configuration in 4.5 and is absent from the 5.x configuration API.

The historical JMX documentation also lists `MongoAdmin`, but `org.springframework.data.mongodb.core.MongoAdmin` and `MongoAdminOperations` remain present in Spring Data MongoDB 5.0 and must **not** be classified as removed 5.0 types merely because they are deprecated for future removal. They may warrant separate migration work when they are actually removed in a later target version.

This list should be treated as the initial authoritative detection boundary and kept aligned with the Spring Data MongoDB 4.x-to-5.x migration/API documentation rather than expanded through generic JMX naming heuristics.

Relevant semantic usage includes, where applicable:

- field and local-variable types;
- constructor calls;
- method parameters and return types;
- inheritance;
- generic type arguments;
- fully-qualified references;
- casts, class literals, and other attributed references.

Imports are secondary evidence. The migration should operate on the affected semantic usage and allow unused-import cleanup to remove obsolete imports after successful transformations.

APIs that remain present in Spring Data MongoDB 5, even if deprecated or historically associated with JMX, must not be reported as removed solely on that basis.

### Safe-removal boundary

A JMX usage is mechanically removable only where the recipe can establish that the affected source exists solely to configure or expose the removed Spring Data MongoDB JMX facility and that removing it does not leave references from surviving application code.

Examples that may be safely removable include dedicated monitoring bean declarations or precise XML namespace configuration whose only purpose is the removed JMX support.

Examples that should block automatic migration unless a safe transformation can be proven include:

- application classes extending a removed monitor type;
- removed JMX types stored in surviving fields;
- removed JMX types passed into or returned from surviving application methods;
- application-specific logic built around removed monitoring objects;
- cross-bean or cross-method dependencies where removing one JMX declaration would leave surviving code uncompilable;
- any case whose ownership or behavioural impact cannot be established locally with confidence.

Each usage should be evaluated independently rather than treating an entire class as unsupported because one member is ambiguous.

### XML detection boundary

Spring Data MongoDB XML JMX configuration should only be removed where the obsolete namespace element or configuration can be identified precisely.

General Spring Framework JMX configuration, application-managed MBean exporters, JDK JMX infrastructure, and unrelated XML configuration must remain unchanged.

If the available Rewrite XML infrastructure cannot distinguish a removed Spring Data MongoDB configuration reliably, that case should remain unsupported and act as a migration blocker rather than falling back to broad textual matching.

### Diagnostic behaviour

For unsupported migration sites, explain that:

- Spring Data MongoDB's dedicated JMX monitoring support was removed in 5.0;
- no universally equivalent source replacement exists;
- the affected application logic must be migrated manually;
- Spring Boot Actuator or another supported observability mechanism may be appropriate depending on the application;
- the MongoDB 5 transition was withheld for the affected module by this composite to avoid leaving known uncompilable source.

## Proposed implementation shape

The following is architectural guidance rather than a prescribed class/visitor design. The implementation should preserve these behaviours while remaining free to use the most appropriate OpenRewrite primitives.

```text
Analyse removed Mongo JMX usage
        |
        +-- safely migratable
        |       `-- remove obsolete JMX-only configuration
        |
        `-- unsupported
                `-- record migration blocker + diagnostic

                     |
                     v
       Spring Data MongoDB 5 transition
                     |
              only when safe
```

A likely shape is to:

- perform project/module-aware analysis before the target dependency transition while Spring Data MongoDB 4.x types are still attributable;
- classify individual usages as safely removable or unsupported rather than treating all JMX usage uniformly;
- expose reusable migration-readiness/blocker state at the module/project level;
- emit diagnostics independently of the guarded migration so blocked modules still explain exactly why they were withheld;
- apply safe removals before the dependency transition;
- guard the MongoDB 5 transition when unsupported removed-JMX usage remains;
- make the resulting readiness signal reusable by higher-level callers such as the Spring Boot 4 migration rather than duplicating the JMX analysis there.

The exact recipe classes, accumulators, visitors, precondition composition, and internal data structures are intentionally left to the implementor.

## Out of scope

- Automatically adding Spring Boot Actuator dependencies.
- Automatically enabling Actuator endpoints or JMX exposure.
- Selecting an application monitoring strategy.
- Deleting application-specific classes or behaviour merely to make the target dependency compile.
- Replacing organisation-specific JMX exporters or monitoring infrastructure.
- Rewriting unrelated Spring Framework, JDK, third-party, or application-owned JMX usage.
- Reporting APIs that remain available in Spring Data MongoDB 5 solely because they are deprecated.
- General Spring Data MongoDB package/import migrations unrelated to removed JMX support.
- MongoDB value-representation configuration or listener-container behaviour.
- Spring Boot Actuator MongoDB package migrations; see #1079.
- Preventing a separate Spring Boot parent/BOM/platform migration from independently selecting Spring Data MongoDB 5; that orchestration is tracked separately.

## Acceptance criteria

- [ ] The removed Spring Data MongoDB JMX Java API set is established from the 4.x to 5.x API change rather than broad JMX naming heuristics.
- [ ] Java references to APIs removed in 5.0 are identified using attributed types.
- [ ] The initial detection set covers the removed `org.springframework.data.mongodb.monitor` types listed above and the `<mongo:jmx/>` configuration surface.
- [ ] `MongoAdmin` / `MongoAdminOperations` are not incorrectly treated as removed 5.0 APIs while they remain present in the target API.
- [ ] Safe, JMX-only Java configuration can be removed without leaving unresolved references in surviving source.
- [ ] Relevant removed XML JMX configuration is removed only where support is precise and testable.
- [ ] Imports/configuration made unused by a successful migration are cleaned up.
- [ ] Removed JMX types escaping into surviving application code are identified as blockers rather than being deleted or speculatively rewritten.
- [ ] Cross-bean/cross-method dependencies are not partially removed in a way that leaves surviving code uncompilable.
- [ ] APIs that remain available in Spring Data MongoDB 5 are not incorrectly reported as removed.
- [ ] Unrelated Spring Framework, JDK, third-party, and application-defined JMX usage remains unchanged.
- [ ] `UpgradeSpringDataMongoDb_5_0` does not apply its Spring Data MongoDB 5 dependency transition to a module with unresolved JMX blockers.
- [ ] Modules whose JMX usage is fully migrated continue through the MongoDB 5 migration normally.
- [ ] Blocked modules receive clear diagnostics at the exact unsupported migration sites.
- [ ] Tests cover pure JMX configuration, mixed configuration, cross-bean dependencies, custom subclasses, fields/parameters/return types, fully-qualified references, unrelated JMX, multi-module behaviour, XML where supported, composite behaviour, and idempotency.
- [ ] The completed JMX migration support is composed into the Spring Data MongoDB 5 migration flow without knowingly leaving removed API references behind.

## References

- Spring Data MongoDB JMX reference documentation.
- Spring Data MongoDB 4.x API documentation.
- Spring Data MongoDB 5.x API documentation.

fregulha:
## What's changed?

* Added `PreserveDefaultMessageListenerContainerStartup`, which preserves manual startup behavior for Spring-managed `DefaultMessageListenerContainer` beans during migration to Spring Data MongoDB 5.0.
* The recipe adds `setAutoStartup(false)` when the transformation can be performed safely.
* Existing explicit `setAutoStartup(...)` configurations remain unchanged.
* Unsupported configurations are marked for manual review.
* Integrated the recipe into `UpgradeSpringDataMongoDb_5_0`.
* Added unit tests and a composite-recipe integration test.
* Updated the recipe catalog and parser classpath resources.

## What's your motivation?

Spring Data MongoDB 5.0 automatically starts `DefaultMessageListenerContainer` beans. Applications that previously started these containers manually could experience an unintended behavior change after upgrading.

This recipe preserves the previous behavior by explicitly disabling automatic startup where appropriate.

- Fixes #1085

The detection of supported `@Bean` factory methods, preservation of existing explicit startup configuration, and manual-review handling for configurations that cannot be transformed safely.

No specific reviewer.

Users could identify and update each listener container manually, but including this transformation in the MongoDB 5.0 migration recipe makes the upgrade safer and repeatable.

## Any additional context

The dedicated recipe and composite MongoDB 5.0 migration test suites pass successfully. Recipe marketplace CSV and license validations also pass.

The complete test suite was attempted on Windows. Six tests outside the affected packages failed: three import expectation tests in `GetErrorAttributesTest` and three Unix path-separator assertions in `RenameLogbackToLogbackSpringTest`.

### Checklist

- [x] I've read and applied the recipe conventions and best practices


Kyran121:
Hi @fregulha - I believe this is tackling #1084, not #1085. I also opened #1119 around the same time as you to tackle the `DefaultMessageListenerContainer` `autoStartup` migration.

My implementation covers a broader set of @Bean return patterns, including direct construction, locally-declared containers, multiple and nested return paths, locally assigned fields, same-named variables in separate scopes, reassignment, and simple one-hop same-class helper methods. It evaluates each return path independently and tracks variables by `JavaType.Variable` identity, so supported paths can still be migrated even where another return path in the same bean needs manual review.

It also preserves explicit `setAutoStartup(...)` configuration on a per-path basis and, for supported helper methods, configures the value returned from the bean rather than modifying the helper itself. This avoids changing the behaviour of other bean or non-bean callers that may share that helper.

For cases where the container provenance cannot be established safely — such as deeper helper chains, cross-class calls, or fields without a locally-visible assignment — it leaves the return unchanged and adds a targeted diagnostic for manual review.

It looks like our PRs overlap substantially, so it would probably make sense to align on which implementation to take forward.

PR #1119, opened 2026-08-06
Preserve DefaultMessageListenerContainer startup behaviour

### LINKED ISSUE #1084: [Spring Data MongoDB 5.0] Preserve DefaultMessageListenerContainer startup behaviour

Kyran121:
## What changed

- Add `org.openrewrite.java.spring.data.PreserveMongoListenerContainerAutoStartup`.
- Compose the recipe into `UpgradeSpringDataMongoDb_5_0`.
- Evaluate each `DefaultMessageListenerContainer` return path independently.
- Preserve manual startup for locally-verifiable paths: direct construction, locally-declared containers, same-block field assignment, and simple one-hop helper methods in the same class.
- Keep explicit `setAutoStartup(...)` configuration unchanged.
- Track variables by `JavaType.Variable` identity so same-named variables in separate scopes are handled independently.
- For a supported same-class helper, configure the bean's returned helper result rather than modifying the helper itself, avoiding behavior changes for other callers.
- Add a targeted diagnostic comment to return paths whose provenance cannot be verified safely, including deeper helper chains, cross-class calls, and fields without a locally-visible assignment.
- Leave non-Spring-managed instances unchanged.
- Avoid synthesized variable-name collisions.

## Detection boundaries

The recipe deliberately stops short of general interprocedural analysis. A same-class helper is followed one hop only when its returned container is directly constructed or held in a directly-constructed local variable. Explicit startup configuration in that helper is respected. Anything beyond that boundary is left unchanged and marked for manual review.

## Tests

Coverage includes:
- direct constructor returns;
- locally-declared containers;
- multiple direct return paths;
- same-named variables in separate scopes;
- explicit startup configuration;
- fields assigned directly before return;
- fields with unknown provenance;
- one-hop same-class helpers returning a direct construction;
- one-hop same-class helpers returning a local variable;
- helpers with explicit startup configuration;
- two bean methods sharing the same helper, with both bean return sites configured;
- a bean and non-bean method sharing the same helper, with only the bean return site configured;
- deeper helper chains;
- cross-class helper calls;
- mixed supported and unsupported return paths in the same bean;
- non-bean usage;
- idempotency.

## Validation

The focused recipe test suite has been expanded around the per-return provenance model. The PR remains draft while CI validates the updated implementation.

- Closes #1084.
- Part of #1081.


Kyran121:
@timtebeek - ready to review

PR #1131, opened 2026-08-18
Automate recipes.csv generation instead of requiring a manual step

Kyran121:
## Summary
- Contributors previously had to remember to run `recipeCsvGenerate` and commit the diff. Since every recipe change touched the same shared file, PRs regularly collided with each other on rebase.
- `recipeCsvValidateCompleteness`/`recipeCsvValidateContent` now depend on `recipeCsvGenerate`, so every `check`/`build` (locally and in CI) regenerates `recipes.csv` from the freshly built jar before validating it - no separate command, and validation always runs against genuinely current content rather than whatever a contributor happened to commit.
- `recipeCsvGenerate` needs the already-built jar to scan for recipes, so it can't run in the same pass that produces that jar - the committed file on `main` still lags by one merge. A new workflow (`.github/workflows/recipes-csv.yml`) closes that gap: on push to `main` it regenerates and re-validates, then commits the result straight to `main` - no PR, since `main` has no branch protection and the workflow never touches (untrusted) fork PR branches.
- Net effect: contributor branches no longer carry any diff against `recipes.csv`, so there's nothing left to conflict when rebasing onto a `main` whose copy has since moved.

## Test plan
- [ ] `./gradlew check --dry-run` confirms `recipeCsvGenerate` now runs ahead of `recipeCsvValidateCompleteness`/`recipeCsvValidateContent` in the task graph
- [ ] `recipes-csv.yml` YAML is well-formed
- [ ] First push to `main` after merge produces a `[Auto] Regenerate recipes.csv` bot commit if content changed

### DIFF
--- a/.github/workflows/recipes-csv.yml
+++ b/.github/workflows/recipes-csv.yml
@@ -1,0 +1,55 @@
+---
+name: Regenerate recipes.csv
+
+# What this does: after every merge to main, regenerates recipes.csv and commits
+# it straight back to main. Nobody has to run a command or open a PR for it.
+#
+# Why this can't just happen as part of the normal build: recipeCsvGenerate needs
+# the recipe jar to already be built (it scans the jar to find the recipes), but
+# the file it writes also gets packaged *into* that same jar. It can't be an
+# input and an output of the same build, so it has to be generated in a
+# follow-up step like this one, once main already has a jar to scan.
+#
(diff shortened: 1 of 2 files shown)


Kyran121:
@timtebeek - https://claude.ai/code/artifact/98bf10d5-6482-4c0f-8511-d31eb4ecca6d?via=auto_preview

PR #1134, opened 2026-08-20
Replace UNPROCESSABLE_ENTITY with UNPROCESSABLE_CONTENT in UpgradeSpr…

KamilPatora:
## What's changed?

Added a `ReplaceConstantWithAnotherConstant` migration to `UpgradeSpringFramework_7_0` that replaces:

`HttpStatus.UNPROCESSABLE_ENTITY` with `HttpStatus.UNPROCESSABLE_CONTENT`

Added regression coverage for static imports and their usages.

## What's your motivation?

Spring Framework 7.0 removes `UNPROCESSABLE_ENTITY`. Without this migration, projects upgraded by the recipe retain the old constant and fail to compile.

Nope

None.

## Any additional context

### DIFF
--- a/src/main/resources/META-INF/rewrite/spring-framework-70.yml
+++ b/src/main/resources/META-INF/rewrite/spring-framework-70.yml
@@ -27,6 +27,9 @@ recipeList:
       groupId: org.springframework
       artifactId: "*"
       newVersion: 7.0.x
+  - org.openrewrite.java.ReplaceConstantWithAnotherConstant:
+      existingFullyQualifiedConstantName: org.springframework.http.HttpStatus.UNPROCESSABLE_ENTITY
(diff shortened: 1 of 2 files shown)


timtebeek (approved):
Thanks for spotting that deprecation and providing the fix here @KamilPatora !

DidierLoiseau:
There are actually more deprecations in [`HttpStatus`](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/http/HttpStatus.html) in 7.0, as well as the related ones in [`StatusResultMatchers`](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/test/web/servlet/result/StatusResultMatchers.html) 🙂 

(I was actually the instigator of those changes in spring-projects/spring-framework#32870 😅)

PR #1139, opened 2026-08-25
Remove rewrite-analysis dependency and ConstantFold usage

timtebeek:
Drops the `org.openrewrite.meta:rewrite-analysis` dependency, whose only usage was a single `ConstantFold.findConstantLiteralValue` call in `SecurityContextToSecurityScheme` to map SpringFox's `ApiKey` `passAs` argument onto `SecurityScheme.In`.

That call is replaced with a direct `J.Literal` check; the `SecurityScheme.In.HEADER` fallback is unchanged. Constant *references* (`static final String PASS_AS = "query"`) are no longer resolved and now fall back to `HEADER` — rewrite-java core has no equivalent for that, and in practice the argument is an inline string literal.

Added `apiKeyPassedAsCookieToSecurityScheme` so a non-default mapping is covered; previously only the `"header"` path had a test.


timtebeek:
There's a slight chance at a small regression here, but given that this is itself already an older migration I think it's worth it to get rid of the dependency. Alternatively we can reimplement this using the new data flow on the Moderne side.

MBoegers (approved):
I can life with that. `static string` as a value in annotations is rare.

PR #1144, opened 2026-09-10
Cache regex patterns in ChangeSpringPropertyKey

steve-aom-elliott:
## Summary

- Hoist the three `Pattern.compile` calls in `ChangeSpringPropertyKey.JavaPropertyKeyVisitor` out of the per-literal path into `getVisitor()`-scoped locals, and pass them into the visitor via a new constructor. Previously each `@Value` / `@ConditionalOnProperty` / `@SpringBootTest` literal reached at recipe run-time triggered a fresh compile of the `${...}` reference regex or the `^oldKey` prefix regex.
- Replace three `String.replaceFirst(Pattern.quote(oldKey), newKey)` calls (each of which compiles a `Pattern` internally) with `Matcher.replaceFirst` on a cached bare-old-key `Pattern`.
- Add `if (!value.contains(oldPropertyKey)) return literal;` at the top of `changePropertyInLiteral` so `@ConditionalOnProperty` and `@SpringBootTest` literals that don't mention the old key skip the regex path entirely.

The composite is instantiated once per key rename, and the SB2.0 → SB4.1 property-migration chain issues hundreds of instances, so the per-literal compile cost multiplies across `#instances × #Java files × #annotations`. Behavior is unchanged; only allocation and CPU are affected.

## Not included, deliberately

The inner `(?:\\.|[^}])*` group has an alternation where both branches can consume `\`, which is a latent ReDoS on pathological inputs (long runs of backslashes with no closing `}`). Making it possessive (`*+`) or atomic would fix that but changes the match on `\${key\\}` (escape-then-close with no real closer), so it's left for a follow-up with dedicated tests.

## Test plan

- [x] `ChangeSpringPropertyKeyTest`
- [x] `SpringBoot41PropertiesTest`
- [x] `UpgradeSpringBoot2ConfigurationTest`
- [x] `UpgradeSpringBoot40ConfigurationTest`
- [x] `SpringProfilesGroupPropertyTest`
- [x] `MigrateEndpointAccessValueSpring34Test`
- [x] `MarkAdditionalSpringConfigFilesTest`


src/main/java/org/openrewrite/java/spring/ChangeSpringPropertyKey.java
Jammy-Louie: This was also a pain point in a recipe run's profiling.

it doesn't look like we use `nested` for anything except to hold the results which means its always doing a tree walk.

```suggestion
Tree newTree = yamlChangePropertyKey.getVisitor().visit(tree, ctx);
if (newTree != tree && isWrittenAsNestedMappings((Yaml.Documents) tree)) {
```

This way it can short circuit early and skip the tree walk

Jammy-Louie (approved):

PR #1145, opened 2026-09-14
Preserve Kotlin `@Value` escape when renaming property

greg-at-moderne:
## What's changed?

Fix one of the recipes, namely preserve Kotlin `@Value` escape when renaming property in `ChangeSpringPropertyKey`.

## What's your motivation?

Fix broken CI after https://github.com/openrewrite/rewrite/pull/8863


greg-at-moderne:
The broken CI run is for an unrelated reason => Merging.

